PR #24, opened 2020-09-04
Update sensitive detectors

paolafer:
This PR adds to the `petalo` branch the latest changes in the usage of sensitive detectors. It also adds automatic tests for two geometries (full ring and tiles), which serve as an example for future tests to be implemented in the new geometries.


carmenromo (approved):
Last changes in sensitive detector and some tests have been added. All tests pass, so I approve!

PR #25, opened 2020-09-04
Add photoelectric effect physics for optical photons

gonzaponte:
Add physics of photoelectric effect in metals for optical photons.
Currently implemented only for stainless steel in the fate mesh.

### DIFF
--- a/source/physics_lists/NexusPhysics.cc
+++ b/source/physics_lists/NexusPhysics.cc
@@ -34,7 +34,7 @@ namespace nexus {
 
   NexusPhysics::NexusPhysics():
     G4VPhysicsConstructor("NexusPhysics"),
-    clustering_(true), drift_(true), electroluminescence_(true)
+    clustering_(true), drift_(true), electroluminescence_(true), photoelectric_(false)
   {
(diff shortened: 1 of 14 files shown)


jmunozv:
The Optical Photoelectric Process should be switchable (on-off) like the rest of physical processes developed by us.
(Take a look to NexusPhysics.cc).

source/physics_lists/NexusPhysics.cc
```diff
@@ -46,6 +47,10 @@ namespace nexus {
 
     msg_->DeclareProperty("electroluminescence", electroluminescence_,
       "Switch on/off the electroluminescence.");
+
+    msg_->DeclareProperty("photoelectric", photoelectric_,
```
jmunozv, line 51: You need to add a default value for this process at the beginning of the method, and update all the related macro files with the new parameter.

jmunozv (changes requested):

jmunozv:
After rebase to master and check that all tests run, this PR is ready to approve.
It implements the photoelectric effect for optical photons.
Good job!

jmunozv (approved):

PR #29, opened 2020-09-07
Change tracking action to default for EL table generation.

paolafer:
This PR simply changes the tracking action of the example macro for the EL table generation in NEXT 100 to the default one. It was set to OPTICAL by mistake, since in these simulations there's no need to save all the information of the tracking of the optical photons.


jmunozv (approved):
Approved

jmunozv (approved):
Approved

jmunozv (approved):

PR #30, opened 2020-09-07
Add tests for Flex100 geometry

jmunozv:
This PR creates the corresponding tests for Flex100 geometry (NEXT_FLEX geometry set to emulate the NEXT100 detector).

### DIFF
--- a/tests/pytest/persistency_test.py
+++ b/tests/pytest/persistency_test.py
@@ -142,5 +142,4 @@ def test_sensor_names_are_the_same_across_tables(detectors):
         assert any(p == label + '_binning' for p in sns_bin_conf)
 
     for p in sns_bin_conf:
-        p_split = p.split('_')
-        assert p_split[0] in pos_labels
+        assert p[:-8] in pos_labels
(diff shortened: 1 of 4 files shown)


tests/pytest/persistency_test.py
```diff
@@ -142,5 +142,4 @@ def test_sensor_names_are_the_same_across_tables(detectors):
         assert any(p == label + '_binning' for p in sns_bin_conf)
 
     for p in sns_bin_conf:
-        p_split = p.split('_')
-        assert p_split[0] in pos_labels
+        assert p[:-8] in pos_labels
```
paolafer, line 145: I think that using `-8` (the length of the `binning` string) is a bit more obscure than splitting the string in two and taking the first part. I seems to me that the previous implementation has better readability, although I guess it's a matter of taste.

paolafer: Also, this has nothing to do with the PR in hand ;-).

jmunozv: May be the solution is more obscure but it's more general.

Previous implementation is not able to deal with sensor names including the character "_", as it is the case for the Flex100 geometry.

paolafer: Alright, agree, then.

paolafer (approved):
The test that have been added executes correctly and succeeds. I approve this PR.

PR #31, opened 2020-09-07
Generic photosensor visibility

jmunozv:
This PR removes the messenger and the visibility parameter from the GenericPhotosensor class, and introduces a setter to set it.

Classes making use of these sensors now have a specific parameter to set these sensors visibility.

Macro files have been updated accordingly.


paolafer (approved):
This PR allows one to use the GenericPhotosensor visibility from any geometry currently  instantiating it. The code works as expected, I approve it.

PR #32, opened 2020-09-16
DEMO++ energy plane

paolafer:
This PR fixes the dimensions and relative distances of the DEMO++ energy plane. 
A copper ring is added to the sapphire windows and the flanges are removed, since they are not relevant for the optical response of the detector.
A number of cosmetic changes are also implemented, following the current coding conventions.

### DIFF
--- a/source/geometries/NextDemoEnergyPlane.cc
+++ b/source/geometries/NextDemoEnergyPlane.cc
@@ -41,25 +41,18 @@ namespace nexus {
   using namespace CLHEP;
 
   NextDemoEnergyPlane::NextDemoEnergyPlane():
-
+    dist_gate_support_surface_(445.5 * mm),
     num_PMTs_ (3),
-    //   energy_plane_posz_ (-209.5 * mm),  //  from STEP-FILE:: Sapphire PosZ (-21.55 * cm)
-
-    // Carrier Plate dimensions
-    // STEP file:  0610-3  PMTs  FLANGE_PMTs::
     carrier_plate_thickness_ (45. * mm),
-     // OUTER ACTIVE is 229*mm // (194.227mm) It must be consistent with ACTIVE diameter (from drawings)/(194.97mm)/160mm in Next1El
-    carrier_plate_diam_ ( 234. * mm), //WidthOut 280.003 mm,thick 23.049 mm,netRAD 233.905 mm
-    carrier_plate_central_hole_diam_ (10. * mm), // It is instead of central OLD-PMT(It is 10cm for Next100: nozzle_external_diam + 1 cm(equal to the ICS hole))  //  external hole only with NO extra 1 cm, here. //  16 mm or 10 mm
-
-    // Ruty - from  Next100EnergyPlane.cc
-    // Enclosures dimensions - like in Next100 ? Since they are the same PMT's
-    enclosure_length_ (18.434 * cm),
-    enclosure_diam_ (9.6 * cm),
-    enclosure_flange_length_ (19.5 * mm),
-    enclosure_window_thickness_ (7.1 * mm),
-    enclosure_window_diam_ (85. * mm),
-    enclosure_pad_thickness_ (1.0 * mm),
+    carrier_plate_diam_ ( 234. * mm),
+    carrier_plate_central_hole_diam_ (10. * mm),
+    pmt_hole_length_ (18.434 * cm),
+    pmt_hole_diam_ (9.6 * cm),
+    pmt_flange_length_ (19.5 * mm),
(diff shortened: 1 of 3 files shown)


source/geometries/NextDemoEnergyPlane.cc
```diff
@@ -1,14 +1,9 @@
 // ----------------------------------------------------------------------------
-//  $Id$
+// nexus | NextDemoEnergyPlane.cc
 //
-//  Authors: <jmunoz@ific.uv.es>
-//  Created: 25 Apr 2012
+// Energy plane geometry of the DEMO++ detector.
```
jmalbos: Geometry of the DEMO++ energy plane.

source/geometries/NextDemoEnergyPlane.cc
```diff
@@ -32,260 +27,267 @@
 #include <Randomize.hh>
 #include <G4RotationMatrix.hh>
 
-#include <CLHEP/Units/SystemOfUnits.h>
-#include <CLHEP/Units/PhysicalConstants.h>
-#include <stdexcept>
-
 namespace nexus {
 
   using namespace CLHEP;
```
jmalbos: You probably don't need this.

source/geometries/NextDemoEnergyPlane.h
```diff
@@ -1,26 +1,17 @@
...
-///  \date     25 Apr 2012
-///  \version  $Id$
-///
-///  Copyright (c) 2012 NEXT Collaboration
+// nexus | NextDemoEnergyPlane.h
 //
-//  Updated to NextDemo++  by  Ruth Weiss Babai <ruty.wb@gmail.com>
-//  From:   "Next100EnergyPlane.h"
-//  Date:       June 2019
+// Energy plane geometry of the DEMO++ detector.
```
jmalbos: Geometry of the DEMO++ energy plane.

source/geometries/NextDemoEnergyPlane.cc
```diff
@@ -325,12 +320,8 @@ namespace nexus {
...
 	pmt_positions_.push_back(position);
 	total_positions++;
-        //G4cout << "\n\ntotal_positions= ", total_positions;
     }
 
-    //}
-
-    // Checking
     if (total_positions != num_PMTs_) {
       G4cout << "\n\nERROR: Number of PMTs doesn't match with number of positions calculated\n";
```
jmalbos: Replace this with a `G4Exception`.

source/geometries/NextDemoEnergyPlane.cc
```diff
@@ -32,260 +27,267 @@
...
-    /// Messenger
-    msg_ = new G4GenericMessenger(this, "/Geometry/NextDemo/", "Control commands of geometry NextDemo.");
-    msg_->DeclareProperty("energy_plane_vis", visibility_, "Energy Plane Visibility");
+    msg_->DeclareProperty("energy_plane_verbosity", verbosity_,
+                          "Energy Plane verbosity");
 
-    msg_->DeclareProperty("energy_plane_verbosity", verbosity_, "Energy Plane verbosity");
-
-    /// The PMT
     pmt_ = new PmtR11410();
```
jmalbos, line 63: I don't think you need this object to be dynamically created here and kept around till the destructor is called. You could simply do
`PmtR11410 pmt();`
in the function where it is used, and let the scope take care of its destruction.

jmalbos (changes requested):

source/geometries/NextDemoEnergyPlane.cc
paolafer: Done.

source/geometries/NextDemoEnergyPlane.cc
paolafer: Removed.

source/geometries/NextDemoEnergyPlane.cc
paolafer: Replaced.

source/geometries/NextDemoEnergyPlane.cc
paolafer: PmtR11410 needs to be instantiated in the constructor, because it has configuration parameters set via messenger.

source/geometries/NextDemoEnergyPlane.h
paolafer: Done.

jmunozv:
Optical surfaces must be added to PEDOT and TPB coating layers

jmunozv (approved):
I approve this first version of DEMO++ Energy Plane

jmalbos (approved):

PR #33, opened 2020-09-24
Pet box geometry

carmenromo:
This PR adds the current geometry of the pet box, the first PETALO prototype, with the three types of sensors that will be used in the detector.


paolafer (changes requested):
Remove from the PR the `source/.DS_Store` file.

paolafer (changes requested):
Also, change from `true` to `false` the variables responsible of the overlapping. We don't want to print overlap messages once the code is approved.

paolafer (changes requested):
A test should be added to `produce_output_test` and `sensitive_detectors_test` with the new geometry.

carmenromo:
> A test should be added to `produce_output_test` and `sensitive_detectors_test` with the new geometry.

right!

source/geometries/PetBox.cc
```diff
@@ -0,0 +1,418 @@
...
+    G4LogicalVolume* source_tube_logic =
+      new G4LogicalVolume(source_tube_solid, carbon_fiber, "SOURCE_TUBE");
+
+    G4RotationMatrix rot;
+    rot.rotateX(pi/2.);
+    G4double source_tube_ypos = source_tube_length_/2.-(ih_y_size_-2.*ih_thickness_)/2.;
+    new G4PVPlacement(G4Transform3D(rot, G4ThreeVector(0., source_tube_ypos, 0.)),
+      source_tube_logic,"SOURCE_TUBE", vacuum_hat_logic, false, 0, true);
+
+    G4double air_source_tube_len = source_tube_length_/2.-source_tube_thickness_;
```
paolafer: It is better to be consistent throughout the code and use `length` variables as lenghts (not half lengths) and in G4Tubs use `length/2`. If we do it always the same way it's easier to check that things are corrects.

carmenromo:
> Remove from the PR the `source/.DS_Store` file.

Done!

carmenromo:
> Also, change from `true` to `false` the variables responsible of the overlapping. We don't want to print overlap messages once the code is approved.

Done!

paolafer (changes requested):
I still find a misalignment of volumes in the VUV tile of Hamamatsu. A particle travelling from the centre goes first through `ACTIVE_LXE_TILE`, then through `TILE_QUARTZ_WINDOW ` and finally through `ACTIVE_LXE_TILE` again (see attached screenshot).

Screenshot 2020-09-28 at 17 31 20

source/geometries/TileHamamatsuVUV.cc
```diff
@@ -0,0 +1,178 @@
...
+    // SiPMs
+    G4LogicalVolume* sipm_logic = sipm_->GetLogicalVolume();
+
+    for (int i=0; i<n_rows_; i++){
+      for (int j=0; j<n_columns_; j++){
+        G4int copy_no = i*2*n_columns_ + j + 1;
+        G4double x_pos = -tile_x_/2. + offset_x + sipm_dim.x()/2. + j * sipm_pitch_;
+        G4double y_pos =  tile_y_/2. - offset_y - sipm_dim.y()/2. - i * sipm_pitch_;
+        G4double z_pos = tile_z_/2. - quartz_thick_ - lxe_thick_ - sipm_dim.z()/2.;
+        G4String vol_name = "SiPMHmstuVUV_" + std::to_string(copy_no);
```
paolafer: It's fine to shorten the name, but let's keep the order of the letters (from Hamamatsu it should be `Hmtsu`, not `Hmstu`).

source/geometries/SiPMpetBlue.cc
```diff
@@ -0,0 +1,201 @@
+// ----------------------------------------------------------------------------
+// nexus | SiPMpetBlue.cc
```
paolafer: Why don't you call this class also SiPMHamamatsuBlue, following the same convention as the other two?

source/geometries/PetBox.h
```diff
@@ -0,0 +1,65 @@
+// ----------------------------------------------------------------------------
+// nexus | PetBox.cc
+//
+// This class implements the geometry of a symmetric box of LXe.
+//
+// The NEXT Collaboration
+// ----------------------------------------------------------------------------
+
+#ifndef PET_BOX_H_
```
paolafer: Drop the trailing underscore, following the [convention rules](https://github.com/next-exp/nexus/wiki/Coding-conventions-and-style-guide).

paolafer (changes requested):

source/geometries/TileHamamatsuVUV.cc
carmenromo: Ups you are right

source/geometries/SiPMFBKVUV.cc
```diff
@@ -0,0 +1,157 @@
...
+    G4String sdname = "/SIPM/SiPMFBKVUV";
+    G4SDManager* sdmgr = G4SDManager::GetSDMpointer();
+
+    if (!sdmgr->FindSensitiveDetector(sdname, false)) {
+      ToFSD* sipmsd = new ToFSD(sdname);
+      sipmsd->SetDetectorVolumeDepth(sensor_depth_);
+      sipmsd->SetMotherVolumeDepth(mother_depth_);
+      sipmsd->SetDetectorNamingOrder(naming_order_);
+      sipmsd->SetTimeBinning(time_binning_);
+      G4cout << "Box geom in SIPMFBK!: " << box_geom_ << G4endl;
```
paolafer: Remove comment.

source/geometries/SiPMHamamatsuVUV.cc
```diff
@@ -0,0 +1,160 @@
...
+			            eff_(1.),
+                                    time_binning_(200.*nanosecond)
+
+  {
+    /// Messenger
+    msg_ = new G4GenericMessenger(this, "/Geometry/SiPMpet/", "Control commands of geometry.");
+    msg_->DeclareProperty("visibility", visibility_, "SiPMpet Visibility");
+    msg_->DeclareProperty("efficiency", eff_, "Efficiency of SiPM");
+
+    G4GenericMessenger::Command& time_cmd =
```
paolafer: Since the tiles are not instantiated in the `PetBox` constructor anymore, we cannot use configuration parameters in the tile and SiPM classes anymore. If you add to your macro, for instance, `/Geometry/SiPMpet/time_binning 2. microsecond`, the job crashes because that command has not been registered in construction.

Everything needs to be set through `setters` methods, and the parameters need to appear in the `PetBox` class. Review all the tile and SiPM classes about this.

source/geometries/PetBox.cc
```diff
@@ -0,0 +1,458 @@
...
+    G4Box* lab_solid = new G4Box("LAB", lab_size/2., lab_size/2., lab_size/2.);
+
+    G4Material* air = G4NistManager::Instance()->FindOrBuildMaterial("G4_AIR");
+    lab_logic_ = new G4LogicalVolume(lab_solid, air, "LAB");
+    lab_logic_->SetVisAttributes(G4VisAttributes::Invisible);
+    this->SetLogicalVolume(lab_logic_);
+
+    if (tile_type_ == "HamamatsuVUV") {
+      tile_ = new TileHamamatsuVUV();
+      box_geom_ = 1;
```
paolafer: I would put all the variables that depend on the type of tile here, so they're in the same place and one can see in one glance what's the difference between the configurations. Also, this way you can avoid another `if else` clause below. So, just make `dist_dice_flange` a global variable and move it here.

source/geometries/PetBox.cc
```diff
@@ -0,0 +1,458 @@
...
+    tile_->SetBoxGeom(box_geom_);
+    tile_->Construct();
+    tile_thickn_ = tile_->GetDimensions().z();
+    full_row_size_ = n_tile_columns_ * tile_->GetDimensions().x();
+    full_col_size_ = n_tile_rows_ * tile_->GetDimensions().y();
+
+    G4LogicalVolume* tile_logic = tile_->GetLogicalVolume();
+
+    if (tile_type_ == "HamamatsuVUV") {
+      x_pos_ = ih_x_size_/2. + dist_ihat_panel_ + panel_thickness_ + active_depth_ + tile_thickn_/2.;
```
paolafer: I think that the `x_pos_` variable doesn't need to be global. As a general rule, if a variable is used only in one scope, make it local: global variables are needed when they have to be used in more than one scope.

source/geometries/PetBox.cc
```diff
@@ -0,0 +1,458 @@
...
+    // TILE CONSTRUCT
+    tile_->SetBoxGeom(box_geom_);
+    tile_->Construct();
+    tile_thickn_ = tile_->GetDimensions().z();
+    full_row_size_ = n_tile_columns_ * tile_->GetDimensions().x();
+    full_col_size_ = n_tile_rows_ * tile_->GetDimensions().y();
+
+    G4LogicalVolume* tile_logic = tile_->GetLogicalVolume();
+
+    if (tile_type_ == "HamamatsuVUV") {
```
paolafer: I would write `x_pos` for this case with the same structure as the other two tiles; just place a placeholder with a reasonable value (let's say, 20 mm), till we know the exact distance. Also add a comment in the code to remind us that we need to update that number.

source/geometries/PetBox.cc
```diff
@@ -0,0 +1,458 @@
...
+    G4Material* LXe = G4NistManager::Instance()->FindOrBuildMaterial("G4_lXe");
+    LXe->SetMaterialPropertiesTable(OpticalMaterialProperties::LXe());
+    LXe_logic_ =
+      new G4LogicalVolume(LXe_solid, LXe, "LXE");
+
+    new G4PVPlacement(0, G4ThreeVector(0., 0., 0.), LXe_logic_,
+                      "LXE", box_logic, false, 0, false);
+
+
+    // Aliminum cylinder
```
paolafer: Typo: aliminum --> aluminum.

paolafer (changes requested):

source/geometries/PetBox.cc
```diff
@@ -0,0 +1,458 @@
...
+    new G4PVPlacement(0, G4ThreeVector(0., (-box_size_/2.+box_thickness_+ih_y_size_/2.), 0.),
+      internal_hat_logic, "INTERNAL_HAT", LXe_logic_, false, 0, false);
+
+    G4double vacuum_hat_xsize = ih_x_size_-2.*ih_thick_wall_;
+    G4double vacuum_hat_ysize = ih_y_size_-ih_thick_roof_;
+    G4double vacuum_hat_zsize = ih_z_size_-2.*ih_thick_wall_;
+    G4Box* vacuum_hat_solid =
+      new G4Box("VACUUM_HAT", vacuum_hat_xsize/2.,vacuum_hat_ysize/2., vacuum_hat_zsize/2.);
+
+    G4Material* vacuum = G4NistManager::Instance()->FindOrBuildMaterial("G4_Galactic");;
```
paolafer: Remove one semicolon.

source/geometries/PetBox.cc
```diff
@@ -0,0 +1,458 @@
...
+      new G4LogicalVolume(horiz_low_lat_panel_solid, pyrex, "LAT_PANEL");
+
+    new G4PVPlacement(0, G4ThreeVector(-low_lat_panel_xpos, low_lat_panel_ypos, 0),
+                      horiz_low_lat_panel_logic, "LAT_PANEL", LXe_logic_, false, 1, false);
+
+    new G4PVPlacement(0, G4ThreeVector(low_lat_panel_xpos, low_lat_panel_ypos, 0),
+                      horiz_low_lat_panel_logic, "LAT_PANEL", LXe_logic_, false, 2, false);
+
+
+    G4double lat_panel_xpos = entry_panel_xpos + dist_entry_panel_lat_panel_ + lat_panel_width_/2.;
```
paolafer: I think that `+ panel_thickness_/2.` is missing in the `lat_panel_xpos` variable.

source/geometries/PetBox.cc
```diff
@@ -0,0 +1,458 @@
...
+    new G4PVPlacement(0, G4ThreeVector(-entry_panel_xpos, entry_panel_ypos, 0), entry_panel_logic,
+                        "ENTRY_PANEL", LXe_logic_, false, 1, false);
+
+    new G4PVPlacement(0, G4ThreeVector(entry_panel_xpos, entry_panel_ypos, 0), entry_panel_logic,
+                        "ENTRY_PANEL", LXe_logic_, false, 2, false);
+
+
+    // PYREX PANELS SURROUNDING THE SIPM DICE BOARDS
+    // LOWER HORIZONTAL PANELS ARE SMALLER IN X
+    G4double low_lat_panel_xpos = entry_panel_xpos + dist_entry_panel_lat_panel_ +
```
paolafer: I think that `+ panel_thickness_/2.` is missing in the `low_lat_panel_xpos` variable.

source/geometries/PetBox.cc
```diff
@@ -0,0 +1,458 @@
...
+                    source_tube_int_radius_(1.1*cm),
+                    dist_source_roof_(10.*mm),
+                    source_tube_thick_roof_(5.*mm),
+                    n_tile_rows_(2),
+                    n_tile_columns_(2),
+                    dist_ihat_panel_(2.*mm),
+                    panel_thickness_(1.75*mm),
+                    active_depth_(3.*cm),
+                    entry_panel_width_(86.*mm),
+                    dist_entry_panel_lat_panel_(1.45*mm),
```
paolafer: For variables that are a bit tricky, and have not self-evident names, it is useful to add a comment with `//`, so that in the future we don't get crazy remembering what they were. For instance I would add a comment for `dist_entry_panel_lat_panel_` such as **x distance between the internal surface of the entry panel and the edge of the lateral panel**.

source/geometries/PetBox.cc
```diff
@@ -0,0 +1,458 @@
...
+                    ih_z_size_(6.*cm),
+                    ih_thick_wall_(3.*mm),
+                    ih_thick_roof_(6.*mm),
+                    source_tube_thick_wall_(1.*mm),
+                    source_tube_int_radius_(1.1*cm),
+                    dist_source_roof_(10.*mm),
+                    source_tube_thick_roof_(5.*mm),
+                    n_tile_rows_(2),
+                    n_tile_columns_(2),
+                    dist_ihat_panel_(2.*mm),
```
paolafer: I would add a comment also here: **x distance between the internal surface of the hat and the external surface of the entry panel**.

paolafer (changes requested):

source/geometries/TileHamamatsuBlue.cc
```diff
@@ -0,0 +1,126 @@
...
+
+    // OPTICAL SURFACE FOR REFLECTION
+    G4OpticalSurface* fr4_opsurf =
+      new G4OpticalSurface("FR4_OPSURF", unified, polished, dielectric_metal);
+    fr4_opsurf->SetMaterialPropertiesTable(OpticalMaterialProperties::ReflectantSurface(reflectivity_));
+
+    new G4LogicalSkinSurface("FR4_OPSURF", tile_logic, fr4_opsurf);
+
+
+    G4int sensor_depth = 1;
```
paolafer: You can write directly `1` and `2` in the setters below; the code is well readable, since the setter methods describe what they're doing in the name

source/geometries/TileHamamatsuBlue.cc
```diff
@@ -0,0 +1,126 @@
...
+
+    sipm_ = new SiPMHamamatsuBlue();
+  }
+
+  TileHamamatsuBlue::~TileHamamatsuBlue()
+  {
+  }
+
+  void TileHamamatsuBlue::Construct()
+  {
```
paolafer, line 55: In this tile we have to add also a layer of wavelength shifter. You can add it made of TPB, for the moment, when we decide which material to use we'll refine the simulation. The thickness can be 1 micrometer, as in NEXT.

source/geometries/SiPMHamamatsuBlue.cc
```diff
@@ -0,0 +1,182 @@
...
+      new G4LogicalVolume(sipm_solid, plastic, "SiPMHmtsuBlue");
+
+    this->SetLogicalVolume(sipm_logic);
+
+
+    // EPOXY PROTECTIVE LAYER ////////////////////////////////////////
+
+    G4double epoxy_depth = 0.1 * mm;
+
+    G4Box* epoxy_solid =
```
paolafer: Move the epoxy layer to the tile class.

source/geometries/PetBox.cc
```diff
@@ -0,0 +1,458 @@
...
+      new G4LogicalVolume(air_source_tube_inside_box_solid, air, "AIR_SOURCE_TUBE");
+
+    new G4PVPlacement(0, G4ThreeVector(0., 0., 0.), air_source_tube_inside_box_logic,
+                      "AIR_SOURCE_TUBE", source_tube_inside_box_logic, false, 0, false);
+
+
+    // 2 ACTIVE REGIONS
+    G4double active_x_pos = ih_x_size_/2. + dist_ihat_panel_ + panel_thickness_ + active_depth_/2.;
+
+    G4Box* active_solid =
```
paolafer, line 279: The active dimensions should be exactly the dimensions of the LXe enclosed in the panels and between the surface of the tiles and the wall of the internal hat. We should write the code in such a way that the x dimension of ACTIVE depends on the particular `dist_dice_flange` of each tile.

paolafer (changes requested):

source/geometries/SiPMHamamatsuVUV.cc
```diff
@@ -0,0 +1,160 @@
...
+
+    G4String sdname = "/SIPM/SiPMHmtsuVUV";
+    G4SDManager* sdmgr = G4SDManager::GetSDMpointer();
+
+    if (!sdmgr->FindSensitiveDetector(sdname, false)) {
+        ToFSD* sipmsd = new ToFSD(sdname);
+
+        sipmsd->SetDetectorVolumeDepth(sensor_depth_);
+        sipmsd->SetMotherVolumeDepth(mother_depth_);
+        sipmsd->SetDetectorNamingOrder(naming_order_);
```
paolafer, line 141: Initialize `naming_order` and `mother_depth` to `0` and `sensor_depth` to `-1` in the initialization list. We should also add a check before `sipmsd->SetDetectorVolumeDepth(sensor_depth_);` to ensure that `sensor_depth_` is different than `-1`, since we always need to set that variable. However, mother_depth and naiming_order can be left to the default. The `box_geom_` variable needs also to be initialized. 

Do it in the 3 SiPM classes.

paolafer (changes requested):

source/geometries/SiPMHamamatsuVUV.cc
carmenromo: I see what you mean, but is it worthy to remove all the conf parameters of these classes? At the end, this is a very concrete case, and if we want to use the SiPM classes for instance in another geometry, we will not be able to use them as in other geometries we have...

paolafer: The problem is that the code as it is now crashes, if you use one of those configuration parameters, and we need to fix it. Either we eliminate the possibility of changing the values of those parameters (which I don't like), or we use setters. I don't see a third option, do you?

paolafer: You will need to use setters in other geometries as well, of course.

source/geometries/PetBox.cc
carmenromo: But in our case, x_pos_ depends on the tile we are using, should it be local too?

source/geometries/PetBox.cc
carmenromo: Oh, I see, this is the answer for my previous question 👍

source/geometries/PetBox.cc
carmenromo: You are right

source/geometries/TileHamamatsuBlue.cc
carmenromo: You are right, I forgot!

source/geometries/SiPMHamamatsuVUV.cc
carmenromo: Ok, I see, done!

paolafer:
More things: tests should be added to check the correct functioning of sensors, the same kind as those for the full body or the old tile geometry.

source/sensdet/ToFSD.h
```diff
@@ -85,6 +88,8 @@ namespace nexus {
 
     G4OpBoundaryProcess* boundary_; ///< Pointer to the optical boundary process
 
+    G4int box_geom_; ///< Boolean required to change the naming_order_ for the case of the BoxSetup
```
paolafer, line 91: This variable needs to be initialized in the `.cc` file.

source/sensdet/ToFSD.cc
```diff
@@ -140,10 +140,25 @@ namespace nexus {
       G4int motherid = touchable->GetCopyNumber(mother_depth_);
       pmtid = naming_order_*motherid + pmtid;
     }
+    if (box_geom_==1) { // Hamamatsu
```
paolafer, line 143: The last `else` makes irrelevant the `if (naming_order_ != 0)` clause.
I would do the following:

```
G4int pmtid = touchable->GetCopyNumber(sensor_depth_);
if (naming_order_ != 0) {
    G4int motherid = touchable->GetCopyNumber(mother_depth_);
    pmtid = naming_order_*motherid + pmtid;
}

if (box_geom_==1) { // Hamamatsu
      std::vector<G4int> init_ids ({ 0, 4, 32, 36, 64, 68, 96, 100 });
      G4int motherid = touchable->GetCopyNumber(mother_depth_);
      G4int first_id = (init_ids)[motherid];
      pmtid = first_id + pmtid;
} else if (box_geom_==2) { // FBK
      std::vector<G4int> init_ids ({ 0, 8, 128, 136, 256, 264, 384, 392 });
      G4int motherid = touchable->GetCopyNumber(mother_depth_);
      G4int first_id = (init_ids)[motherid];
      pmtid = first_id + pmtid;
}

return pmtid;
```

paolafer (changes requested):

source/sensdet/ToFSD.cc
carmenromo: Done!

source/geometries/SiPMHamamatsuBlue.cc
carmenromo: Done!

source/geometries/PetBox.cc
carmenromo: Done!

source/geometries/TileHamamatsuBlue.cc
```diff
@@ -0,0 +1,148 @@
...
+    G4Box* wls_solid =
+      new G4Box("WLS", tile_x_/2., tile_y_/2., wls_depth_/2);
+
+    G4Material* wls = MaterialsList::TPB();
+    wls->SetMaterialPropertiesTable(OpticalMaterialProperties::TPB_LXe());
+
+    G4LogicalVolume* wls_logic =
+      new G4LogicalVolume(wls_solid, wls, "Epoxy");
+
+    new G4PVPlacement(0, G4ThreeVector(0., 0., tile_z_/2. - wls_depth_/2.),
```
paolafer: Be careful! You're placing the wavelengthshifting layer on top of the epoxy layer!

source/geometries/PetBox.cc
```diff
@@ -0,0 +1,476 @@
...
+    G4double active_x_pos_max = box_size_/2. - box_thickness_ - dist_dice_flange_ - tile_thickn_;
+    G4double active_x_pos_min = ih_x_size_/2. + dist_ihat_panel_ + panel_thickness_;
+    G4double active_depth = active_x_pos_max - active_x_pos_min;
+    G4double active_x_pos = active_x_pos_min + active_depth/2.;
+
+    G4Box* active_solid =
+      new G4Box("ACTIVE", active_depth/2., dist_lat_panels_/2., dist_lat_panels_/2.);
+    G4LogicalVolume* active_logic =
+      new G4LogicalVolume(active_solid, LXe, "ACTIVE");
+    new G4PVPlacement(0, G4ThreeVector(-active_x_pos, 0., 0.), active_logic,
```
paolafer: There's an overlap warning between ACTIVE and TILE_7. When you add a new volume, or modify an existing one, always make sure that no overlaps are generated.

source/geometries/TileHamamatsuBlue.cc
carmenromo: You are right! But the order coming from outside is: wavelength shifter, epoxy and then tile, right?
So for the wavelength shifter the placement is correct but no the epoxy one, isn't it?

source/geometries/PetBox.cc
carmenromo: I don't have this overlap... Which tile are you using?

source/geometries/TileHamamatsuBlue.cc
paolafer: Yes, that's the correct order. You have to put the epoxy behind the wavelength shifter.

source/geometries/PetBox.cc
paolafer: HamamatsuBlue.

paolafer:
Tests are failing because you modified the `conftest` file without adding the corresponding test. Please, fix it.

paolafer:
In fact, I think we should have one test for each configuration of the box, so 3 tests.

source/geometries/TileHamamatsuBlue.cc
```diff
@@ -0,0 +1,148 @@
...
+    sipm_ = new SiPMHamamatsuBlue();
+  }
+
+  TileHamamatsuBlue::~TileHamamatsuBlue()
+  {
+  }
+
+  void TileHamamatsuBlue::Construct()
+  {
+    SetDimensions(G4ThreeVector(tile_x_, tile_y_, tile_z_+epoxy_depth_+wls_depth_));
```
paolafer: If you make the thickness of the volume tile a bit larger so that the effective tile thickness remain the same, you have to use this total thickness also in the placement of the epoxy and the wavelength shifter.
For instance, the wavelength shifting should be placed as in the following:

`new G4PVPlacement(0, G4ThreeVector(0., 0., (tile_z_+epoxy_depth_+wls_depth_)/2. - wls_depth_/2.)`

carmenromo:
> In fact, I think we should have one test for each configuration of the box, so 3 tests.

Yeah, absolutely, this was my idea too, but I found some problems and I left the problem there... I have to finish them.

source/geometries/TileHamamatsuBlue.cc
carmenromo: You are right!

source/geometries/TileHamamatsuVUV.h
```diff
@@ -0,0 +1,58 @@
...
+    // Rows and columns of SiPMs
+    G4int n_rows_, n_columns_;
+
+    // Thickness of the LXe and quartz in front of the sensors
+    G4double lxe_thick_;
+    G4double quartz_rindex_;
+    G4double quartz_thick_;
+
+     // Messenger for the definition of control commands
+    G4GenericMessenger* msg_;
```
paolafer: This variable is unused, please, delete it.

paolafer (commented):
Have you already modified the panels according to the new drawings?

source/geometries/TileHamamatsuVUV.h
carmenromo: oh, I just saw this. I'll do it

carmenromo:
> Have you already modified the panels according to the new drawings?

Yes!

paolafer:
Reading the full geometry together, I realized that we use the `x` and `y` coordinates in the natural way for the tile geometries (the positions of the SiPMs), while in the box geometry we use `z` and `y` instead. I think this could lead to a lot of headaches, later, because in the database the SiPM positions will probably be `x` and `y`. I think it is better to change it now, simply placing the arrays of SiPMs along z instead of along x, what do you think? Would this change have any implications I'm not seeing?

carmenromo:
> Reading the full geometry together, I realized that we use the `x` and `y` coordinates in the natural way for the tile geometries (the positions of the SiPMs), while in the box geometry we use `z` and `y` instead. I think this could lead to a lot of headaches, later, because in the database the SiPM positions will probably be `x` and `y`. I think it is better to change it now, simply placing the arrays of SiPMs along z instead of along x, what do you think? Would this change have any implications I'm not seeing?

Yes, I think you are right... And don't you think it would be clearer if we change the pet box geometry instead of the sensors? It is more work but not so difficult I think, and maybe it's worthy. What do you think?

paolafer:
> > Reading the full geometry together, I realized that we use the `x` and `y` coordinates in the natural way for the tile geometries (the positions of the SiPMs), while in the box geometry we use `z` and `y` instead. I think this could lead to a lot of headaches, later, because in the database the SiPM positions will probably be `x` and `y`. I think it is better to change it now, simply placing the arrays of SiPMs along z instead of along x, what do you think? Would this change have any implications I'm not seeing?
> 
> Yes, I think you are right... And don't you think it would be clearer if we change the pet box geometry instead of the sensors? It is more work but not so difficult I think, and maybe it's worthy. What do you think?

Yes, I wouldn't touch the tile geometry, simply placing the tiles along z instead of along x (which is what you mean with changing the pet box geometry, I understand). The box is a cube, so it is irrelevant if we put the tiles in (0, 0, +/-z) or in (+/-x, 0, 0). The other things to change are the hat, which is not symmetric and should be changed accordingly, and the position of the panels.

source/geometries/PetBox.cc
```diff
@@ -0,0 +1,519 @@
...
+      G4Material* wls = MaterialsList::TPB();
+      wls->SetMaterialPropertiesTable(OpticalMaterialProperties::TPB());
+
+      G4LogicalVolume* wls_logic =
+        new G4LogicalVolume(wls_solid, wls, "WLS");
+
+      G4double wls_xpos = box_size_/2. - box_thickness_ - dist_dice_flange_ - tile_thickn_
+                                - dist_sipms_panel_sipms_ - panel_thickness_ - wls_depth_/2.;
+
+      new G4PVPlacement(0, G4ThreeVector(wls_xpos, 0., 0.), wls_logic,
```
paolafer: You forgot to add the wavelength shifter to the second panel.

paolafer: In fact, if you place the wavelength shifter inside the panel, instead of in the LXe, you save a line of code (only 1 placement instead of 2).

paolafer: In the case of the Hamamatsu blue-sensitive SiPMs we need to simulate the panels in a different way, because they are transparent to the optical photons in that range of wavelengths (this is why we can put them in front of the SiPMs). We can think about it later and focus on finishing this PR for the other two configurations.

paolafer:
@carmenromo, can you rebase this branch on upstream/petalo? There are some modifications there that affect test files. You'll probably have to solve conflicts.

carmenromo:
> @carmenromo, can you rebase this branch on upstream/petalo? There are some modifications there that affect test files. You'll probably have to solve conflicts.

Sure! I'll do it

paolafer:
Persistency tests are failing for me after these last commits, with `FileNotFoundError`. Are they for you?

carmenromo:
> Persistency tests are failing for me after these last commits, with `FileNotFoundError`. Are they for you?

You are right, I thought I did check them... I'll fix it!

carmenromo:
> Persistency tests are failing for me after these last commits, with `FileNotFoundError`. Are they for you?

Fixed! There was an extra `.h5` term in the macros for the pet box cases.

paolafer (approved):
This PR adds the geometry of the first PETALO set-up (a.k.a. _cajita_) to nexus. The case with wavelength shifter needs further development, but the geometry is ready to be used with VUV-sensitive SiPMs. It has been a huge work, following the always evolving technical drawings and our own understanding of the simulation. Congratulations @carmenromo!

PR #34, opened 2020-10-05
Fix margin between edge of NEXT-100 SiPM board and external crown of SiPMs

jmalbos:
The margin between the edge of the NEXT-100 SiPM board and the first crown of SiPMs is changed from 7.33 mm to 7.275 mm. That way, the sensors are symmetrically positioned with respect to the center of the board: 
```
board_size = pitch * 7 + margin * 2 = 15.55 * 7 + 7.275  * 2 = 123.40 mm
```


paolafer (approved):
This PR is simple and fixes a bug in the positioning of SiPMs in the NEXT100 boards. I approve it.

PR #38, opened 2020-10-20
Flex Tracking Plane Reference Distance

jmunozv:
This PR changes the "FLEX" geometry configuration parameter used to set the Tracking Plane position, by using the distance between ANODE and Kapton surface instead of the distance between ANODE and the SiPMs surface.
Macro file is updated accordingly.

### DIFF
--- a/source/geometries/NextFlexTrackingPlane.cc
+++ b/source/geometries/NextFlexTrackingPlane.cc
@@ -49,7 +49,7 @@ NextFlexTrackingPlane::NextFlexTrackingPlane():
   SiPM_visibility_   (false),
   msg_               (nullptr),
   wls_mat_name_      ("TPB"),
-  SiPM_ANODE_dist_   (10.  * mm),   // Distance from ANODE to SiPM surface
+  kapton_anode_dist_ (15.  * mm),   // Distance from ANODE to Kapton surface
   SiPM_size_x_       ( 1.3 * mm),   // Size X (width) of SiPMs
   SiPM_size_y_       ( 1.3 * mm),   // Size Y (height) of SiPMs
   SiPM_size_z_       ( 2.0 * mm),   // Size Z (thickness) of SiPMs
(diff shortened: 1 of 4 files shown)


source/geometries/NextFlexTrackingPlane.cc
```diff
@@ -45,11 +45,12 @@ NextFlexTrackingPlane::NextFlexTrackingPlane():
   BaseGeometry(),
   mother_logic_      (nullptr),
   verbosity_         (false),
+  sipm_verbosity_    (false),
   visibility_        (false),
   SiPM_visibility_   (false),
   msg_               (nullptr),
   wls_mat_name_      ("TPB"),
-  SiPM_ANODE_dist_   (10.  * mm),   // Distance from ANODE to SiPM surface
+  kapton_anode_dist_ (15.  * mm),   // Distance from ANODE to Kapton surface
```
paolafer, line 53: Is the default value the correct one for Flex100? Or is it a random value?
I'm asking because the value in the initialization list and that in the example macro are different.

paolafer:
You need to update the `test_create_nexus_output_file_flex100` test.

source/geometries/NextFlexTrackingPlane.cc
jmunozv: This value here is the default value for FLEX100, but it could be anyone.
Same for the config file.

paolafer: Ok, I just wanted to make sure that you were aware that the two values were different, in case you wanted any of them to be a sensible value.

jmunozv:
> You need to update the `test_create_nexus_output_file_flex100` test.

Done

paolafer (approved):
This PR changes the reference distance for the tracking plane placement in NextFlex. The tests have been updated accordingly. I approve it.

PR #39, opened 2020-10-21
Add environment file to run tests.

paolafer:
This PR adds a conda environment file, to be used to create and environment to run tests.
A minimal set of packages is included, and it will be expanded in the future if we need more packages.

### DIFF
--- a/scripts/test_environment.yml
+++ b/scripts/test_environment.yml
@@ -5,5 +5,7 @@ dependencies:
   - pandas=1.1.3
   - pytables=3.6.1
   - pytest=6.1.1
-  - pytest-ordering=0.6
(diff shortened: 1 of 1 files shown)


gonzaponte:
I would recommend fixing the version of each package. The default is to take the latest release which can lead to test failures if
- The release contains a bug, which happens from time to time, or
- The release contains a backwards-incompatible change

This option has the drawback of having to update the versions from time to time, but I think it is worth it.

gonzaponte (approved):
Environment file with minimal dependencies. `pytest-ordering2` is, in principle, temporary until they sort out the future of that package.

Approved.

PR #43, opened 2020-10-30
Tests and macros

paolafer:
This PR adds the DEMO++ geometry to the basic tests and two sets macros as example of DEMO++ simulations.


tests/pytest/conftest.py
```diff
@@ -36,27 +36,47 @@ def nexus_full_output_file_next100(output_tmpdir, full_base_name_next100):
...
+def nexus_full_output_file_demopp(output_tmpdir, full_base_name_demopp):
+    return os.path.join(output_tmpdir, full_base_name_demopp + '.h5')
+
+
 @pytest.fixture(scope = 'session')
 def new_detector(nexus_full_output_file_new):
-    num_pmts        = 12
-    num_boards      = 28
+    pmt_ids         = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]
+    board_ids       = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28]
```
jmunozv: Why don't you use the shortest formula for all these IDs definition:
`[i for i in range(12)]`

tests/pytest/sensitive_detectors_test.py
```diff
@@ -1,37 +1,35 @@
...
     assert len(sns_positions.sensor_name.unique()) == 2
 
     # Assert the total number of sensors is correct
-    pmt_ids  = sns_positions[sns_positions.sensor_name == "PmtR11410"].sensor_id
-    sipm_ids = sns_positions[sns_positions.sensor_name == "SiPM"].sensor_id
-    assert     len(pmt_ids)  == num_pmts
-    assert 1 < len(sipm_ids) <= num_sipms
+    sns_pmt_ids  = sns_positions[sns_positions.sensor_name == "PmtR11410"].sensor_id
+    sns_sipm_ids = sns_positions[sns_positions.sensor_name == "SiPM"].sensor_id
+    assert 1 < len(sns_sipm_ids) <= num_sipms
```
jmunozv, line 21: The comparison of PMTs is missing, although may be is redundant with following one.
If we consider that it mustn't be here, then move the line `pmt_ids = ...` to wherever is needed.

jmunozv (changes requested):

jmunozv (changes requested):
Prior to approving this PR I would prefer to wait until DEMO Vessel is accepted, and then update tests and macros of this PR accordingly.

Anyway, I think It would be a good idea to have default macro files set with parameters (presuure, EL intensity, ...) closer to real detectors.

tests/pytest/sensitive_detectors_test.py
paolafer: This assert was redundant, since it is included in `assert np.all(sns_pmt_ids.sort_values().tolist() == pmt_ids)`. 
I'll move the line mentioned below.

tests/pytest/conftest.py
paolafer: Done. I didn't use it for DEMO++, though, because the list is short and I think that seeing the numbers explicitly is clearer.

paolafer:
> Anyway, I think It would be a good idea to have default macro files set with parameters (presuure, EL intensity, ...) closer to real detectors.

The intensity of the EL field of the macro examples are usually kept low, so that tests don't take forever to run. 

Anyway, in the DEMO++ case, the pressure in run 5 and 6 is 8.5 bar, so I could change it, if you want, and the EL field is not far from 10 kV/cm, at least in run 5, so we can keep it that low.

tests/pytest/conftest.py
jmunozv: Good

jmunozv (approved):
I approve this PR that fixes a problem in PMT sensors, and introduces DEMO++ macros with its corresponding tests.

PR #44, opened 2020-10-31
DEMO++ Geometry — Pressure Vessel and Main Class

jmalbos:
This PR overhauls the main class of the DEMO++ geometry and updates the description of the detector's pressure vessel.

### DIFF
--- a/source/geometries/NextDemo.cc
+++ b/source/geometries/NextDemo.cc
@@ -41,12 +41,11 @@ void NextDemo::Construct()
   ConstructLab();
 
   vessel_geom_->Construct();
-  G4VPhysicalVolume* vessel_phys_vol =
-    new G4PVPlacement(nullptr, G4ThreeVector(0., 0., 0.),
-                      vessel_geom_->GetLogicalVolume(),
-                      vessel_geom_->GetLogicalVolume()->GetName(),
-                      BaseGeometry::GetLogicalVolume(),
-                      false, 0, false);
+  new G4PVPlacement(nullptr, G4ThreeVector(0., 0., 0.),
+                    vessel_geom_->GetLogicalVolume(),
+                    vessel_geom_->GetLogicalVolume()->GetName(),
+                    BaseGeometry::GetLogicalVolume(),
+                    false, 0, false);
 
   inner_geom_->SetELzCoord(vessel_geom_->GetGateEndcapDistance());
   inner_geom_->SetMotherLogicalVolume(vessel_geom_->GetGasPhysicalVolume()->GetLogicalVolume());
--- a/source/geometries/NextDemo.cc
+++ b/source/geometries/NextDemo.cc
@@ -1,521 +1,83 @@
-// ----------------------------------------------------------------------------
-//  $Id$  Inheriting from previous code in Next1EL.cc
+// -----------------------------------------------------------------------------
+// nexus | NextDemo.cc
 //
-//  Author:  <neuslopezmarch@gmail.com>
+// Geometry of the NEXT-DEMO++ detector.
 //
-//  Created: 16 Feb 2018
-//
-//  Copyright (c) 2010, 2018 NEXT Collaboration
-//
-//  Updated to NextDemo++  by  Ruth Weiss Babai <ruty.wb@gmail.com>
-//  Date:    June 2019
-//  Add "energy_endcap"  &  "tracking_endcap"
-// ----------------------------------------------------------------------------
-
+// The NEXT Collaboration
+// -----------------------------------------------------------------------------
 
 #include "NextDemo.h"
-#include "MaterialsList.h"
-#include "OpticalMaterialProperties.h"
-#include "DiskSource.h"
-#include "Na22Source.h"
-#include "NaIScintillator.h"
-#include "CylinderPointSampler.h"
-#include "MuonsPointSampler.h"
-#include "Visibilities.h"
 
-#include <G4Orb.hh>
+#include "NextDemoVessel.h"
+#include "NextDemoInnerElements.h"
+
 #include <G4Box.hh>
-#include <G4Tubs.hh>
-#include <G4UnionSolid.hh>
-#include <G4SubtractionSolid.hh>
 #include <G4LogicalVolume.hh>
-#include <G4PVPlacement.hh>
-#include <G4NistManager.hh>
-#include <G4Region.hh>
 #include <G4VisAttributes.hh>
-#include <G4SDManager.hh>
-#include <G4UserLimits.hh>
-#include <G4RotationMatrix.hh>
-#include <G4Transform3D.hh>
-#include <G4RunManager.hh>
-#include <G4RandomDirection.hh>
-#include <Randomize.hh>
-#include <G4GenericMessenger.hh>
-#include <G4UnitsTable.hh>
-#include <G4UIcmdWith3VectorAndUnit.hh>
-
-#include <CLHEP/Units/SystemOfUnits.h>
-#include <CLHEP/Units/PhysicalConstants.h>
-#include <stdexcept>
+#include <G4NistManager.hh>
+#include <G4PVPlacement.hh>
 
 using namespace nexus;
-using namespace CLHEP;
+
 
 NextDemo::NextDemo():
-    BaseGeometry(),
-    // LABORATORY //////////////////////////////////
-    lab_size_ (2. * m),
-    //_rot_angle(pi),
-    // VESSEL ////////////////////
-    vessel_diam_   (300. * mm),
-    vessel_length_ (600. * mm),
-    vessel_thickn_ (  3. * mm),
-    // ENDCAPS /////////////////////////////////////
-    endcap_diam_   (330. * mm),
-    endcap_thickn_ ( 30. * mm),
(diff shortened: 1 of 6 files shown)


jmunozv (approved):
Sorry. This comment was for another PR.

Current one is still pending of review.

jmunozv:
Current implementation does not allow to set by parameter the pressure of the gas, and it must.

NextDemo::GenerateVector(...) must be implemented.

CATHODE_GRID & VESSEL overlap.

source/geometries/NextDemoInnerElements.h
```diff
@@ -1,21 +1,13 @@
...
-///           Energy plane, field cage and tracking plane.
-///  \author   <miquel.nebot@ific.uv.es>
-///  \date     18 Sept 2013
-///  \version  $Id$
-///
-///  Copyright (c) 2013 NEXT Collaboration
+// ----------------------------------------------------------------------------
+// nexus | NextDemoInnerElements.h
 //
+// Geometry
```
jmunozv: Complete. Here and in .cc

source/geometries/NextDemo.cc
```diff
@@ -1,521 +1,84 @@
...
-                        sideport_pos.getZ());
-
-      source_gen_side_ = new CylinderPointSampler(0., source_thick, source_diam/2., 0., side_pos_gen, sideport_rot);
-
-    }
-
+  ConstructLab();
+
+  vessel_geom_->Construct();
+  G4VPhysicalVolume* vessel_phys_vol =
```
jmunozv: This variable is not needed

source/geometries/NextDemoVessel.cc
```diff
@@ -1,334 +1,117 @@
...
-      G4Exception("[NextDemoVessel]", "Construct()", FatalException,
-		  "Unknown kind of gas, valid options are: naturalXe, enrichedXe, depletedXe, Ar.");
-    }
+NextDemoVessel::NextDemoVessel():
+  BaseGeometry(),
+  vessel_diam_  (298.8*mm), // Internal diameter
+  vessel_length_(840.0*mm), // Internal length
+  vessel_thickn_(  3.0*mm),
+  gate_endcap_distance_(-vessel_length_/2. + 92.1*mm),
+  gas_pressure_(15.*bar),
```
jmunozv: I think it is better to set a default value closer to reality (8.30 for instance). I think the maximum pressure that DEMO vessel can hold is 10 bar.

source/geometries/NextDemo.cc
```diff
@@ -1,521 +1,84 @@
...
+    new G4PVPlacement(nullptr, G4ThreeVector(0., 0., 0.),
+                      vessel_geom_->GetLogicalVolume(),
+                      vessel_geom_->GetLogicalVolume()->GetName(),
+                      BaseGeometry::GetLogicalVolume(),
+                      false, 0, false);
+
+  inner_geom_->SetELzCoord(vessel_geom_->GetGateEndcapDistance());
+  inner_geom_->SetMotherLogicalVolume(vessel_geom_->GetGasPhysicalVolume()->GetLogicalVolume());
+  inner_geom_->SetMotherPhysicalVolume(vessel_geom_->GetGasPhysicalVolume());
```
jmunozv, line 77: I think it makes little sense to have `SetMotherPhysicalVolume ` and `SetMotherLogicalVolume ` for the same class. The "Physical" method should be enough and than get the LogicalVolume directly in the class.

Anyway I think this situation occurs in many places, so may be we could fix it in a dedicated PR devoted to this.

source/geometries/NextDemo.cc
```diff
@@ -1,521 +1,84 @@
...
-  else {
-    G4Exception("[NextDemo]", "GenerateVertex()", FatalException,
-                "Unknown vertex generation region!");
-  }
-
-  G4ThreeVector displacement = G4ThreeVector(0., 0., -gate_zpos_in_vessel_);
-  vertex = vertex + displacement;
-
-  return vertex;
+  return G4ThreeVector();
```
jmunozv: You forgot to implement it.

source/geometries/NextDemoVessel.cc
```diff
@@ -1,334 +1,117 @@
...
+  gas_pressure_cmd.SetRange("pressure>0.");
 
-     //   VESSEL      ///////////////////////////////
-      // To avoid overlaps between volumes, the can solid volume
-      // is the result of the union of the port and the CF-300 cylinder.
+  G4GenericMessenger::Command& sc_yield_cmd =
+     msg_->DeclareProperty("sc_yield", sc_yield_,
+                           "Scintillation yield (photons/MeV) of xenon gas.");
+  sc_yield_cmd.SetParameterName("sc_yield", true);
+  sc_yield_cmd.SetUnitCategory("1/Energy");
```
jmunozv, line 59: Define the unit "1/Energy" first.

jmunozv (changes requested):
In addition to single comments, you need to:

- Fix the overlap between VESSEL & GAS
- Modify macro files according to new implementation.
- Modify tests according to new implementation.

source/geometries/NextDemo.cc
jmalbos: I agree, but I have tried to modify as little as possible the existing classes in this PR. As you suggest, this should be addressed in a separate one.

jmalbos:
All comments have been addressed. Tests and macros do not require any change, as far as I can see.

jmunozv:
Add VESSEL visibility and verbosity parameters and use them accordingly, in the same way than in the rest of the geometry sub-systems.

There are some "persistency" and "sensitive detector" tests failing. 

The Vertex generators from ACTIVE and TP_PLATE have stopped working.

jmalbos:
The recent commits fix the issues with the failing tests. It was all connected with the relative position of the vessel with respect to the origin of coordinates.

jmalbos:
> Add VESSEL visibility and verbosity parameters and use them accordingly, in the same way than in the rest of the geometry sub-systems.

Done.

jmunozv:
> > Add VESSEL visibility and verbosity parameters and use them accordingly, in the same way than in the rest of the geometry sub-systems.
> 
> Done.

The verbosity has not been added but I guess it is not strictly necessary.

jmunozv (approved):
This PR implements a simplified version of DEMOpp vessel and main class which is more than enough by the moment.

Tests run succesfully.

I approve this PR.

source/geometries/NextDemo.cc
```diff
@@ -1,521 +1,128 @@
...
-  }
-
-  // VESSEL VERTICES
-  else if (region == "SOURCE_PORT_LATERAL_DISK") {
-    vertex =  source_gen_side_->GenerateVertex("BODY_VOL");
+    vtx = G4ThreeVector(specific_vertex_X_, specific_vertex_Y_, specific_vertex_Z_);
   }
-  else if (region == "SIDEPORT") {
-    vertex = sideport_ext_position_;
+  else if (region == "CALIBRATION_SOURCE") {
```
paolafer, line 112: Is this vertex given in absolute coordinates as well as the `AD_HOC` one? If I understand correctly from the `NextDemoVessel::GenerateVertex()` method, it is not, therefore it should be also displaced, is that correct?

jmalbos: You're right. Fixed!

source/geometries/NextDemoVessel.cc
```diff
@@ -1,334 +1,137 @@
...
+
+  G4GenericMessenger::Command& e_lifetime_cmd =
+    msg_->DeclareProperty("e_lifetime", e_lifetime_, "Electron lifetime in gas.");
+  e_lifetime_cmd.SetParameterName("e_lifetime", false);
+  e_lifetime_cmd.SetUnitCategory("Time");
+  e_lifetime_cmd.SetRange("e_lifetime>0.");
+
+  //G4GenericMessenger::Command& calsrc_position_cmd =
+    msg_->DeclareProperty("calsrc_position", calsrc_position_,
```
paolafer: Is this parameter the z position? I would add it explicitly in the comment, and also add the unit category of length, to avoid errors.

source/geometries/NextDemoVessel.cc
```diff
@@ -1,334 +1,137 @@
...
+                          "Position of calibration with respect to anode endcap.");
+
+  //G4GenericMessenger::Command& calsrc_angle_cmd =
+    msg_->DeclareProperty("calsrc_angle", calsrc_angle_,
```
paolafer: Is it possible to add a unit category for the angle? I tried to use `45 deg` and it was not recognized. Otherwise we have to state explicitly in the comment that the value must be in radians.

paolafer (changes requested):

jmalbos:
> Is this parameter the z position? I would add it explicitly in the comment, and also add the unit category of length, to avoid errors. 

I've changed the name of the parameter to `calsrc_zpos`, amended the comment and added the unit category.

jmalbos:
> Is it possible to add a unit category for the angle? I tried to use `45 deg` and it was not recognized. Otherwise we have to state explicitly in the comment that the value must be in radians. 

I've added the unit following the example B5.

paolafer (approved):
I have no further comments to make, and consider this PR approved.

PR #45, opened 2020-11-02
Fix for EL productions

paolafer:
This PR fixes the position of the anode plate. It needs to be slightly displaced back, because EL scintillation is produced only if the post step volume is gas. The offset is set to 0.1 mm, so it won't affect light distribution significantly.


jmunozv:
I approve this PR that fixes the lack of EL when the ANODE plate was built

jmunozv (approved):

PR #46, opened 2020-11-03
Update DEMO++ EL grids transparencies

jmunozv:
This PR sets the DEMO EL grids transparency to the correct value of 0.83


paolafer (approved):
Straightforward change, based on private communication with people involved in the DEMO assembling. I approve it.

PR #48, opened 2020-11-06
 Set Next100 and FLEX SiPMs detection probability of UV light to 0

jmunozv:
This PR sets the detection probability of generic photosensors used in NEXT100 and FLEX geometries to 0.0 for UV light.

### DIFF
--- a/source/geometries/Next100SiPMBoard.cc
+++ b/source/geometries/Next100SiPMBoard.cc
@@ -184,11 +184,11 @@ void Next100SiPMBoard::Construct()
   // of the new Hamamatsu SiPMs are known.
 
   G4MaterialPropertiesTable* photosensor_mpt = new G4MaterialPropertiesTable();
-  G4double energy[]       = {0.2 * eV, 11.5 * eV};
-  G4double reflectivity[] = {0.0     ,  0.0};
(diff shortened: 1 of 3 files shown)


jmalbos:
I'd make this settable by parameter. After all, there are UV-sensitive sensors.

paolafer:
> I'd make this settable by parameter. After all, there are UV-sensitive sensors.

Either that, or rename the class to Not-So-GenericPhotosensor... just kidding ;-).

jmunozv:
My mistake. The PR description is a little bit misleading.

What this PR really does is to set (using the already implemented method of GenericPhotosensors) the UV light detection probability of NEXT100 and FLEX SiPMs to 0.

Future uses of GenericPhotosensors could, of course, set the detection probability to whatever values one may want.

paolafer:
How did you choose 3.6 eV es a cut-off? I guess it is the end-point of the TPB emission spectrum?

jmunozv:
It's just a first approach energy, well above the visible spectrum of TPB, and well below the xenon UV.

paolafer (approved):
The change in the code is straightforward, tests run, PR is approved.

PR #49, opened 2020-11-11
Fix GenericPhotoSensors to shield photons incoming by window sides

jmunozv:
This PR slightly changes the GenericPhotoSensor geometry, used in NEXT100 and FLEX geometries, to prevent photons to enter the SiPMs by the window lateral faces.


paolafer (approved):
This PR fixes the position of the SiPM window so that photons cannot enter from the sides. The code runs and the tests succeed. It would be good to think about a way of testing this behaviour automatically, since it's not the first time we find this problem, but I reckon it deserves some thoughts. I approve this PR.

PR #50, opened 2020-11-16
Add new generator SingleParticleAngle.

MiryamMV:
Add a new generator (already implemented in PETALO) which allows to define a solid angle for the momentum of the particle.


jmalbos:
I think it would be better to integrate this feature in the `SingleParticleGenerator` rather than creating a new generator that is 95% identical.

paolafer:
> I think it would be better to integrate this feature in the `SingleParticleGenerator` rather than creating a new generator that is 95% identical.

I agree.

CMakeLists.txt
```diff
@@ -7,11 +7,11 @@
 cmake_minimum_required(VERSION 3.0 FATAL_ERROR)
 project(nexus)
 
-IF(COMMAND cmake_policy)
-  IF(${CMAKE_VERSION} VERSION_GREATER "3.8.2")
-    cmake_policy(SET CMP0074 NEW)
-  ENDIF(${CMAKE_VERSION} VERSION_GREATER "3.8.2")
-ENDIF(COMMAND cmake_policy)
+#IF(COMMAND cmake_policy)
```
paolafer: Why do you need to comment this? I remember these lines were added to deal with different versions of cmake.

MiryamMV: It gives me errors while compiling if I don't comment it. Should I reverse that change for the PR?

paolafer: I think we should understand why you get errors now and before you didn't. It should not depend on the new class you added.

MiryamMV: It always gives me errors when I'm in local. I just commited it to be able to switch branch, nothing to do with the new class.

paolafer: You should remove it from the PR, otherwise this change will get included when we approve it. On the other hand, we can investigate why you get this error in your local repository.

source/base/GeneratorFactory.cc
```diff
@@ -41,6 +41,7 @@ GeneratorFactory::~GeneratorFactory()
 #include "ELTableGenerator.h"
 #include "ScintillationGenerator.h"
 #include "ElecPositronPairGenerator.h"
+#include "SingleParticleAngle.h"
```
paolafer: Remove this include, since it's no longer necessary and this class doesn't exist.

source/base/GeneratorFactory.cc
```diff
@@ -69,6 +70,8 @@ G4VPrimaryGenerator* GeneratorFactory::CreateGenerator() const
 
   else if (name_ == "E+E-PAIR")        p = new ElecPositronPairGenerator();
 
+  else if (name_ == "SINGLE_PARTICLE_ANG") p = new SingleParticleAngle();
```
paolafer: Same as above.

paolafer:
With this new feature, the `SingleParticle2PiGenerator` class becomes obsolete. I would delete it and add the exact values that correspond to a forward generation from the DEMO vessel sideport as a comment.

paolafer:
It would be also very useful to add a test that checks that the randomly produced direction is indeed within the specified range. In order to do that in a simple way, within the `catch2` framework, you will first need to extract the code that generates the direction as an individual function (in a similar way to the already existing `RandomEnergy()` function) and then use that function in a test, such as that in `source/tests/utils/BoxPointSamplerTests.cc`. We should create a new file, called RandomUtils, where these functions could live, in such a way that they can be used anywhere in the code. We can talk on zoom, if you need any explanation on how to do that.

source/utils/RandomUtils.h
```diff
@@ -0,0 +1,46 @@
+// ----------------------------------------------------------------------------
+// nexus | Visibilities.h
```
paolafer: Please, change the brief description of the class, accordingly.

source/generators/SingleParticleGenerator.cc
```diff
@@ -146,10 +162,10 @@ void SingleParticleGenerator::GeneratePrimaryVertex(G4Event* event)
...
 
 
-G4double SingleParticleGenerator::RandomEnergy() const
-{
-  if (energy_max_ == energy_min_)
-    return energy_min_;
-  else
-    return (G4UniformRand()*(energy_max_ - energy_min_) + energy_min_);
-}
+//G4double SingleParticleGenerator::RandomEnergy() const
```
paolafer: Delete commented lines, since they are not used anymore.

source/generators/SingleParticleGenerator.cc
```diff
@@ -127,6 +138,11 @@ void SingleParticleGenerator::GeneratePrimaryVertex(G4Event* event)
     px = pmod * momentum_X_/mom_mod;
     py = pmod * momentum_Y_/mom_mod;
     pz = pmod * momentum_Z_/mom_mod;
+  }else if (costheta_min_ != -1. || costheta_max_ != 1. || phi_min_ != 0. || phi_max_ !=2.*pi) {
+    G4ThreeVector p = nexus::Direction(costheta_min_, costheta_max_, phi_min_, phi_max_);
+    px = p.x();
```
paolafer: This way, you always get a momentum with module equal to 1. You need to multiply `p.x()`, `p.y()` and `p.z()` by `pmod`, right?

source/utils/RandomUtils.h
```diff
@@ -0,0 +1,46 @@
...
+  {G4bool dir = false;
+    while (dir == false) {
+      G4double cosTheta  = 2.*G4UniformRand()-1.;
+      if (cosTheta > costheta_min && cosTheta < costheta_max){
+	G4double sinTheta2 = 1. - cosTheta*cosTheta;
+	if( sinTheta2 < 0.)  sinTheta2 = 0.;
+	G4double sinTheta  = std::sqrt(sinTheta2);
+	G4double phi = twopi*G4UniformRand();
+	  if (phi > phi_min && phi < phi_max){
+	    dir = true;
```
paolafer: Since the function returns immediately after, there is no need of setting the `dir` variable to `true`.

paolafer (changes requested):

paolafer:
The test is still missing.

source/tests/utils/RandomUtilsTest.cc
```diff
@@ -0,0 +1,40 @@
...
+  for (G4int i=0; i<20; i++) {
+
+    auto costheta_max = 2*(G4UniformRand()-0.5);
+    auto costheta_min = 2*(G4UniformRand()-0.5);
+    auto phi_max = 2*3.14*G4UniformRand();
+    auto phi_min = 2*3.14*G4UniformRand();
+    G4double x;
+    G4double y;
+    if (costheta_max < costheta_min) {
+      x = costheta_max;
```
paolafer: Have you considered using the std::swap function to perform this operation?

source/tests/utils/RandomUtilsTest.cc
```diff
@@ -0,0 +1,40 @@
+#include <RandomUtils.h>
+#include <Randomize.hh>
+#include "CLHEP/Units/SystemOfUnits.h"
+
+#include <catch.hpp>
+
+TEST_CASE("Direction Function") {
```
paolafer, line 7: This test is failing for me, I attach the error message.

Screenshot 2021-01-19 at 17 34 49

PR #51, opened 2020-11-20
Add rule about pytest cache to .gitignore

jmalbos:


jmunozv (approved):
I approve this simple PR.

PR #53, opened 2020-11-24
Vertex generator for EL tables

jmalbos:
This _pull request_ adds new vertex generators for the EL gaps of three detector geometries (NEXT-100, NEW and FLEX), as part of a new scheme for the generation of EL lookup tables.


jmalbos:
@jmunozv,
Please add to this PR the corresponding commits for the NextFlex classes. 
I'd advice you follow what I've done for the NEXT-100 geometry.

jmunozv (changes requested):
Would it be interesting to assure that Zmax > Zmin to prevent from wrong behaviour ??
Parameters disk_x and disk_y don't need to be >0.
In NEW, if you set a disk_x, it is changed of sign.

jmalbos:
> Would it be interesting to assure that Zmax > Zmin to prevent from wrong behaviour ??
Done.
> Parameters disk_x and disk_y don't need to be >0.
I'm not requiring that, am I?
> In NEW, if you set a disk_x, it is changed of sign.
I don't understand.

PR #27, opened 2020-09-05
Lab rock

andLaing:
Implements a simplified geometry for the walls of HallA at LSC and the option to invoke them for Next-100 and NEW.

PR #54, opened 2020-11-26
Lab generator

andLaing:
Improves the way muons are generated using the angular correlations expected in a lab.

Introduces `ProjectToRegion` methods in the geometries (NEW, NEXT100 and NEXT100Shielding for now) and `GetIntersect` to the `BoxPointSampler` and `CylinderPointSampler2020`.

Tests added to nexus-test

### DIFF
--- a/source/generators/MuonAngleGenerator.cc
+++ b/source/generators/MuonAngleGenerator.cc
@@ -34,7 +34,7 @@ using namespace CLHEP;
 MuonAngleGenerator::MuonAngleGenerator():
   G4VPrimaryGenerator(), msg_(0), particle_definition_(0),
   angular_generation_(true), rPhi_(NULL), energy_min_(0.),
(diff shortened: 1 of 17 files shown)


andLaing:
Added the CylinderPointSampler2020 tests which had not been properly staged

andLaing:
I'd like to integrate the changes in PR #27 into this PR before approval.

andLaing:
There's one of the `if exceptions` in `CylinderPointSampler2020::GetIntersect` that I want to check but other than that I think this is ready for review from my point of view.

paolafer:
I'm a bit confused with the example macros `NEXT100_muons_hallA` and `NEXT100_muons_lsc`. I understand that the former generates muons from outside the lab rock. What about the latter? Maybe you could update the description to point out the difference.

PR #55, opened 2020-11-30
Histo sampler

mmkekic:
This PR introduces new class to sample from Geant4 histograms, allowing nexus to drop root dependency.

### DIFF
--- a/source/tests/utils/HistoTests.cc
+++ b/source/tests/utils/HistoTests.cc
@@ -1,0 +1,56 @@
+#include <HistoSampler.h>
+#include <catch.hpp>
+
+TEST_CASE("HistoSampler") {
+  SECTION("1D sanity check"){
+    auto analysisManager = G4AnalysisManager::Instance();
+    analysisManager->SetFirstHistoId(1);
+    analysisManager->CreateH1("1","hist1", 2, 0., 2.);
+    analysisManager->FillH1(1, 0.5, 1.0);
+    auto h1 = analysisManager->GetH1(1);
+    PDF * h1_pdf = new PDF(h1);
+    for (int i=0; i<10; i++){
+      double x;
+      h1_pdf->sample(x, G4UniformRand());
+      REQUIRE (x <= 1);
+      REQUIRE (x >= 0);
+    }
+    delete h1_pdf;
+  }
+  SECTION("2D sanity check"){
+    auto analysisManager = G4AnalysisManager::Instance();
+    analysisManager->SetFirstHistoId(1);
+    analysisManager->CreateH2("1","hist2", 2, 0., 2., 2, 10., 30.);
+    analysisManager->FillH2(1, 0.5, 10.5, 1.0);
+    auto h2 = analysisManager->GetH2(1);
+    PDF * h2_pdf = new PDF(h2);
+    for (int i=0; i<10; i++){
+      double x, y;
+      h2_pdf->sample(x, y, G4UniformRand(), G4UniformRand());
+      REQUIRE (x <= 1. );
+      REQUIRE (x >= 0. );
+      REQUIRE (y <= 20.);
+      REQUIRE (y >= 10.);
+    }
+    delete h2_pdf;
+  }
+  SECTION("3D sanity check"){
+    auto analysisManager = G4AnalysisManager::Instance();
+    analysisManager->SetFirstHistoId(1);
+    analysisManager->CreateH3("1","hist3", 2, 0., 2., 2, 10., 30., 2, 100., 300.);
+    analysisManager->FillH3(1, 0.5, 10.5, 150., 1.0);
+    auto h3 = analysisManager->GetH3(1);
+    PDF * h3_pdf = new PDF(h3);
+    for (int i=0; i<10; i++){
+      double x, y, z;
+      h3_pdf->sample(x, y, z, G4UniformRand(), G4UniformRand(), G4UniformRand());
+      REQUIRE (x <= 1.  );
+      REQUIRE (x >= 0.  );
+      REQUIRE (y <= 20. );
+      REQUIRE (y >= 10. );
+      REQUIRE (z <= 200.);
+      REQUIRE (z >= 100.);
+    }
+    delete h3_pdf;
+  }
+}
--- a/source/tests/utils/HistoTests.cc
+++ b/source/tests/utils/HistoTests.cc
@@ -8,14 +8,14 @@ TEST_CASE("HistoSampler") {
     analysisManager->CreateH1("1","hist1", 2, 0., 2.);
     analysisManager->FillH1(1, 0.5, 1.0);
     auto h1 = analysisManager->GetH1(1);
-    PDF * h1_pdf = new PDF(h1);
+    std::unique_ptr <HistoSampler1D>  h1_pdf (new HistoSampler1D(h1));
     for (int i=0; i<10; i++){
       double x;
       h1_pdf->sample(x, G4UniformRand());
       REQUIRE (x <= 1);
       REQUIRE (x >= 0);
     }
-    delete h1_pdf;
+
   }
   SECTION("2D sanity check"){
     auto analysisManager = G4AnalysisManager::Instance();
@@ -23,7 +23,7 @@ TEST_CASE("HistoSampler") {
     analysisManager->CreateH2("1","hist2", 2, 0., 2., 2, 10., 30.);
     analysisManager->FillH2(1, 0.5, 10.5, 1.0);
     auto h2 = analysisManager->GetH2(1);
-    PDF * h2_pdf = new PDF(h2);
+    std::unique_ptr <HistoSampler2D>  h2_pdf (new HistoSampler2D(h2));
     for (int i=0; i<10; i++){
       double x, y;
       h2_pdf->sample(x, y, G4UniformRand(), G4UniformRand());
@@ -32,7 +32,6 @@ TEST_CASE("HistoSampler") {
       REQUIRE (y <= 20.);
       REQUIRE (y >= 10.);
     }
-    delete h2_pdf;
   }
   SECTION("3D sanity check"){
     auto analysisManager = G4AnalysisManager::Instance();
@@ -40,7 +39,7 @@ TEST_CASE("HistoSampler") {
     analysisManager->CreateH3("1","hist3", 2, 0., 2., 2, 10., 30., 2, 100., 300.);
     analysisManager->FillH3(1, 0.5, 10.5, 150., 1.0);
     auto h3 = analysisManager->GetH3(1);
-    PDF * h3_pdf = new PDF(h3);
+    std::unique_ptr <HistoSampler3D>  h3_pdf (new HistoSampler3D(h3));
     for (int i=0; i<10; i++){
       double x, y, z;
       h3_pdf->sample(x, y, z, G4UniformRand(), G4UniformRand(), G4UniformRand());
@@ -51,6 +50,5 @@ TEST_CASE("HistoSampler") {
       REQUIRE (z <= 200.);
       REQUIRE (z >= 100.);
     }
-    delete h3_pdf;
   }
 }
--- a/source/utils/HistoSampler.cc
+++ b/source/utils/HistoSampler.cc
@@ -1,0 +1,173 @@
+#include "HistoSampler.h"
+#include <G4Exception.hh>
+#include <G4Voxelizer.hh>
+
+void PDF::create_cdf(){
+  double integral = 0;
+  double sum = 0;
+  cdf.clear();
+  switch(dim){
+  case 1:{
+    nentries = h1G4->all_entries();
+    auto axis = h1G4->axis();
+    nbins = axis.bins();
+    edges_x.reserve(nbins+1);
+    cdf.reserve(nbins+1);
+    size_t i;
+    for (i = 0; i < nbins; i++)
+      {
+	double sig = h1G4->bin_height(i);
+	if (sig<0) G4Exception("", "", FatalException, "Negative entries in hist");
+	integral += sig;
+      }
+    cdf.push_back(0.);
+    for (i = 0; i < nbins; i++)
+      {
+	sum += (h1G4->bin_height(i) /integral);
+	cdf.push_back(sum);
+      }
+    for (i = 0; i < nbins; i++)
+      {
+	edges_x.push_back(axis.bin_lower_edge(i));
+      }
+    edges_x.push_back(axis.bin_upper_edge(nbins-1));
+    break;
+  }
+  case 2:{
+    nentries = h2G4->all_entries();
+    auto axis_x = h2G4->axis_x();
+    auto axis_y = h2G4->axis_y();
+    nbins_x = axis_x.bins();
+    nbins_y = axis_y.bins();
+    nbins = nbins_x*nbins_y;
+    cdf.reserve(nbins+1);
+    edges_x.reserve(nbins_x+1);
+    edges_y.reserve(nbins_y+1);
+    size_t i, j;
+    for (i = 0; i < nbins_x; i++)
+      for (j = 0; j < nbins_y; j++)
+	{
+	  double sig = h2G4->bin_height(i, j);
+	  if (sig<0) G4Exception("", "", FatalException, "Negative entries in hist");
+	  integral += sig;
+	}
+    cdf.push_back(0.);
+    for (i = 0; i < nbins_x; i++)
+      for (j = 0; j < nbins_y; j++)
+	{
+	  sum += (h2G4->bin_height(i, j) /integral);
+	  cdf.push_back(sum);
+	}
+    for (i = 0; i < nbins_x; i++)
+      {
+	edges_x.push_back(axis_x.bin_lower_edge(i));
+      }
+    edges_x.push_back(axis_x.bin_upper_edge(nbins_x-1));
+    for (j = 0; j < nbins_y; j++)
+      {
+	edges_y.push_back(axis_y.bin_lower_edge(j));
+      }
+    edges_y.push_back(axis_y.bin_upper_edge(nbins_y-1));
+    break;
+  }
+  case 3:{
+    nentries = h3G4->all_entries();
+    auto axis_x = h3G4->axis_x();
+    auto axis_y = h3G4->axis_y();
+    auto axis_z = h3G4->axis_z();
+    nbins_x = axis_x.bins();
+    nbins_y = axis_y.bins();
+    nbins_z = axis_z.bins();
+    nbins = nbins_x*nbins_y*nbins_z;
+    cdf.reserve(nbins+1);
+    edges_x.reserve(nbins_x+1);
+    edges_y.reserve(nbins_y+1);
+    edges_z.reserve(nbins_y+1);
+    size_t i, j, k;
+    for (i = 0; i < nbins_x; i++)
+      for (j = 0; j < nbins_y; j++)
+	for (k = 0; k < nbins_z; k++)
+	{
+	  double sig = h3G4->bin_height(i, j, k);
+	  if (sig<0) G4Exception("", "", FatalException, "Negative entries in hist");	  
+	  integral += sig;
+	}
+    cdf.push_back(0.);
+    for (i = 0; i < nbins_x; i++)
+      for (j = 0; j < nbins_y; j++)
+	for (k = 0; k < nbins_y; k++)
+	{
+	  sum += (h3G4->bin_height(i, j, k) /integral);
+	  cdf.push_back(sum);
+	}
+    for (i = 0; i < nbins_x; i++)
+      {
+	edges_x.push_back(axis_x.bin_lower_edge(i));
+      }
+    edges_x.push_back(axis_x.bin_upper_edge(nbins_x-1));
+    for (j = 0; j < nbins_y; j++)
+      {
+	edges_y.push_back(axis_y.bin_lower_edge(j));
+      }
+    edges_y.push_back(axis_y.bin_upper_edge(nbins_y-1));
+    for (k = 0; k < nbins_z; k++)
+      {
+	edges_z.push_back(axis_z.bin_lower_edge(k));
+      }
+    edges_z.push_back(axis_z.bin_upper_edge(nbins_z-1));
+    break;
+  }
+  }
+}
+
+void PDF::sample(double &x, double r){
+  if (dim!= 1){
+    G4Exception("", "", FatalException, "The histogram is not 1D");}
+  if ((r < 0) || (r>1)){
+    G4Exception("", "", FatalException, "Bad input");}
+  if (nentries != h1G4->all_entries()) {create_cdf();}
+  r = (r==1.0) ? 0.0 : r;
+  auto i = G4Voxelizer::BinarySearch(cdf, r);
+  x = edges_x[i];
+  double delta = (r - cdf[i]) / (cdf[i + 1] - cdf[i]);
+  x += delta * (edges_x[i+1]-edges_x[i]);
+}
+
+
+void PDF::sample(double &x, double &y, double r1, double r2){
+  if (dim != 2){
+    G4Exception("", "", FatalException, "The histogram is not 2D");}
+  if ((r1 < 0) || (r1>1) || (r2 < 0) || (r2>1)){
+    G4Exception("", "", FatalException, "Bad input");}
+  if (nentries != h2G4->all_entries()) {create_cdf();}
+  r1 = (r1==1.0) ? 0.0 : r1;
+  r2 = (r2==1.0) ? 0.0 : r2;
+  auto k = G4Voxelizer::BinarySearch(cdf, r1);
+  size_t i = k / nbins_y;
+  size_t j = k - (i * nbins_y);
+  x = edges_x[i];
+  double delta = (r1 - cdf[k]) / (cdf[k + 1] - cdf[k]);
+  x += delta * (edges_x[i+1]-edges_x[i]);
+  y = edges_y[j] + r2 * (edges_y[j+1]-edges_y[j]);
+}
+
+void PDF::sample(double &x, double &y, double &z, double r1, double r2, double r3){
+  if (dim != 3){
+    G4Exception("", "", FatalException, "The histogram is not 3D");}
+  if ((r1 < 0) || (r1>1) || (r2 < 0) || (r2>1) || (r3 < 0) || (r3>1)){
+    G4Exception("", "", FatalException, "Bad input");}
+  if (nentries != h3G4->all_entries()) {create_cdf();}
+  r1 = (r1==1.0) ? 0.0 : r1;
+  r2 = (r2==1.0) ? 0.0 : r2;
+  r3 = (r3==1.0) ? 0.0 : r3;
+
+  auto l = G4Voxelizer::BinarySearch(cdf, r1);
+  size_t i = l / (nbins_z*nbins_x);
+  size_t j = (l/nbins_z)*(1-  (i *nbins_y));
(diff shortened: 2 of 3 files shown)


source/utils/HistoSampler.h
```diff
@@ -0,0 +1,30 @@
+#ifndef HISTO_SAMPLER_H
+#define HISTO_SAMPLER_H
+#include <vector>
+#include <G4RunManager.hh>
+#include <g4csv.hh>
+
+class PDF{
```
jacg: A class called `PDF` in a header called `HistoSampler.h` is an interesting choice.

source/utils/HistoSampler.h
```diff
@@ -0,0 +1,30 @@
...
+  unsigned int nbins    = 0;
+  unsigned int nentries = 0;
+  unsigned int nbins_x  = 0;
+  unsigned int nbins_y  = 0;
+  unsigned int nbins_z  = 0;
+  const tools::histo::h1d * h1G4;
+  const tools::histo::h2d * h2G4;
+  const tools::histo::h3d * h3G4;
```
jacg: `const T *` (which is identical in meaning to `T const *` is a pointer-to-a-constant-T. Don't you mean `T * const` which is a constant-pointer-to-a-T?

source/utils/HistoSampler.h
```diff
@@ -0,0 +1,30 @@
...
+  std::vector<double> cdf;
+  std::vector<double> edges_x;
+  std::vector<double> edges_y;
+  std::vector<double> edges_z;
+  void create_cdf();
+  int dim;
+public:
+  PDF(const tools::histo::h1d * hG4):h1G4(hG4), dim(h1G4->dimension()){};
+  PDF(const tools::histo::h2d * hG4):h2G4(hG4), dim(h2G4->dimension()){};
+  PDF(const tools::histo::h3d * hG4):h3G4(hG4), dim(h3G4->dimension()){};
```
jacg: Each of these constructors leaves 2 (constant!) pointers uninitialized! Dangerous, at best, but I'm surprised that the compiler doesn't warn or even error on uninitialized constants.

jacg: ... ah, no, sorry, I forgot that you declared `T const *` rather than `T * const`: the pointer isn't `const` so the compiler won't complain ... but uninitialized pointers are still a bad idea. (It doesn't much matter whether the C++ standard says that this is one of the situations in which it is zero-initialized: it forces the programmer to waste time wondering about it. Please initialize variables explicitly.)

jacg:
It seems that PDF instances can either be 1-, 2- or 3-D histograms. Whatever dimension they are, they stay that way for their entire lifetime. There are many operations which can only be applied to some specific dimension, so there are lots of run-time checks which throw exceptions if an operation is attempted on an instance with an inappropriate dimension. Logically speaking, the three different histogram dimensions are 3 different types, and they have been forced together into a single type, with the type checks deferred until run-time.

Is there any good reason why these aren't 3 different types? If this were a dynamic language, like Python, this might make sense, but in a statically typed language, play to its strengths, and get all this error checking done at compile time!

source/tests/utils/HistoTests.cc
```diff
@@ -0,0 +1,56 @@
...
+    analysisManager->FillH1(1, 0.5, 1.0);
+    auto h1 = analysisManager->GetH1(1);
+    PDF * h1_pdf = new PDF(h1);
+    for (int i=0; i<10; i++){
+      double x;
+      h1_pdf->sample(x, G4UniformRand());
+      REQUIRE (x <= 1);
+      REQUIRE (x >= 0);
+    }
+    delete h1_pdf;
```
jacg: Unless you are writing *very* low-level code, there should be no `new` or `delete` in your C++ >= 11 code! Use the standard smart pointers.

mmkekic:
> Is there any good reason why these aren't 3 different types? If this were a dynamic language, like Python, this might make sense, but in a statically typed language, play to its strengths, and get all this error checking done at compile time!

Probably not, the (only) 3 types are tight to the 3 types of G4 hence it would probably be cleaner to simply construct 3 different types for the sampler as well. Ironically, I think I was guided by the root implementation where the CDF object does not depend on the dimension and exists in H1 class (and H2 and H3 are derived from H1). 

Looking at the code now it really makes no sense to force them into a single type...

jacg:
> I was guided by the root implementation

There's your problem right there.

Rule of thumb: in software design decisions Root usually shows you how **NOT** to do it.

mmkekic:
> > I was guided by the root implementation
> 
> There's your problem right there.
> 
> Rule of thumb: in software design decisions Root usually shows you how **NOT** to do it.

Seems to be addictive, the purpose of this PR was to remove Root dependency and I ended up copying Root  design (:

source/tests/utils/HistoTests.cc
```diff
@@ -8,22 +8,22 @@ TEST_CASE("HistoSampler") {
     analysisManager->CreateH1("1","hist1", 2, 0., 2.);
     analysisManager->FillH1(1, 0.5, 1.0);
     auto h1 = analysisManager->GetH1(1);
-    PDF * h1_pdf = new PDF(h1);
+    std::unique_ptr <HistoSampler1D>  h1_pdf (new HistoSampler1D(h1));
```
jacg, line 11: It would be better to use `make_unique` instead of the `unique_ptr` constructor ... unless we're stuck in C++11 (the standard committee, in its wisdom, forgot to add it until C++ 14!).

jacg:
Could you give a high-level overview of the problem that we are trying to solve here?

mmkekic:
The general idea is to remove Root which has been currently used to construct (and manipulate/save) and sample from histograms in 1, 2 and 3D. Geant4 does have a histogram class but it does not provide a way to sample from them, hence the goal of this PR is to implement a way to sample from G4 histograms. Note that we still need to actually replace nexus code that uses root with the G4 histogram + this new sampler; this can probably be done in another PR

source/utils/HistoSampler.h
```diff
@@ -0,0 +1,51 @@
+#ifndef HISTO_SAMPLER_H
+#define HISTO_SAMPLER_H
+#include <vector>
+#include <G4RunManager.hh>
```
jmalbos, line 4: Do you need to include this header?

source/utils/HistoSampler.h
```diff
@@ -0,0 +1,51 @@
...
+#include <g4csv.hh>
+
+class HistoSampler1D{
```
jmalbos, line 7: Some conventions we follow:
```
class ClassName
{
public:
    (…)
private:
    (…)
};
```

source/utils/HistoSampler.h
```diff
@@ -0,0 +1,51 @@
...
+  unsigned int nbins    = 0;
```
jmalbos, line 8: Define these in the initialization list.

source/utils/HistoSampler.h
```diff
@@ -0,0 +1,51 @@
...
+  unsigned int nentries = 0;
+  unsigned int nbins_x  = 0;
+  const tools::histo::h1d * h1G4;
+  std::vector<double> cdf;
+  std::vector<double> edges_x;
+  void create_cdf();
+public:
+  HistoSampler1D(const tools::histo::h1d * hG4):h1G4(hG4){};
```
jmalbos, line 16: We avoid inline definitions in the class declaration. Move it to the `.cc` file.

source/utils/HistoSampler.h
```diff
@@ -0,0 +1,51 @@
...
+#include <g4csv.hh>
+
+class HistoSampler1D{
+  unsigned int nbins    = 0;
+  unsigned int nentries = 0;
+  unsigned int nbins_x  = 0;
+  const tools::histo::h1d * h1G4;
+  std::vector<double> cdf;
+  std::vector<double> edges_x;
+  void create_cdf();
```
jmalbos, line 14: We use camel case for function names: `void CreateCDF();`

source/utils/HistoSampler.h
```diff
@@ -0,0 +1,51 @@
...
+public:
+  HistoSampler1D(const tools::histo::h1d * hG4):h1G4(hG4){};
+  void sample(double &x, double r);
```
jmalbos, line 17: Use camel case.

source/utils/HistoSampler.cc
```diff
@@ -0,0 +1,167 @@
...
+  double integral = 0;
+  double sum = 0;
+  cdf.clear();
+  nentries = h1G4->all_entries();
+  auto axis = h1G4->axis();
+  nbins = axis.bins();
+  edges_x.reserve(nbins+1);
+  cdf.reserve(nbins+1);
+  size_t i;
+  for (i = 0; i < nbins; i++)
```
jmalbos, line 15:
```
for () {
    DoStuff();
}
```

source/utils/HistoSampler.h
jacg: Why?

This is the archetypical example of a function that should be defined in the header!

source/utils/HistoSampler.h
jacg: ... with *braced* initializers.

jacg:
I'm sorry @mmkekic I should have been clearer about what I meant. Could you please provide a description of the feature/algorithm that is needed, without making any reference to C++, Root, Geant4 or any other circumstantial implementation details which pollute the high-level signal. Well, maybe you need to say what Geant4 provides that is useful in this context, but probably not on the first iteration.

source/utils/HistoSampler.h
jmalbos: We do define them in the header, but out of the class declaration and with an explicit `inline` to avoid mixing declarations and definitions. An example: https://github.com/next-exp/nexus/blob/master/source/base/Trajectory.h

jmalbos:
> I'm sorry @mmkekic I should have been clearer about what I meant. Could you please provide a description of the feature/algorithm that is needed, without making any reference to C++, Root, Geant4 or any other circumstantial implementation details which pollute the high-level signal. Well, maybe you need to say what Geant4 provides that is useful in this context, but probably not on the first iteration.

Generation of pseudo-random numbers following an arbitrary pdf.

source/utils/HistoSampler.h
jacg: Apologies, I meant in-class.

I now understand that Nexus style prescribes

``` c++
namespace nexus {
  class Foo
  {
  public:
    Foo(Bartype b);
    Bartype GetBar() const;
    void SetBar(Bartype b);
  private:
    Bartype bar_;
  };
}

inline nexus::Foo::Foo(Bartype bar) : bar_(bar)
{}

inline Bartype nexus::Foo::GetBar() const
{ return bar_; }

inline void nexus::Foo::SetBar(Bartype bar)
{ bar_ = bar; }
```

and echews

``` c++
namespace nexus {
  class Foo
  {
  public:
    Foo(Bartype bar) : bar_(bar) {}
    Bartype GetBar() const { return bar_ }
    void SetBar(Bartype bar) { bar_ = bar; }
  private:
    Bartype bar_;
  };
}
```

I would like to understand the reason for having adopted this style.

I can go on for hours about why I think it's a bad choice of style, but, to be brief: it reduces the signal-to-noise ratio of the code and violates the DRY principle.

Now, if that's the official style of Nexus, and that style is set in stone, then we should all stick to it. However I have observed poor coding style being adopted in HEP codes (both officially and randomly) for over 2 decades, and am morally bound to try to fix this. Hence my question: is there some logic supporting adoption of this style, or are we simply following some misguided style guide document that circulated around HEP C++ programmers around the turn of the millennium?

[Incidentally, the style guide in the Wiki, erroneously refers to members as globals in the discussion of the trailing-underscore convention.]

jacg:
> Generation of pseudo-random numbers following an arbitrary pdf.

OK, that's a bit too high-level, but it's a good start.

Would it be correct to say that you want to 

1. take a histogram (1, 2 or 3D)
2. infer a PDF from said histogram
3. generate pseudo-random numbers that obey said PDF

?

source/utils/HistoSampler.h
jmalbos: Nothing is set in stone, and we are happy to adapt the style if there are **good** reasons (as it would involve revising many classes). Can you share a link to what you'd consider a non-misguided C++ style guide?

jmalbos:
> Would it be correct to say that you want to
> 
> take a histogram (1, 2 or 3D)
> infer a PDF from said histogram
> generate pseudo-random numbers that obey said PDF
> ?
> 

Yes, that's correct. In practice, @mmkekic is using [inverse transform sampling](https://en.wikipedia.org/wiki/Inverse_transform_sampling) with a binned distribution (the normalized histogram) to do all of that.

source/utils/HistoSampler.h
jacg: In the hypothetical situation that we adopt a change in style, I don't think it would be a good idea to rewrite existing code, en masse, to adapt it to the new style. Rather, new code should be written in the new style, and old code that is being rewritten for some other purpose *might* be updated to the new style in the process.

As for pointing to existing C++ style guides, there will be almost as many mutually incompatible style guides as there are projects. The language is so huge that each of these documents will be enormous (or woefully incomplete), and reading them from cover to cover will not be a good use of our time. What's more, the choices made on any single topic in any particular environment or project may make sense there, but not elsewhere. Rather than following others' guides, I recommend *reasoning* about any given style consideration. Referring to others' guides for inspiration on any given topic would make sense, but I certainly don't believe that there is any single external guide that should serve as our guide.

As for this particular topic (in-class vs out-of-class inlined member functions), a quick (non-exhaustive, flawed, unreliable) search came up with one pronouncement against this style:

+ https://kks32.github.io/cpp-style-guide/?showone=Namespaces#Inline_Functions

  For example, accessors and mutators should certainly be inside a class definition.
  
and zero in its favour. 

But, resorting to reason rather than the pronouncements of others, to me it is clear that (looking at the examples I provided above) the sample which is half the length of the other, and avoids repeating the same information in two places without any negative impact on code performance, is obviously the better. I am open to having flaws in my argument pointed out to me.

source/utils/HistoSampler.cc
jacg: So, in class definitions the brace goes on its own line, but in for-loops the brace goes on the end of the introduction line?

jacg:
Yesterday I could see 12 forks of this repo. Today GitHub still claims that there are 12, but only shows me my own. I don't know whether this is related to the fact than clicking on the link to `mmkekic:histo_sampler` gives me a 404, and that I cannot add `mmkekic` as a remote in my repo. I don't understand what is going on, but it makes reviewing this code a huge PITA!

Anyway, as I see it (forgetting about multiple dimensions, for now) we need (roughly speaking) a function which takes a histogram as input, and returns a random number generator that can be used to generate random numbers distributed according to the PDF implied by the histogram. Taking this at face value, I would expect an interface like this:

```
1. histogram -> RNG
2. 1. RNG() -> Real
   2. RNG() -> Real
   3. RNG() -> Real
   4. RNG() -> Real
   5. RNG() -> Real
   6. etc.
   7. etc.
```

Or maybe this:

```
1. histogram -> RNG
2. RNG(N) -> Container with N Reals
```

`RNG` requires some nontrivial internal state, so it's natural to implement it as a class in C++. However, in this PR

1. Why is `create_cdf` a separate step from construction of the RNG from the histogram? What possible use could there be for this intermediate object? One of the principles that the designers of C++ got right is that a client should not be able to get his hands on an object in an improperly configured state, and that is why they came up with constructors, and that is why constructors are called automatically (for user-defined types). AFAICT, during the time between construction of `HistoSamplerND` and calling `create_cdf` the object is in a state that violates its class invariants! Thus it violates one of the truly useful principles of C++ programming.

2. I am completely perplexed by the use of mutable references to floats in the interface of `sample`.

jjgomezcadenas:
Sorry to intrude here, perhaps missing the point. 

Is this code (generation of random numbers according to a histogram) something we need to get rid of root? if yes (or in any case) may I suggest to look at GSL:

https://www.gnu.org/software/gsl/doc/html/histogram.html

This wheel is, a one could expect, invented. Dressing GSL C functions should not be difficult and accepting GSL dependence o Nexus appears a reasonable thing to do, given the fact that it provides a full scientific library in C.

jacg:
> Sorry to intrude here, perhaps missing the point.

... no, to me it looks like you've got the point ...

> This wheel is, a one could expect, invented. 

Thank you for this: I was (as I'm sure *you* can guess) looking for this wheel in the background of this exchange. First I needed to understand exactly what kind of wheel it was, and I hadn't found anything yet, but it looks like you've hit the nail on the head: https://www.gnu.org/software/gsl/doc/html/histogram.html#resampling-from-histograms

jacg:
... unfortunately it only seems to go up to 2D.

mmkekic:
> 1. Why is create_cdf a separate step from construction of the RNG from the histogram? What possible use could there be for this intermediate object? One of the principles that the designers of C++ got right is that a client should not be able to get his hands on an object in an improperly configured state, and that is why they came up with constructors, and that is why constructors are called automatically (for user-defined types). AFAICT, during the time between construction of HistoSamplerND and calling create_cdf the object is in a state that violates its class invariants! Thus it violates one of the truly useful principles of C++ programming.
> 

`create_cdf`is called inside `sample`method if the number of elements in the histogram were changed.
 I guess it can be  called in the constructor as well (and `sample` still calls it again if the number of elements inside the histogram have changed). The idea is that one could create a histogram, sample some numbers, update the G4 histogram and use the same HistoSample instance to sample some more numbers. Now, it only checks for the change of the total number of elements; I think a user could add and remove some elements leaving the total number unchanged and the `create_cdf` would not be recalculated,  so there is a room for error (the condition on how to check if a recalculation of cdf is needed is again copied from root)
 
> 2. I am completely perplexed by the use of mutable references to floats in the interface of sample.

That is not a definite choice, @jmalbos was suggesting to use G4TwoVector and G4ThreeVector as return type in case of 2 and 3 dimensions (since the number of returned floats depends on the dimension of the problem). What would you suggest here?


> Is this code (generation of random numbers according to a histogram) something we need to get rid of root? if yes (or in any case) may I suggest to look at GSL:

GSL was indeed our first choice, especially since it is a dependency of nexus, but the problem is that it does not support  3D histograms. The other library that does have histograms seems to be boost, but it seemed an overkill and would introduce a new dependency

jmalbos:
> This wheel is, a one could expect, invented. Dressing GSL C functions should not be difficult and accepting GSL dependence o Nexus appears a reasonable thing to do, given the fact that it provides a full scientific library in C.

In fact, nexus has GSL as a dependency **already** due to a single use in an obscure piece of code translated from Fortran. As explained above by @mmkekic, we did test it and dropped it. I honestly think this problem, the histogram sampling, is simple enough to have our class with 5 lines of code doing the job. Notice that we don't need a random generator (Geant4 has its own), only a class that takes a histogram, normalizes it and calculates the cdf.

jacg:
> `create_cdf` is called inside samplemethod if the number of elements in the histogram were changed.

Hmm, if the number of elements in the histogram changes, the PDF changes, and a new object should be made that corresponds to the new PDF. This scheme silently and implicitly changes the PDF if something changes somewhere else in the program. A classical case of mutable shared state which leads to pain and suffering. Don't do it!

> I guess it can be called in the constructor as well (and sample still calls it again if the number of elements inside the histogram have changed). 

Nope. If the histogram changes, and you want the PDF corresponding to the new histogram, create a new object corresponding to the new PDF. Each of these PDF objects should be immutable (save for, perhaps, internal RNG state), so that it continues to behave the same way throughout its lifetime.

> The idea is that one could create a histogram, sample some numbers, update the G4 histogram and use the same HistoSample instance to sample some more numbers. 

Mutable shared state. Just say "No!".

You'd be buying a tiny bit of (debatable) convenience for a whole load of confusion and pain.

>> I am completely perplexed by the use of mutable references to floats in the interface of sample.

> GSL was indeed our first choice, especially since it is a dependency of nexus, but the problem is that it does not support 3D histograms.

But it does support 1D and 2D. So, faced with the choice of using a pre-existing, tried and tested solution for 1D and 2D and winging it for 3D on the one hand, and winging it for all 3 dimensions on the other, we chose the latter? Interesting choice.

In an ideal world, we'd look at the 1D and 2D source that GLS provides, and contribute the 3D version back to the community by submitting it to GSL. But this isn't an ideal world.

> The other library that does have histograms seems to be boost, but it seemed an overkill and would introduce a new dependency

I recommend avoiding Boost to preserve your sanity. Boost is *really* great ... until you make a compile time error. Then it's game over.


> That is not a definite choice, @jmalbos was suggesting to use G4TwoVector and G4ThreeVector as return type in case of 2 and 3 dimensions (since the number of returned floats depends on the dimension of the problem). What would you suggest here?

Modern C++ tuples are becoming more and more bearable/usable. Are we stuck in C++ 11?

jacg:
> I honestly think this problem, the histogram sampling, is simple enough to have our class with 5 lines of code doing the job.

In that case I'll reserve judgement until I see the 5-line version of the code :-)

source/utils/HistoSampler.h
jmalbos: > As for this particular topic (in-class vs out-of-class inlined member functions), a quick (non-exhaustive, flawed, unreliable) search came up with one pronouncement against this style:
> 
> https://kks32.github.io/cpp-style-guide/?showone=Namespaces#Inline_Functions
> 
> For example, accessors and mutators should certainly be inside a class definition.
> 
> and zero in its favour.

Just for the sake of the argument: we have always tried to follow the advice at isocpp.org. This is what they say about inline definitions: https://isocpp.org/wiki/faq/inline-functions#inline-member-fns-more

jacg: The argument seems to be

> it is harder on all the readers since it mixes what a class does (the external behavior) with how it does it (the implementation).

This sounds reasonable until you illustrate it with an example. The claim is that 

``` c++
namespace nexus {
  class Foo
  {
  public:
    Foo(Bartype b);
    Bartype GetBar() const;
    void SetBar(Bartype b);
  private:
    Bartype bar_;
  };
}
```

is clear to all readers, while extending it to

``` c++
namespace nexus {
  class Foo
  {
  public:
    Foo(Bartype bar) : bar_(bar) {}
    Bartype GetBar() const { return bar_ }
    void SetBar(Bartype bar) { bar_ = bar; }
  private:
    Bartype bar_;
  };
}
```

makes it sufficiently 'harder on all the readers' for it to be worth talking about.

Furthermore, the claim is that adding all the following junk to the bottom of the former (with its concomitant burden of DRY-violation), is *better* than turning the former into the latter.

``` c++
inline nexus::Foo::Foo(Bartype bar) : bar_(bar)
{}

inline Bartype nexus::Foo::GetBar() const
{ return bar_; }

inline void nexus::Foo::SetBar(Bartype bar)
{ bar_ = bar; }
```

Really?

Perhaps a more realistic example would be a typical HEP class with some 39 member functions, approximately 12 of which will be trivial. The claim then becomes that annotating (with half-line-long definitions) 12 of the 39 member function declarations in the class definition, hampers the understanding of the class structure.

Please understand, I spend a loooooot of my time thinking about clarity of code, and making code as legible as possible. I find this argument completely spurious. Why? Because we are talking about functions with *trivial* implementations, and those trivial implementations do not add a significant burden to the understanding of the declarations. On the other hand, there is the cost of DRY-violation. Furthermore it is debatable whether the clarity gained outweighs the burden of having to look elsewhere when you are interested in the implementation.

However, if you are convinced by or attached to this style, let's just stick to it. We have better things to do with our resources than debate this minor point.

jmalbos: No, I'm not convinced by or attached to that style. I was simply bringing to the discussion the reference I had followed historically. I'm not questioning your expertise and I appreciate the advice, but, in general, I like to understand the reasons behind guidelines. Otherwise, they just become _arguments from authority_.

jacg: > I like to understand the reasons behind guidelines. Otherwise, they just become *arguments from authority*.

A highly commendable philosophy to which I subscribe.

> I'm not questioning your expertise

Why not? You should! I don't mean randomly or just for the sake of being contrarian, but you absolutely *should* question *why* I believe something, and I should be able to give you a reasonable answer. As I hope I have done in this case. And if you don't find my answer in this case convincing, then let's just stick to the old style, because this particular case is not important enough to merit any further discussion.

mmkekic:
> In that case I'll reserve judgement until I see the 5-line version of the code :-)

Even G4-> gsl histogram conversion is more than 5 lines (at least my implementation). Maybe adding a test to compare gsl and new-custom-class result for 1 and 2d case? That would also allow to easily drop gsl dependency, if we decide so,  since, as @jmalbos  pointed out, gsl is not extensively used in nexus.

jacg:
> Even G4-> gsl histogram conversion is more than 5 lines (at least my implementation).

... my point being that this is *never* going to be done in 5 lines. As such, it is a wheel that you should avoid reinventing unless there are some very good reasons to do so.

mmkekic:
I meant that even a HistoSampler class with similar interface : takes G4 histogram in a constructor and has a `sample`method, but internally calls gsl, would need to create and fill gsl histogram, and then create the pdf and sample from there, all together being more than 5 lines

jacg:
Yes, I understand.

It doesn't matter which of the following paths you take:

1. write an adapter around some existing, tried, tested and optimized functionality written and maintained by somebody else;

2. write the functionality yourself, thereby assuming the concomitant burden of testing, optimizing and maintaining it for evermore;

it's not going to be done in 5 lines.

The argument put forward by @jmalbos in favour of option 2, was that it's only 5 lines.

jjgomezcadenas:
Here is how I see things:

1. GSL dependency should not be a burden. On the contrary, I would claim that NEXUS should link with GSL just in case (It is one of the best scientific libraries in the market, is free, is C, is easy to include in the path, is maintained, is documented, is open software and ah, yes, we have contributed to it ——se tensor class—). 

2. As far as I know, our uses cases are 1D and 2D histograms. We have no user case for 3D histograms so far, the day we have it, we could think how to implement them. 

3. If our uses cases are 1D and 2D histograms, a problem solved by someone else, I would not reinvent the wheel. I would favour, actually, writing our own (limited) implementation of histograms ***as a test*** of our wrap-up of GSL histograms. 

4. Here is an “old boy” tale. Back in the day (~20 years ago), Jose Angel and I were starting to program python, and found that there was no obvious histograms available at the time (this was the pre-numpy, pre-scipy, pre-conda, pre-pandas, pre-almost-everything, but, alas, not pre-root). So we decided we would write our own “five line” histogram classes, which we call “hippies”. Many months later, we were not done yet, and our initial effort had branches into “hoppies”, “huppies” and “happies”, all of them different implementations (with a similar level of messiness). I understand that the problem here is a bit different. but the lesson I took from hippies and hopies was, never do it yourself if somebody has done it for you.  

5. So, definitively, I would go for a GSL wrap-up (we may end up writing a few of these if we get rid of root).

BUT, in any case: Please let’s take a decision and let’s move forward. 

Just my five cents

jacg:
You, of course, already know my strong opposition to reinventing the wheel, especially if it's going to be done badly (and we end up with one of those lovely square wheels that tend to emerge in these situations). So you know that I agree with most of the above.  However, I strongly disagree with 

> Please let’s take a decision and let’s move forward.

This is one of those rare and delightful situations where there is absolutely no time pressure. The functionality is already available (IIUC) via our current dependence on Root. No vital production runs will be delayed if we take the time to think about this carefully. What's more, there is no point in doing this at all, if it is going to be done badly!

We have all the time in the world to get this right. So let's get it right, rather than rushing in a crappy solution.

jjgomezcadenas:
OK, agreed, it is true that the problem is “solved” (via ROOT), so we can take the time to come up with a better solution.

PR #56, opened 2020-12-07
Fix condition to save optical photons.

paolafer:
The current code has been found not to work as expected, meaning that the condition to find the `OPTICAL` tracking option in the macro file was never met. This PR fixes this bug.


paolafer:
This fix is no longer needed.

PR #58, opened 2021-01-08
Use pytest-order instead of pytest-ordering2.

paolafer:
The pytest-order package replaces pytest-ordering2, which was a
temporary fix. The syntax of the code has been updated accordingly.


gonzaponte (approved):
The temporary solution has been replaced with the long term, stable solution. The syntax changes slightly and the code has been modified accordingly.

PR #59, opened 2021-01-11
Demopp refinement

jmunozv:
This PR makes some refinements to DEMOpp.

* It adds the possibility of defining different transparencies for ANODE and GATE.
* It unifies the different config parameters from the TP and FC into a single one.
* It adds the Optical Material Properties degradedTPB, which aims to simulate a posible degradation of PLATE coatings in NEXT detectors, and that has shown promising preliminary results.

Macro files have been updated accordingly.

### DIFF
--- a/macros/DEMOPP_fullKr.config.mac
+++ b/macros/DEMOPP_fullKr.config.mac
@@ -14,6 +14,7 @@
 
 
 # GEOMETRY
+/Geometry/NextDemo/config run5
 /Geometry/NextDemo/elfield true
 /Geometry/NextDemo/pressure 8.5 bar
 /Geometry/NextDemo/max_step_size 1. mm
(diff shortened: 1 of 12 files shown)


paolafer:
A minor comment on the visibility of the quartz plate or the grid: I think it could be useful to draw them even when the field cage visibility is zero (only the shape, like the copper support of the PMTs) since it is something that characterises the geometry. Just a suggestion.

paolafer (approved):
These changes are clear and improve the organization of the code. The updated macros run correctly. I have no further comments, beyond the one on visualization, therefore I consider this PR approved.

jmunozv:
Respect to the comment on the visualization, I guess it is a matter of each one's experience.
Mine is that many times I've had to go through a painful path of zooms, pans and rotations to view the details of small parts of the detector that I was developing just because some default visualization of the biggest parts of the detector made my components so small that was really imposible to see them in detail.

PR #61, opened 2021-01-13
Update xenon EL-yield to manage pressure dependant values

jmunozv:
This PR modifies the xenon EL-yield formula to manage the different pressure dependant values.

As the way in which the yield varies is unknown, it has been implemented in a simple way in which given a pressure it is assigned the constants of the immediately lower known pressure.

The constants for the different pressures come from:
Freitas-2010  // Physics Letters B 684 (2010) 205–210

(Closed to be re-opened as #74 due to the repo change from private to public)

### DIFF
--- a/source/materials/XenonGasProperties.cc
+++ b/source/materials/XenonGasProperties.cc
@@ -212,12 +212,23 @@ G4double XenonGasProperties::ELLightYield(G4double field_strength) const
 
   // Y/x = (a E/p - b) p,
   // where Y/x is the number of photons per unit lenght (cm),
-  // E is the electric field strength, p is the pressure, and a and b
-  // are empirically determined constants:
+  // E is the electric field strength, p is the pressure
+  // a and b are empirically determined constants depending on pressure.
+  // Values for different pressures are found in: Freitas-2010
+  // Physics Letters B 684 (2010) 205–210
 
-  const G4double a = 140. / kilovolt;
   const G4double b = 116. / (bar*cm);
+  G4double a = 140. / kilovolt;
 
+  // Updating the slope
+  if (pressure_ >= 2. * bar) a = 141. / kilovolt;
+  if (pressure_ >= 3. * bar) a = 141. / kilovolt;
+  if (pressure_ >= 4. * bar) a = 142. / kilovolt;
+  if (pressure_ >= 5. * bar) a = 151. / kilovolt;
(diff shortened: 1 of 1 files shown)


source/materials/XenonGasProperties.cc
```diff
@@ -212,12 +212,23 @@ G4double XenonGasProperties::ELLightYield(G4double field_strength) const
...
+  // Values for different pressures are found in: Freitas-2010
+  // Physics Letters B 684 (2010) 205–210
 
-  const G4double a = 140. / kilovolt;
   const G4double b = 116. / (bar*cm);
+  G4double a = 140. / kilovolt;
 
+  // Updating the slope
+  if (pressure_ >= 2. * bar) a = 141. / kilovolt;
+  if (pressure_ >= 3. * bar) a = 141. / kilovolt;
```
paolafer, line 225: For 2 and 3 bar the constant is the same. Why don't you drop the line with 3 bar?

jmunozv: I was trying to replicate all the values that appeared in Freitas paper (the one we are following), and the values for both pressures appeared in the list. I thought it was a good idea to maintain all the values although they could be the same.

But I don't have a strong opinion on this, so we can delete it if you think it's better.

paolafer: It's a superfluous line of code, I think it's better to drop it.

jmunozv: Done

jacg: There's something to be said for directly reflecting in the code, the structure of a paper that the code follows. But, in that case, it's vital that you add comments explaining why the code is doing something that appears odd, lest readers waste valuable time and effort on wondering why you are doing something apparently pointless.

jmunozv: I fully agree. I thought my comments at the beginning of the function were enough.

jacg: Ah, sorry, those comments were not visible in the little extract that GitHub showed me.

jacg:
I guess that you're closing and reopening PRs because of your fork getting unlinked when the repo went public:

1. At the very least you should leave comments in each pair of PRs, referring to the other, and explaining why it was closed/re-opened.
2. As suggested [here](https://stackoverflow.com/a/51502462/4122880), GitHub support should be able to re-link the forks, which would remove the need to go around closing existing PRs and opening new ones.

PR #62, opened 2021-01-13
black_box

MiryamMV:
This PR will add the geometries for the black box. The particle generator will be updated when the PR for SingleParticleGenerator is finished.


jmunozv:
I think we should implement all the possibilities of DiceBoards into a single class, and specify via configuration parameters the shape and dimensions of it. Moreover you should re-use the current implementation of NextDemoSiPMBoard that already handles the setup of all the main config parameters via methods.

Another general comment, I think you are using leading-underscore names for private vars, while the nexus convention is trailing-underscores.

jmunozv (changes requested):
I think all possible DiceBoards geometries must be implemented into a single class, and specify via configuration parameters the shape and dimensions of it. Moreover you should re-use the current implementation of NextDemoSiPMBoard that already handles the setup of all the main config parameters via methods.

Another general comment, I think you are using leading-underscore names for private vars, while the nexus convention is trailing-underscores.

MiryamMV:
Due to messy history I'd rather close this PR and open a new one.

PR #64, opened 2021-01-20
Enable installation via Nix

jacg:
# Trivial installation process for Nexus

This PR enables trivial installation of Nexus and all its dependencies on Nix-enabled Linux machines.

That is, on a Linux machine with Nix and direnv, the *complete* installation and testing process becomes:

``` shell
git checkout https://github.com/next-exp/nexus.git
cd nexus
direnv allow  # ... or `nix-shell` if direnv is not installed
scons
bin/nexus-test
pytest
```

In other words, a huge improvement on the complex and fiddly process described [here](https://github.com/next-exp/nexus/wiki/Installing-and-running-nexus).

# OMG, does this really work?

A GitHub action run, which verifies this process can be found [here](https://github.com/jacg/nexus/runs/1734855849?check_suite_focus=true).

I haven't added the action to this PR, as this is a private repo, and therefore has a limit of 2000 GHA minutes per month. We would need to do some calculations before adding a GHA.

# MacOS

Unfortunately, Geant4 is currently not supported on Nix/Darwin, which means that this process only works on Linux, as it stands.

In principle, there shouldn't be any fundamental reason why Geant4 cannot be supported by Nix. If any Mac users express an interest in having this, I can have a look into it.

# There is no downside

This PR adds a grand total of 2382 bytes to the repository. Those unable or unwilling to install Nix can continue to do whatever they were doing before, without any interference from this PR.


jmalbos:
Notice that nexus can be built with both SCons and CMake, and both are actually in use (yet another chasm). Personally, I favor the latter because, among other reasons, it is better integrated with Geant4 (and ROOT, although we will eventually remove this dependency).

jacg:
> Notice that nexus can be built with both SCons and CMake

Well, this process has been tested with SCons. The Nix part essentially provides all the dependencies and sets required environment variables. The build of Nexus itself is left to the Nexus tooling. If CMake normally works, I don't see why it shouldn't work on top of this too. Forgive me for not bothering to test it: one convenient way to get Nexus running is enough for me for now. If anyone cares about making sure CMake works with this, I'd be happy to assist.

>  ROOT, although we will eventually remove this dependency

Remove Root as a dependency? What can I do to help to make this happen sooner rather than later?

paolafer:
> In other words, a huge improvement on the complex and fiddly process described here.

That would be great! 

Regarding the use of Nix, yesterday the topic was brought up in the technical meeting and it was proposed to have a discussion dedicated to that with you, right @mmkekic?

paolafer:
> Remove Root as a dependency? What can I do to help to make this happen sooner rather than later?

PR #55 is work in progress to get rid of histogram sampling, which is one of the reasons why we use ROOT. I've also a branch open where I use G4 tools to save csv files which can be analyzed in a second moment to replace the generation of control histograms (the second reason we use ROOT). We need to put all together and we'll be ROOT-free forever.

jacg:
> Regarding the use of Nix, yesterday the topic was brought up in the technical meeting and it was proposed to have a discussion dedicated to that with you

Fine. But please note, that, unlike the Nix PR in IC (which proposes to *replace* `manage.sh` with Nix, and therefore requires that *everyone* be willing and able to use Nix *everywhere* and thus needs careful thought before accepting) this PR removes absolutely nothing and breaks absolutely nothing: it's pure (huge) benefit in those situations where Nix is available. [i.e. There's not much to think about here, no caveats, no downsides ...]

Of course, IC's `manage.sh` (and its Nix PR) aim for absolutely reliable reproducibility of software configurations and therefore results. It seems that here, any version of the dependencies which seems to work is deemed acceptable, which makes the job here much easier. But I would think that reliable reproducibility is desirable here too!

paolafer:
> Fine. But please note, that, unlike the Nix PR in IC (which proposes to replace manage.sh with Nix, and therefore requires that everyone be willing and able to use Nix everywhere and thus needs careful thought before accepting) this PR removes absolutely nothing and breaks absolutely nothing: it's pure (huge) benefit in those situations where Nix is available. [i.e. There's not much to think about here, no caveats, no downsides ...]

Yes, I see that and I like it :-).

paolafer:
It would be useful to specify the version of the packages, at least GEANT4 and the python packages used in the tests. Is it possible to do that in Nix?

jacg:
> It would be useful to specify the version of the packages, at least GEANT4 and the python packages used in the tests. Is it possible to do that in Nix?

To cut a very long story short: you probably don't want to do that!

The precise versions that are being used are nailed down by the hash used as `nixpkgs-commit-id`, so they certainly won't change without your permission: you have to change that hash to allow even a single bit (as in binary digit) of your entire installation to change!

For some packages, Nix provides alternative versions (e.g. Python 3.7, 3.8, 3.9; gcc 4.8, 4.9, 6, 7, 8, 9, 10; etc.) for you to pick from, but for most things you should be satisfied with the versions that come together in that particular version of `nixpkgs`.

If you *really* want to use a different version of some package from the one that comes in that very `nixpkgs` commit, that can be achieved with overlays. This will sometimes be trivial (if the dependencies and build procedure of the package haven't changed) but can range through all levels of difficulty up to essentially impossible. Broadly speaking, if you have a good reason to use some specific-but-different version of a package, and it isn't fundamentally incompatible with other packages present in that version of `nixpkgs`, then it can usually be done. But Nix doesn't provide alternative versions out of the box, for the majority of things. The obvious exceptions are language or compiler versions. Root is available in this PR in versions 6.18.04 and 5.34.38. 

The Geant4 version in this PR is 10.6.2. If you think it's important, I can have a look at switching that to some other version. (Which one?)

But let me reiterate: the packages that you get by default at any one time are 100% guaranteed (well, if you trust cryptographic hash functions) to be stable (as in: won't evolve without your permission), and often what you care about is stability and interoperability, rather than having some random but very specific version of a package.

paolafer:
We usually want to use a specific version of GEANT4 for compatibility issues and to be sure it doesn't introduce bugs (it did in the past with some versions). We also decided to fix the versions of the python packages used in the tests to prevent that some changes in the packages make some tests fail, do you think it's not needed?

jacg:
>> It would be useful to specify the version of the packages, at least GEANT4 [...]

> [...] that can be achieved with overlays.

As a proof of concept, I've added a commit to this PR, which uses an overlay to change the Geant4 version from 10.6.2 to 10.6.3.


Here is the [GitHub Action output](https://github.com/jacg/nexus/runs/1736887403?check_suite_focus=true).

The downside is that this random version of Geant4 is not present in the default Nix binary cache, so it ends up compiling Geant4 locally (70 minutes) rather than downloading the binary from the cache (< 1 minute). Geant4 is an extreme example: for smaller packages the build time will probably not be an issue. If build times of our very specific package versions become an issue, we have the option of setting up our own binary cache.


> We usually want to use a specific version of GEANT4 for compatibility issues and to be sure it doesn't introduce bugs (it did in the past with some versions). We also decided to fix the versions of the python packages used in the tests to prevent that some changes in the packages make some tests fail, do you think it's not needed?

You need to fix various package versions, for the most part not because you want some very specific version, but because 

1. you don't want the version to change without you explicitly giving permission,

2. you want to be able to run exactly the same version that was originally used, if you ever need to check out an old commit to verify or compare some results.

As I have already pointed out, the pinning of the `nixpgs` version via the hash assigned to `nixpkgs-commit-id` satisfies both of these requirements.

In the very rare, but possible, cases where you *really* want some very specific version of some package that deviates from what is available in any given `nixpkgs` version, the overlay approach shown above can do the job.

> do you think it's not needed?

I think that specific-versions-that-deviate-from-an-available-version-of-nixpkgs are needed *extremely* rarely, and when this occurs, an overlay can be used.

jmalbos:
@jacg We _really_ want to be able to pick the version of Geant4 we're using, as the simulated physics has changed sometimes dramatically even between minor versions. I'd say that is a requirement for the installer.

jacg:
> We *really* want to be able to pick the version of Geant4 we're using, as the simulated physics has changed sometimes dramatically even between minor versions. I'd say that is a requirement for the installer.

In which case I find it remarkable that the [Geant4 installation](https://github.com/nextsw/nexus/wiki/GEANT4-installation) section on the Wiki fails to mention this. There, 'the latest distribution' is suggested! (Whilst we're on the Wiki, there are also numerous inaccuracies in the [list of names of G4 data files](https://github.com/next-exp/nexus/wiki/Installing-and-running-nexus#geant4) that need to be downloaded.)

In any case, the proof of concept above demonstrates installation of a specific version of Geant4. I'll look into setting up a binary cache. Also into specialist compilation on the likes of Theta. This last one could be ... *fun*.

Does anyone care about this working on MacOS? (Speak now, or forever hold your peace!)

jacg:
>  We really want to be able to pick the version of Geant4 we're using, as the simulated physics has changed sometimes dramatically even between minor versions. I'd say that is a requirement for the installer.

... in which case I'll need to know which is the specific version that is being used right now by everybody!

jjgomezcadenas:
Yes, we do care about MacOS. Macs are used by many members (perhaps by *all* members) of the simulation team, and by some old timers who eventually use G4 for their toys.

paolafer:
> In any case, the proof of concept above demonstrates installation of a specific version of Geant4.

Yes, this is what we need, I think it's fine. As for the specific version being used *now*, we haven't done any official production for a while. Most developers are using any patch of version 10.6 for their studies, so I think that `10.6.p03` is fine as an example in the repository. Before a specific production we'll fix the version number. 

> As I have already pointed out, the pinning of the nixpgs version via the hash assigned to nixpkgs-commit-id satisfies both of these requirements.

Ok, got it. Just a question: how can one know which version of a specific package is included in nixpkgs-commit-id?

jacg:
> how can one know which version of a specific package is included in nixpkgs-commit-id?

If you have Nix installed, the easiest way is probably

``` shell
nix repl -I nixpkgs=fetchTarball "https://github.com/NixOS/nixpkgs/archive/88f00e7e12d2669583fffd3f33aae01101464386.tar.gz"
Welcome to Nix version 2.3.9. Type :? for help.

Loading 'https://github.com/NixOS/nixpkgs/archive/88f00e7e12d2669583fffd3f33aae01101464386.tar.gz'...
Added 12443 variables.

nix-repl> geant4.version
"10.6.2"

nix-repl> python38Packages.pandas.version
"1.1.1"
```

That is `nix repl -I ...` drops you into an interactive Nix REPL with tab completion, where you can interactively explore all the available derivations (Nix-speak for package definitions). Note that the first line specifies a `nixpkgs-commit-id`.


If you don't have Nix installed, then it's probably easiest to check out the `nixpkgs` repo (or explore it on GitHub). Here's how I might look for the Geant4 version in a local checkout of `nixpkgs`:

``` shell
$ cd src/nixpkgs
$ fd geant4
pkgs/development/libraries/physics/geant4
pkgs/development/libraries/physics/geant4/geant4-hook.sh
$ rg version pkgs/development/libraries/physics/geant4 
pkgs/development/libraries/physics/geant4/datasets.nix
1:{ stdenv, fetchurl, geant_version }:
4:  mkDataset = { name, version, sha256, envvar }:
6:      inherit name version;
7:      inherit geant_version;
10:        url = "https://cern.ch/geant4-data/datasets/${name}.${version}.tar.gz";
18:      datadir = "${placeholder "out"}/share/Geant4-${geant_version}/data/${name}${version}";
38:      version = "4.6";
45:      version = "7.13";
52:      version = "5.7";
59:      version = "5.6";
66:      version = "2.0";
73:      version = "3.1";
80:      version = "3.1";
87:      version = "1.0";
94:      version = "1.3";
101:      version = "2.3";
108:      version = "2.2";
115:      version = "1.3.2";

pkgs/development/libraries/physics/geant4/default.nix
51:  version = "10.7.0";
111:          geant_version = version;
```

That is:

1. `cd` to the repo
2. use `fd` to find where Geant4 lives in the huuuuuuge, and convoluted tree of packages
3. use `rg` to search for the `version`
4. note that the G4 version is 10.7.0 ... and that there also seems to be information about dataset versions, which I could explore if needed

[`rg` and `fd` are two examples of modern CLI tooling which everybody should start using right away to make their lives easier. These two in particular, are a great improvement on `grep` and `find` and combinations of the two.]

jacg:
Before Nix can be adopted as the primary build-system for Nexus, a number of things need to be done, including but probably not limited to

+ Getting Nix to support Geant4 on Darwin.
+ Getting our own Nix binary cache up and running. (Using a private repo probably removes the obvious pre-packaged option: https://cachix.org/)
+ Understanding viability of Nix deployment on various HPC clusters and production machines.
+ Understanding how to take advantage of specialized-compilation for hardware such as Theta.

However, none of this stands in the way of approving this PR. Let me summarize the pros and cons of accepting this PR in its current state:

Pros:

+ The Nexus installation becomes trivial on any Linux machine which is Nix-enabled, for those who care to use it.

+ People unfamiliar with Nix can take it for a spin and see it in action in the context of a project they care about, discover how wonderful it is, identify problems that need addressing, etc.

+ Anyone who doesn't care, won't notice anything because this has absolutely no impact whatsoever on the pre-existing functionality of the repo, except ...

Cons:

+ A checkout of Nexus `master` grows by 2382 bytes! (That's less than twice the size of this comment.)

In other words, I don't see any point in further delaying accepting this PR. Discussions of if and how Nix could be adopted as Nexus' primary build system are, of course, encouraged: these can only benefit from the availability of the work in this PR.

jjgomezcadenas:
Agreed, could any of the usual suspects, please, accept this PR?

paolafer:
As far as I know, @mmkekic was reviewing this PR, trying to use it on her laptop and was having problems. I would wait for the review, before approving it.

jacg:
If there are problems, then it would be useful to hear about them ...

[Note that I have reverted the custom G4 version, to avoid the annoying > 1h compile time. It would be good to know if the problems are with the standard G4 version, or the custom one.]

mmkekic:
I did manage to install with nix (the problems @paolafer  were referring to seem to be related to my direnv installation, but using nix-shell it worked). However, I'd like to check it again with a fresh installation since in my local repository I made a mess building both with scons (using nix) and cmake.
Since @paolafer  made nexus public we have problems of detached forks and I cant fetch any more from @jacg  branch. I am not aware of a way to reattach forks to the main repository even after making them public manually

paolafer:
> Since @paolafer made nexus public we have problems of detached forks and I cant fetch any more from @jacg branch. I am not aware of a way to reattach forks to the main repository even after making them public manually

I think the fastest way is to make a new fork from the scratch and push there our branches.

jacg:
> I cant fetch any more from @jacg branch. 

@mmkekic, please try now.

> I am not aware of a way to reattach forks to the main repository even after making them public manually

I think I've managed to reattach my fork, but it looks like the link between the PR and my branch has been lost. (I just pushed a dummy commit which has appeared in my fork, but I can't see it in the PR).

jacg:
> I think the fastest way is to make a new fork from the scratch and push there our branches.

I think that this *mostly* works:

1. Fetch from your fork into your local repo (just as a precaution).
2. Delete your fork.
3. Create a new fork.
4. Your local repo now points at the new fork (as long as you deleted the old fork before making the new one).
5. Everything should now be working again ...
6. ... except any open PRs associated with branches in the old fork. At the moment I can't see any solution beyond closing them and re-submitting new ones from the new forks.

jacg:
https://stackoverflow.com/questions/36225381/reattach-a-detached-fork-in-github suggests that GitHub support can reattach them for you ... and that the process I describe in my previous comment does indeed break currently open PRs.

mmkekic:
Ok, I confirm that this does install nexus as expected, so I guess this PR can be accepted. Can we accept it even if the fork is detached @jacg ?

paolafer (approved):

jacg:
> Can we accept it even if the fork is detached @jacg ?

Looks like @paolafer just pragmatically demonstrated that it can :-)

jacg:
One of the major things missing from Nix-support is Geant4 on MacOS. Not having access to a Mac myself (except the GitHub Actions build server) makes it very difficult to make any progress on this. Some assistance/participation from someone who has a (preferably fairly powerful: G4 compile times are significant) Mac, would be very useful.

Also, has anyone got, or is anyone planning on getting an M1 Mac any time soon? This might also be an 'interesting' issue to deal with!

PR #65, opened 2021-01-21
Demopp run8

jmunozv:
This PR implements the "Run-8" configuration in DEMOpp geometry.

(Closed to be re-opened as #73 due to the repo change from private to public)

### DIFF
--- a/source/geometries/NextDemoTrackingPlane.cc
+++ b/source/geometries/NextDemoTrackingPlane.cc
@@ -3,8 +3,8 @@
 //
 // Tracking plane of the Demo++ geometry.
 // It implements 2 different TP implementations:
-// "original": the one used from the very beginning with passing-holes masks
-// "type 1": the one used in the so-called Run7, with coated masks and membranes.
+// "run5": the one used from the very beginning with passing-holes masks
+// "run7" and "run8": the ones used with coated masks and membranes.
 //
 // The NEXT Collaboration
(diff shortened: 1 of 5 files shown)


source/geometries/NextDemoTrackingPlane.cc
```diff
@@ -3,8 +3,8 @@
 //
 // Tracking plane of the Demo++ geometry.
 // It implements 2 different TP implementations:
-// "original": the one used from the very beginning with passing-holes masks
-// "type 1": the one used in the so-called Run7, with coated masks and membranes.
+// "run5": the one used from the very beginning with passing-holes masks
```
paolafer, line 6: I would remove "from the very beginning" or replace it with "at the beginning", since it applies only to that run.

paolafer (changes requested):

paolafer:
This PR adds a configuration to the DEMO geometry, parametrizing the characteristics of the teflon masks, following the current tests in the detector. The code works fine, so, except for a minor comment, I approve this PR.

source/geometries/NextDemoTrackingPlane.cc
jmunozv: Done

paolafer: I don't see the change...

PR #68, opened 2021-01-22
Add GitHub action: install with Nix + run all tests

jacg:
Now that the repo is public, the 2000 minute per month GitHub Actions limit no longer apples, so I can't see any obvious reason why we shouldn't run automatic tests on GHA.

The proposed action 

1. uses the recently-approved Nix installation method to install the Nexus dependencies
2. builds Nexus with SCons
3. runs `bin/nexus-test`
4. runs `pytest`

### DIFF
--- a/.github/workflows/install-via-nix-and-test.yml
+++ b/.github/workflows/install-via-nix-and-test.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-10.15]
+        os: [ubuntu-latest]
         py: [38]
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


jacg:
Would be good to convince SCons to compile with all available cores, but it's not obvious to me how to do that portably across Linux and MacOS. This doesn't matter so much right now, as MacOS is disabled in the current version, but it will need to be done eventually.

jmbenlloch (commented):
This PR adds a way to automatically run nexus tests taking advantage of the Nix building process setup in #64. In @jacg's fork we can see [everything works well](https://github.com/jacg/nexus/runs/1748211474?check_suite_focus=true).

The dependencies used are the ones available with the Nix setup currently included. If the developers are happy running automatically the tests with those (@paolafer ?). I would recommend the approval of this PR. Basically this PR "comes for free" once we have Nix included in the repository.

PS: I see @jacg has included a convenient check: if the commit title includes: `[skip ci]` then the tests won't run. It might be useful in some occasions.

paolafer:
> The dependencies used are the ones available with the Nix setup currently included. If the developers are happy running automatically the tests with those (@paolafer ?). I would recommend the approval of this PR. Basically this PR "comes for free" once we have Nix included in the repository.
 
The current configuration is fine, so we can go ahead.

jmbenlloch (approved):
> Would be good to convince SCons to compile with all available cores, but it's not obvious to me how to do that portably across Linux and MacOS. This doesn't matter so much right now, as MacOS is disabled in the current version, but it will need to be done eventually.

I don't think this is critical now, maybe can be done later at another PR.

> The current configuration is fine, so we can go ahead.

Since @paolafer is happy with the configuration used. I approve the PR.

PR #71, opened 2021-01-23
Add binary cache for Nix

jacg:
# Drastic improvement in build time

The headings in the following table are active links to two different GitHub action runs (PR #68), which build Nexus with Nix and run both `bin/nexus-test` and `pytest`.

|  |[without cache](https://github.com/jacg/nexus/runs/1750041091?check_suite_focus=true) | [with cache](https://github.com/jacg/nexus/runs/1754072031?check_suite_focus=true)|
--|-----------|----| 
Total time |115 minutes | 20 minutes |
Install dependencies via Nix |93 minutes |  2 minutes |

The point to note is that the cache reduces the run time by around 90 minutes.

# How?

This has been achieved by caching the result of compiling Geant4 10.6.3, which is not the version of Geant4 that comes with the `nixpgs` version that we are currently using, and is therefore not available in the default Nix binary cache.

The cache added in this PR is hosted by [Cachix](https://cachix.org), using their free *Community plan*. At present the cache is owned by me personally: Cachix is planning to add support for GitHub organizations on a time-scale of around a month, at which point the cache should be handed over to `next-exp`.

This PR ensures that the results of Nix compilations in our GitHub action are added to the cache, and that subsequent GHA runs use those results, thus avoiding the need to repeat the compilation.

# Cache is available to everyone

Everyone else can benefit from this cache by following these two simple steps:

1. installing `cachix` (a number of trivial ways, if Nix is installed)
2. `cachix use jacg-next-exp`

In step 2, `jacg-next-exp` is the name of the cache. We will probably choose to rename this to something like `next-exp` once Cachix supports organizations.

# Geant4 10.6.2 -> 10.6.3

In PR #64 we accepted a Nix expression which built Geant4 10.6.2: this is the version that comes with the `nixpkgs` version we are using, and was thus available in the standard Nix binary cache.
Now that we are caching our build-products, Geant4 10.6.3 is no longer prohibitively expensive to use, hence this PR bumps the Geant4 version from 10.6.2 to 10.6.3.

### DIFF
--- a/.github/workflows/install-via-nix-and-test.yml
+++ b/.github/workflows/install-via-nix-and-test.yml
@@ -17,9 +17,10 @@ jobs:
       - uses: cachix/install-nix-action@v12
         with:
           nix_path: nixpkgs=channel:nixos-20.09
-      - name: Enable Cachix binary cache
-        run: |
-          nix-env -iA cachix -f https://cachix.org/api/v1/install # Developers should prefer installation via home-manager
-          cachix use jacg-next-exp
+      - uses: cachix/cachix-action@v8
+        with:
+          name: jacg-next-exp
+          # If you chose API tokens for write access OR if you have a private cache
+          authToken: '${{ secrets.CACHIX_JACG_NEXT_EXP }}'
       - name: Install dependencies via Nix
         run: nix-shell --argstr py ${{ matrix.py }}
--- a/.github/workflows/install-via-nix-and-test.yml
+++ b/.github/workflows/install-via-nix-and-test.yml
@@ -24,9 +24,3 @@ jobs:
           authToken: '${{ secrets.CACHIX_JACG_NEXT_EXP }}'
       - name: Install dependencies via Nix
         run: nix-shell --argstr py ${{ matrix.py }}
-      - name: Build Nexus with SCons
-        run: nix-shell --argstr py ${{ matrix.py }} --run "scons"
-      - name: Run bin/nexus-test
-        run: nix-shell --argstr py ${{ matrix.py }} --run "bin/nexus-test"
-      - name: Run pytests
-        run: nix-shell --argstr py ${{ matrix.py }} --run "pytest -x"
--- a/.github/workflows/install-via-nix-and-test.yml
+++ b/.github/workflows/install-via-nix-and-test.yml
@@ -17,10 +17,9 @@ jobs:
       - uses: cachix/install-nix-action@v12
         with:
           nix_path: nixpkgs=channel:nixos-20.09
-      - uses: cachix/cachix-action@v8
-        with:
-          name: jacg-next-exp
-          # If you chose API tokens for write access OR if you have a private cache
-          authToken: '${{ secrets.CACHIX_JACG_NEXT_EXP }}'
+      - name: Enable Cachix binary cache
+        run: |
+          nix-env -iA cachix -f https://cachix.org/api/v1/install # Developers should prefer installation via home-manager
+          cachix use jacg-next-exp
       - name: Install dependencies via Nix
         run: nix-shell --argstr py ${{ matrix.py }}
--- a/.github/workflows/install-via-nix-and-test.yml
+++ b/.github/workflows/install-via-nix-and-test.yml
@@ -17,6 +17,11 @@ jobs:
       - uses: cachix/install-nix-action@v12
         with:
           nix_path: nixpkgs=channel:nixos-20.09
+      - uses: cachix/cachix-action@v8
+        with:
+          name: jacg-next-exp
+          # If you chose API tokens for write access OR if you have a private cache
+          authToken: '${{ secrets.CACHIX_JACG_NEXT_EXP }}'
       - name: Install dependencies via Nix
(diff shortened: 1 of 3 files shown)


jmbenlloch:
This looks very good, but there are some details I'm missing.

> This PR ensures that the results of Nix compilations in our GitHub action are added to the cache, and that subsequent GHA runs use those results, thus avoiding the need to repeat the compilation.

What exactly will be cached? In your workflok I see the cachix-action is among the firsts steps to be executed. I understand that during the dependencies installation, Nix will search for the packages (like Geant4) in the binary cache from cachix. But what happens later? I see you have define a Github secret with the API key. Will nexus itself be uploaded to the cache after compilation? Would that be only for the dependencies?

jacg:
> This looks very good, but there are some details I'm missing.

Please bear in mind that I'm not 100% sure on every single detail myself :-)

>> This PR ensures that the results of Nix compilations in our GitHub action are added to the cache, and that subsequent GHA runs use those results, thus avoiding the need to repeat the compilation.

> What exactly will be cached? 

Any Nix derivations that are built (rather than downloaded from a cache) in the workflow, after the initial cachix action. You can see exactly what was cached by looking in the post-run-cachix-action. Hmm, for boring reasons, it's not visible in the Action runs linked at the top of this PR, so you have to look here instead:

 https://github.com/jacg/nexus/runs/1750751074?check_suite_focus=true#step:9:2
 
I don't know how to create a link that will fully expand it, so you have to click the arrow on the highlighted line, to see the items being written. You will see **two** versions of geant4-10.6.3: I don't understand why there are two!

> In your workflok I see the cachix-action is among the firsts steps to be executed. I understand that during the dependencies installation, Nix will search for the packages (like Geant4) in the binary cache from cachix. But what happens later?

The Post run cachix action determines which derivations were built *here*, and uploads them to the cache.

> I see you have define a Github secret with the API key. 

Yup. This is a public cache (Cachix lets you have private ones, but not on the free plan, I think). Anyone can read it, but you need permission to write to it. (Therefore, as it stands, anything that you build on your own machine, will not be cached. By installing the key on your machine, you will be able to populate the cache.)

> Will nexus itself be uploaded to the cache after compilation? 

Nope, because there is no Nix derivation that builds Nexus. We build Nexus with SCons, outside of any derivation, simply as a GitHub Action `run` instruction. We currently only use Nix to provide the dependencies needed for Nexus to be built.

That is, we use Nix to help develop Nexus, but not to package or distribute Nexus.

We could (and probably should, eventually) write a Nix derivation that builds Nexus itself, then we could cache and or distribute Nexus via Nix.

If we cached a Nexus build, this would save us however much time it takes for SCons to build Nexus, on subsequent builds, **as long as absolutely nothing has changed in Nexus**: change one single bit in the repo, and the derivation needs to be rebuilt.

> Would that be only for the dependencies?

At present, yes.

jmbenlloch:
> Any Nix derivations that are built (rather than downloaded from a cache) in the workflow, after the initial cachix action. You can see exactly what was cached by looking in the post-run-cachix-action.

Understood. I had a look too at the code of the cachix/cachix-action and it is pretty clear the way it works. It checks the nix store folder before and after building the project. Any new folder in /nix/store would be uploaded.

>  You will see two versions of geant4-10.6.3: I don't understand why there are two!

This is very odd. I did look at the two folders after downloading them to my machine and the look pretty similar. Checking the hash of all the files I've seem some dynamic libraries are different, but I don't understand either why two versions where created... both folders has the same structure and files.

> This is a public cache (Cachix lets you have private ones, but not on the free plan, I think). Anyone can read it, but you need permission to write to it. (Therefore, as it stands, anything that you build on your own machine, will not be cached. By installing the key on your machine, you will be able to populate the cache.)

I have checked locally that enabling cachix, nexus can be compiled from scratch in a few minutes. Everything seems to be working well outside of github too.

> there is no Nix derivation that builds Nexus. We build Nexus with SCons, outside of any derivation, simply as a GitHub Action run instruction. We currently only use Nix to provide the dependencies needed for Nexus to be built.

Understood. I think, for this PR, providing a nexus derivation would be out of scope.

jmbenlloch (approved):
This PR improves dramatically the nix-based building procedure and the associated github action. It sets up a binary cache the is very useful for current and future dependencies. I'd recommend merging this PR (@paolafer).

jacg:
>> You will see two versions of geant4-10.6.3: I don't understand why there are two!

>This is very odd. I did look at the two folders after downloading them to my machine and the look pretty similar. Checking the hash of all the files I've seem some dynamic libraries are different, but I don't understand either why two versions where created... both folders has the same structure and files.

Yes, it's odd. It may be because of some mistake in the our `shell.nix`. In any case it is orthogonal to this PR.

>> there is no Nix derivation that builds Nexus. We build Nexus with SCons, outside of any derivation, simply as a GitHub Action run instruction. We currently only use Nix to provide the dependencies needed for Nexus to be built.

> Understood. I think, for this PR, providing a nexus derivation would be out of scope.

Agreed. Whether or not we *want* to build Nexus itself with Nix, let alone how we would do it exactly, is beyond the scope of this PR.

PR #72, opened 2021-01-25
Removed warnings printed when testing Interpolation and XenonGasPrope…

tcontrer:
…rties

In order to remove the warnings printed while testing, I changed the G4Exceptions to c++ exceptions. This way the exceptions can be caught using catch2, and not printed to the terminal.

### DIFF
--- a/source/materials/XenonGasProperties.cc
+++ b/source/materials/XenonGasProperties.cc
@@ -73,8 +73,7 @@ G4double XenonGasProperties::Density(G4double pressure)
       density = data[n_pressures-1][1];
     }
     else {
-      G4Exception("[XenonGaseousProperties]", "Density()", FatalException,
-      "Unknown xenon density for this pressure!");
+      throw "Unknown xenon density for this pressure!";
     }
   }
 
@@ -234,11 +233,11 @@ void XenonGasProperties::MakeDataTable()
   // Open file
   G4String path(std::getenv("NEXUSDIR"));
   G4String filename = path + "/data/gxe_density_table.txt";
+
   std::ifstream inFile;
   inFile.open(filename);
   if (!inFile) {
-    G4Exception("[XenonGaseousProperties]", "Density()", FatalErrorInArgument,
-    "File could not be opened!");
+    throw "File could not be opened";
   }
 
   std::vector<std::vector<G4double>> data;
@@ -251,6 +250,7 @@ void XenonGasProperties::MakeDataTable()
   G4double thistemp = 0;
   G4double temp, press, dens;
   char comma;
+
   while (inFile>>temp>>comma>>press>>comma>>dens){
     std::vector<G4double> thisdata {temp*kelvin, press*bar, dens*(kg/m3)};
     data_.push_back(thisdata);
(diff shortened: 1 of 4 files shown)


paolafer:
Catch2 requires to deal with a C++ exception for tests that want to ensure that an exception is raised. To our knowledge, the `G4Exception` class doesn't work. This PR shows a workaround, replacing G4 exceptions with pure C++ ones. What do you think @jmalbos, @jmunozv? I admit it's not ideal to have different ways to deal with exceptions across the code.  If we decide to go with this solution, at least I would modify the exception message to match the style of G4 exceptions.

Another option would be to keep the `G4Exception` class and use the code previously in `XenonGasPropertiesTests.cc`, which I personally found a bit obscure.

Either way, the exceptions to be tested cannot be warnings, but need to bring to the termination of the program, which I think it's appropriate in this case (xenon gas properties and interpolation). 

This issue made me wonder which software the G4 developers use for testing the code, since it may be worth it to explore if we can use the same one.

jjgomezcadenas:
Again, my five cents here:

The reason why G4 decides to overwrite such things as types (integers, floats etc) or exceptions is a mystery to me. But if a pure C++ exception does the job better than a G4Exception I would not hesitate to replace it. Actually, I would probably not hesitate to replace all G4Exceptions by C++ exceptions (or at least use C++ exceptions in the future). 

As for the software used to test the code, this is an interesting question, perhaps @jacg who has taught C++ for 20 years can suggest an approach. I would not limit this exploration to find out what G4 has done. There may be more optimal ways to test the code.

jacg:
> The reason why G4 decides to overwrite such things as types (integers, floats etc) or exceptions is a mystery to me. 

Hehe ... you know my hypothesis :-)

> But if a pure C++ exception does the job better than a G4Exception I would not hesitate to replace it. Actually, I would probably not hesitate to replace all G4Exceptions by C++ exceptions (or at least use C++ exceptions in the future).

I agree completely ... in principle. However, without fully understanding *why* G4 chooses this apparent insanity, and to what extent the framework's correct functioning relies on actual use of these things, we won't know whether our use of the standard language features merely appears to work, or whether it will blow up in our face later on.



> As for the software used to test the code, this is an interesting question, perhaps @jacg who has taught C++ for 20 years can suggest an approach. 

I have mostly been teaching the language itself, not the tooling that surrounds it. And I have been trying to distance myself from C++ and its tooling in recent yars. Consequently, I am really not up to date on the state of the art of testing in C++.

> I would not limit this exploration to find out what G4 has done. There may be more optimal ways to test the code.

This is an important point. However, precisely because of the aforementioned ... how shall I put it politely ... *idiosyncratic* nature of G4, we have to understand how they test themselves, to see if there are any crucial idiosyncrasies that must be adhered to.

jmunozv:
Regardless of the type of exceptions thrown and how they are handled, and although not sure at 100%, I agree with @paolafer Paola that a Warning is not enough for these exceptions, and the simulations should be aborted.

paolafer (approved):
While we investigate the best way of doing tests in nexus and whether using G4exceptions or C++ ones, I would approve this PR.
This way we get rid of the warnings in the tests, which add noise to the interpretation of test output, and we change these particular warnings to exceptions, which is the correct behaviour.

PR #73, opened 2021-01-26
Demopp run8

jmunozv:
This PR adds the Run8 configuration configuration to the DEMO geometry.

It also adds some parameters to NextDemoSiPMBoards.

(Coming from #65, due to the repo change from private to public)


paolafer (approved):
Approved as detailed in #65.

PR #74, opened 2021-01-26
Update xenon EL-yield to manage pressure dependant values

jmunozv:
This PR modifies the xenon EL-yield formula to manage the different pressure dependant values.

As the way in which the yield varies is unknown, it has been implemented in a simple way in which given a pressure it is assigned the constants of the immediately lower known pressure.

The constants for the different pressures come from:
Freitas-2010 // Physics Letters B 684 (2010) 205–210

(Coming from #61, due to the repo change from private to public)


paolafer (approved):
This PR is the same as #61, closed because of an issue with the change of the repository visibility from private to public.
We considered that the code is clearer without the line of 3 bar pressure and that the reference to the paper is enough for people interested in knowing where these values come from. I approve it.

PR #75, opened 2021-01-27
Upgrade Geant4 10.6.3 -> 11.0.0

### LINKED ISSUE #57: Build fails with Geant4 10.7
Geant4 10.7: https://geant4.web.cern.ch/node/1914

jacg:
Fix #57.

One of the great things about using Nix, is that only one person in the community needs to fix a problem, for everyone to benefit. That is, we don't need to re-invent the wheel of getting Geant 10.7 to build, if we can use the solution provided by someone else (presumably someone who has much more expertise in this area than we do!) in `nixpkgs`.

This PR fixes the Geant 10.7 build, by bumping `nixpkgs-commit-id` to one which contains Geant 10.7.

This switches from the the `nixos-20.09` channel to `nixpkgs-unstable`, which will change the versions of all sorts of other packages. If this has some adverse effects, we might look into using our geant4 overlay to change the Geant4 version only.

Note that 

+ Geant 10.7 builds successfully
+ `bin/nexust-test` passes
+ some of the `pytest`s now fail:

    https://github.com/next-exp/nexus/runs/1776873953?check_suite_focus=true
   
  Someone who is familiar with these tests should look into it.
  
This PR removes Nix' ability to build Geant4 10.6.3 If we are interested in being able to compare runs across Geant4 10.6.3 and 10.7, we should add an overlay which adds the non-default version as an 'extra package'.

I shall leave this as a draft PR until the `pytest` problems are resolved.


paolafer:
I'm confused, does nexus compile with Geant4 10.7? I thought we had some incompatibility. How does this PR solve that problem?

jacg:
> I'm confused, does nexus compile with Geant4 10.7?

Nexus compiles with Geant4 10.7, as can be seen [here](https://github.com/jacg/nexus/runs/1776873769?check_suite_focus=true#step:6:7)

> I thought we had some incompatibility. 

As #57 offers no details whatsoever about the nature of this problem, I have no idea about this.

> How does this PR solve that problem?

This PR solves whatever problem there may or may not have been with getting Nexus to build with Geant4 10.7, by using a version of nixpkgs which provides Geant4 10.7. With that version, our `shell.nix` successfully builds Nexus. Caveat: the basic tests (`bin/nexust-test`) pass, but the more thorough tests (`pytest`) now fail. Nexus *builds* with Geant4 10.7, but fails *some tests*. If that (failing tests, rather than failing to build) is what was meant by "Build fails with Geant4 10.7" in #57 ... well, I'm not a mindreader, I can only go by what is written.

paolafer:
Ok, I'll have a look into the tests, as soon as I can.

jacg:
> How does this PR solve that problem?

If the problem really was a build problem ... perhaps this sort of fiddling around


``` shell
    # Fix for boost 1.67+
    substituteInPlace environments/g4py/CMakeLists.txt \
      --replace "REQUIRED python" \
                "REQUIRED COMPONENTS $boost_python_lib"
    substituteInPlace environments/g4py/G4PythonHelpers.cmake \
      --replace "Boost::python" "Boost::$boost_python_lib"
```

which is done in the Geant4 10.7 derivation in `nixpkgs` is something that was necessary but we didn't know how to do.

jacg:
> does nexus compile with Geant4 10.7? I thought we had some incompatibility. 

Note that our `shell.nix` uses `SCons` to build Nexus. Maybe the problem only appeared with `CMake`?

paolafer:
Maybe @MiryamMV can help here, since she was the one who had problems. I haven't had the chance to try G4 10.7 so far.

jacg:
> I haven't had the chance to try G4 10.7 so far.

With Nix, it's trivial!

``` shell
cd wherever/nexus
git fetch jacg geant-10.7
shell-nix # not necessary if you have direnv configured
```
... wait a few minutes, and you have Nexus compiled with Geant4 10.7!


Couldn't really be easier, could it?


(Unfortunately, only on Linux so far. I'll keep on hammering on about this point of how easy things become with Nix, and how it's not working on Mac yet, in the hope that someone who actually has a Mac (unlike me) offers some assistance in getting it to work. Meanwhile, maybe you can try it on some production machine: I know that @mmkekic has proven that Nix works on at least one. Alternatively run Linux in a VM.)

paolafer:
I've compiled and run nexus against Geant.10.7.p01 with success. 
However, as @jacg pointed out, some tests are failing. I have a branch with the fixes to that, should I open a new PR or add it to this one?

jacg:
Any reason not to add the fixes here? This branch is waiting for those fixes, after all.

Maybe rebase it first.

paolafer:
Occasionally a test still fails (not always). I'm looking into that.

paolafer:
I have produced some validation plots for K40, Co60, Bi214 and Tl208 isotopes + Xe137, comparing the GEANT4.11.0.0 version with a previous one, GEANT4.10.5.p01 (which was the last one that had been validated), which can be found [here](https://next.ific.uv.es/DocDB/0013/001305/002/Geant4_11_validation.pdf).

source/physics/Electroluminescence.h
```diff
@@ -10,7 +10,7 @@
 #define ELECTROLUMINESCENCE_H
 
 #include <G4VDiscreteProcess.hh>
-
+#include <G4PhysicsOrderedFreeVector.hh>
```
paolafer, line 13: Since you had to add the `include` here, you can delete it from the source file. The same for the other classes.

gondiaz:
@jacg any clue on why pytest does not find `numpy` and `pandas` modules in the github action?

Comment: We need to use nixos-unstable instead of nixos 21.11 because we updated nexus to G4 11, being G4 10.7 not longer compatible.

jacg:
> any clue on why pytest does not find numpy and pandas modules in the github action?

Yes, I probably know the reason. The mystery is why it worked at all, the way it was written! I recall being surprised that it worked, and deciding to stick with this simpler version, even though it wasn't *supposed* to work. Looks like whatever bug the simplification relied on, has been fixed in current Nix.

The fundamental point is that Nix forgoes finding anything in well-known locations in the filesystem, in order to give you precise control over the exact versions of absolutely everything in the dependency graph. In the case of Python, such well-known locations would be `site-packages` and the like. 

If you want a python with numpy, you should declare a python-with-numpy. As it stands, `shell.nix` declares that you want python, and, separately, declares that you want numpy, and *somehow* this python finds this numpy ... or maybe some other numpy. This is *not* supposed to happen.

The way to declare a python-with-stuff is to use `pythonWithPackages`. [Here](https://github.com/jacg/nix-starters/blob/c45d38d2b0e927e8ad342143a85f6919d9dbab97/python/shell.nix#L28-L31) is a minimal example of how it works. The example passes CI using nixos 21.11.

jacg:
I'm getting a bit lost among all the force-pushes every 5 minutes, so I can't find the last comment I wrote, but you seem to have figured it out for yourself while my network connection was down. Well done.

gondiaz:
It is already there, it was duplicated before

source/materials/OpticalMaterialProperties.cc
```diff
@@ -528,17 +528,18 @@ namespace opticalprops {
...
 
     // CONST PROPERTIES
     mpt->AddConstProperty("SCINTILLATIONYIELD", sc_yield);
-    mpt->AddConstProperty("FASTTIMECONSTANT",   6.*ns);
-    mpt->AddConstProperty("SLOWTIMECONSTANT",   37.*ns);
-    mpt->AddConstProperty("YIELDRATIO",         .52);
+    mpt->AddConstProperty("SCINTILLATIONTIMECONSTANT1",   6.*ns);
+    mpt->AddConstProperty("SCINTILLATIONTIMECONSTANT2",   37.*ns);
+    mpt->AddConstProperty("SCINTILLATIONYIELD1", .342);
+    mpt->AddConstProperty("SCINTILLATIONYIELD2", .657);
```
paolafer: The sum of `SCINTILLATIONYIELD1` and `SCINTILLATIONYIELD2` should give 1, right? Therefore, I would change the second to `.658` (which is the rounding I find from the calculation).

source/materials/OpticalMaterialProperties.cc
```diff
@@ -1503,24 +1421,24 @@ namespace opticalprops {
 
     // REFRACTIVE INDEX
     const G4int rIndex_numEntries = 2;
```
gondiaz: Remove

source/materials/OpticalMaterialProperties.cc
```diff
@@ -683,7 +658,7 @@ namespace opticalprops {
                                       G4double e_lifetime,
                                       G4double photoe_p)
   {
-    G4MaterialPropertiesTable* mpt      = new G4MaterialPropertiesTable();
+    G4MaterialPropertiesTable* mpt = new G4MaterialPropertiesTable();
 
     // PROPERTIES FROM XENON
     G4MaterialPropertiesTable* xenon_pt = opticalprops::GXe(pressure, temperature, sc_yield, e_lifetime);
```
gondiaz, line 664: The variable `temperature` in `GXe` is defined but not used, do we want to keep it?

paolafer: I know, it is there, because we have written (but not used) a function that calculates the density, which depends also on the temperature. However, we are not using it, so I think we can remove the current dependence on the temperature and add it later, if needed.

source/materials/OpticalMaterialProperties.cc
paolafer: Done.

source/materials/OpticalMaterialProperties.cc
paolafer: On a second thought, since the idea is to use the other function, it's probably better to stick to the current code and open a new PR to change the density function to the new version.

paolafer (approved):
This PR adapts the nexus code to be used with GEANT4 version >= 11.0. It also changes the way we use sensitive detectors to store the sensor response, in a more logical way. Now it is GEANT4 that checks if the photon has reached an optical boundary and the status is `Detection`, and only in that case the sensitive detector is triggered. Great job, approved!

PR #76, opened 2021-01-29
Pyrex panel optical properties

carmenromo:
In the `PetBox` geometry there are three possible tiles the user can choose with a configuration parameter. One of them is the `TileHamamatsuBlue`, which has an additional pyrex panel defined in front of the sensors close to the the wavelength shifter, both defined in the `PetBox` class, not in the `TileHamamatsuBlue` class.

New optical properties for this pyrex have been defined according with the datasheet of the company ([Vidrasa](http://www.vidrasa.com/esp/productos/borofloat/borofloat_pf.html#:~:text=Borofloat%20(vidrio%20plano%20borosilicato%203.3)%3A%20Propiedades%20f%C3%ADsicas%20y%20qu%C3%ADmicas&text=Las%20planchas%20de%20BOROFLOAT%C2%AE,tiempo%20es%20de%20500%20%C2%BAC.)) in the class `OpticalMaterialProperties` and applied only for these two panels, which are the only ones where the optical photons pass through.

### DIFF
--- a/source/geometries/PetBox.cc
+++ b/source/geometries/PetBox.cc
@@ -299,7 +299,7 @@ namespace nexus {
       new G4Box("ENTRY_PANEL", entry_panel_x_size_/2., entry_panel_y_size_/2., panel_thickness_/2.);
 
     G4Material* pyrex = G4NistManager::Instance()->FindOrBuildMaterial("G4_Pyrex_Glass");
-    pyrex->SetMaterialPropertiesTable(OpticalMaterialProperties::Pyrex_vidrasa(panel_thickness_));
+    pyrex->SetMaterialPropertiesTable(OpticalMaterialProperties::Pyrex_vidrasa());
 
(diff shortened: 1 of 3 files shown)


carmenromo:
image

paolafer:
Can you add in the code  the reference where you took the transmittance from?

carmenromo:
> Can you add in the code the reference where you took the transmittance from?

Sure!

paolafer:
I think http://www.vidrasa.com/esp/productos/borofloat/borofloat_pf.html works as well and it's shorter.

paolafer (approved):
This PR adds transmission properties for photons in the visible range to pyrex. It will be used in the simulations with conventional SiPMs and a panel coated with wavelength shifter in front of them.
The code compiles and runs and it has been checked that the photons reach the sensors. I approve it.

PR #77, opened 2021-02-01
Implement the assymetric geometry and comment the symmetric one

carmenromo:
The initial measurements with the PetBox geometry will be performed with an asymmetric configuration of the tiles. Four tiles in one side and one tile in the opposite side. This PR implements this change and the symmetric geometry has been commented in order to be used in the future.


paolafer (approved):
This PR replaces the four tiles on one side of the detector with just one, matching to the current condition of operation.

PR #78, opened 2021-02-02
fix NEW-ICS barrel holes position

halmamol:
Adding correction in NEW-ICS barrel holes. After 5cm prolongation of the ICS barrel into the energy plane, there was shift in the position of the upper and lateral nozzles with respect to the vessel ones.


jmunozv (approved):
I approve this PR that fixes the misalignment between NEW vessel ports and the corresponding ICS holes introduced by the extension of the ICS bars up to the Energy Plane.

PR #79, opened 2021-02-02
Change back PMT IDs to original values.

paolafer:
The DEMO++ IC database has been updated, so there is no mismtach in the position of real sensors and simulation, so there's no need  for different IDs. This PR changes back the PMT IDs to [0, 1, 2].


jmunozv (approved):
I approve this PRs that changes back NextDemo PMT ids to their original values.

PR #80, opened 2021-02-04
Fix bug in the generation of event vertices in EL gap

jmalbos:
In the geometry of NEW, the event vertices in the EL gap were not being repositioned in the global frame of reference at the level of the main class (`NextNew`). This pull request fixes that bug.

### DIFF
--- a/macros/NEW_elgap.config.mac
+++ b/macros/NEW_elgap.config.mac
@@ -1,7 +1,7 @@
 /control/verbose  0
 /run/verbose      0
 /event/verbose    0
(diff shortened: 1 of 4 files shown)


jmalbos:
I've pushed two more commits. They fix a sign change in the x coordinates introduced by the rotation of the world reference frame and they add macros for the generation of vertices in the EL gap of NEW.

jmunozv (approved):
I approve this PR that fixes the global coordinates of the EL_GAP vertex generator of NEW geometry.

PR #81, opened 2021-02-04
Fix ITO composition

jmunozv:
Simple PR that fixes the current ITO material composition by adding the missing Sn element, fixing the number of Oxygens and updating its density.

### DIFF
--- a/source/materials/MaterialsList.cc
+++ b/source/materials/MaterialsList.cc
@@ -478,6 +478,9 @@ G4Material* MaterialsList::TPB()
 
 G4Material* MaterialsList::ITO()
 {
+  // Density data from  https://doi.org/10.1116/1.1371326
+  // Composition data from: https://en.wikipedia.org/wiki/Indium_tin_oxide
+
   G4String name = "ITO";
 
   G4Material* mat = G4Material::GetMaterial(name, false);
@@ -486,11 +489,13 @@ G4Material* MaterialsList::ITO()
     G4NistManager* nist = G4NistManager::Instance();
 
     G4Element* In = nist->FindOrBuildElement("In");
-    G4Element* O = nist->FindOrBuildElement("O");
+    G4Element* O  = nist->FindOrBuildElement("O");
+    G4Element* Sn = nist->FindOrBuildElement("Sn");
 
-    mat = new G4Material(name, 7.14*g/cm3, 2, kStateSolid);
-    mat->AddElement(In, 2);
-    mat->AddElement(O, 3);
+    mat = new G4Material(name, 6.8*g/cm3, 3, kStateSolid);
+    mat->AddElement(In, 74. * perCent);
+    mat->AddElement(O , 18. * perCent);
+    mat->AddElement(Sn,  8. * perCent);
   }
   return mat;
 }


source/materials/MaterialsList.cc
```diff
@@ -486,11 +488,13 @@ G4Material* MaterialsList::ITO()
     G4NistManager* nist = G4NistManager::Instance();
 
     G4Element* In = nist->FindOrBuildElement("In");
-    G4Element* O = nist->FindOrBuildElement("O");
+    G4Element* O  = nist->FindOrBuildElement("O");
+    G4Element* Sn = nist->FindOrBuildElement("Sn");
 
-    mat = new G4Material(name, 7.14*g/cm3, 2, kStateSolid);
+    mat = new G4Material(name, 6.8*g/cm3, 3, kStateSolid);
     mat->AddElement(In, 2);
```
jmalbos: Where are these number coming from? I can't find them in the link you provide. In any case, everywhere else the composition is given by weight.

source/materials/MaterialsList.cc
```diff
@@ -478,6 +478,8 @@ G4Material* MaterialsList::TPB()
 
 G4Material* MaterialsList::ITO()
 {
+  // Data from http://www.mit.edu/~6.777/matprops/ito.htm
```
jmalbos: This appears to be a random reference, and the link may eventually stop working. Instead, you could include the DOI of the papers they cite.

jmalbos (changes requested):

source/materials/MaterialsList.cc
paolafer: In the Wikipedia I found 


_Indium tin oxide is typically encountered as an oxygen-saturated composition with a formulation of 74% In, 18% O2, and 8% Sn by weight._


 (https://en.wikipedia.org/wiki/Indium_tin_oxide)

paolafer: Here is another, specific, ITO by Kurt J. Lesker: https://www.lesker.com/newweb/deposition_materials/depositionmaterials_sputtertargets_1.cfm?pgid=in3.
I'm not sure about the way we should implement it.

jmunozv: The density comes from the commented link.
Respect the composition of the materials, in nexus some of them are given by weight, and many others by the number of atoms (TPB, HDPE, OpticalSilicone, Kevlar, PVT ...).
But if you prefer, I can change it to compositions by weight.

source/materials/MaterialsList.cc
jmunozv: It's weird. The link works for me without any problem.
In it you can find several paper cites. The one associated to the density is https://doi.org/10.1116/1.1371326.
I can change the current one  by this one if you prefer.

jmalbos: Sorry, I didn't mean the link doesn't work. I meant that it is someone's personal website, and thus the link may stop working any time. I'd rather include the link to the paper, yes.

jacg: Just because the link works now, doesn't mean it won't be gone by next week!

source/materials/MaterialsList.cc
jmalbos: > Respect the composition of the materials, in nexus some of them are given by weight, and many others by the number of atoms (TPB, HDPE, OpticalSilicone, Kevlar, PVT ...). But if you prefer, I can change it to compositions by weight.

But where did you get the numbers for the composition by number of atoms? I can't find them in that website. I think I'd use the composition by weight in this particular case, as it appears to have varying proportions.

jmunozv: Done

source/materials/MaterialsList.cc
jmunozv: Done

jmunozv:
I've found in different ITO supplier web pages little changes in the density and weight composition of the material.
This was the reason for the original idea to implement the composition in number of atoms.
In any case, these little changes don't have any impact on our detector simulations, so I propose to approve the PR as it is, with all the suggestions from the reviewers implemented.

jmalbos (approved):
This PR updates the definition of ITO (to the best of our knowledge). Approved.

PR #83, opened 2021-02-05
Change xenon by air in Flex lab

jmunozv:
Simple fix that changes xenon by air in NextFlex lab.


paolafer (approved):
Simple PR, which fixes a wrong material in the lab volume.

PR #84, opened 2021-02-15
Fix DEMOpp-Run8 GATE - SiPMBoard distance

jmunozv:
Simple PR that fixes the GATE - SiPMBoard distance for DEMOpp geometry in its run8 configuration.


paolafer:
I understand that this number comes from private communication with experts?

jmunozv:
> I understand that this number comes from private communication with experts?

Yes. It does.

paolafer:
Therefore, there's not much we can say about it. I approve this PR.

paolafer (approved):

PR #85, opened 2021-02-15
Fix order in the tile definition

carmenromo:
A bug was introduced when the coordinates in the PetBox geometry were changed. This bug is related to the order in which the tiles are defined in the geometry, affecting the sensor identification. This PR fixes the problem changing the order in which the tiles are defined, both in the case of the asymmetric box and the symmetric one.

### DIFF
--- a/source/geometries/TileHamamatsuBlue.cc
+++ b/source/geometries/TileHamamatsuBlue.cc
@@ -111,7 +111,7 @@ namespace nexus {
          G4int copy_no = (j+1)*10 + i + 1;
          G4double x_pos = -tile_x_/2. + offset_x + sipm_dim.x()/2. + i * sipm_pitch_;
          G4double y_pos =  tile_y_/2. - offset_y - sipm_dim.y()/2. - j * sipm_pitch_;
-         G4double z_pos =  tile_z_/2. - quartz_thick_ - lxe_thick_ - sipm_dim.z()/2.;
+         G4double z_pos = (tile_z_+epoxy_depth_)/2. - epoxy_depth - sipm_dim.z()/2.;
          G4String vol_name = "SiPMHmtsuBlue_" + std::to_string(copy_no);
          new G4PVPlacement(0, G4ThreeVector(x_pos, y_pos, z_pos),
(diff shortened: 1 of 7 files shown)


source/geometries/TileHamamatsuBlue.cc
```diff
@@ -106,15 +106,15 @@ namespace nexus {
...
-	       G4int copy_no = i*2*n_columns_ + j + 1;
-         G4double x_pos = -tile_x_/2. + offset_x + sipm_dim.x()/2. + j * sipm_pitch_;
-         G4double y_pos = tile_y_/2. - offset_y - sipm_dim.y()/2. - i * sipm_pitch_;
-         G4double z_pos = (tile_z_+epoxy_depth_)/2. - epoxy_depth - sipm_dim.z()/2.;
+     for (int j=0; j<n_rows_; j++){
+       for (int i=0; i<n_columns_; i++){
+         G4int copy_no = (j+1)*10 + i + 1;
+         G4double x_pos = -tile_x_/2. + offset_x + sipm_dim.x()/2. + i * sipm_pitch_;
+         G4double y_pos =  tile_y_/2. - offset_y - sipm_dim.y()/2. - j * sipm_pitch_;
+         G4double z_pos =  tile_z_/2. - quartz_thick_ - lxe_thick_ - sipm_dim.z()/2.;
```
paolafer: The code is not compiling, because the following variables are not defined: `quartz_thick_`  and  `lxe_thick_`. Please, be careful and review the code: I don't think these variables belong to this class...

carmenromo: Fixed, sorry about that...

carmenromo:
The only problem with the FBK sensor ids is that in the sensor position table it is only saved the last sensor id that has been touched (since every 4 sensors share id), and the outcome is:

Captura de pantalla 2021-03-01 a las 12 25 20

However I think we can deal with it and take the sensor positions from the database.

paolafer (approved):
This PR sets the numbering of the sensors of the first set-up to the scheme used in data taking. The tests have been updated accordingly. I approve it.

PR #86, opened 2021-02-17
Turn cherenkov off

paolafer:
This PR adds a line to the examples using full simulation, to turn off the Cherenkov effect.
This is just a temporary fix, to deal with materials that don't have their optical properties defined in the whole range of wavelengths of the Cherenkov photons. It will be removed once we revisit the optical properties of the whole set of materials used in nexus.
The ChangeLog has also been updated for the next tag.


jmunozv (approved):
Approved

PR #88, opened 2021-02-24
New generic VUV-sensitive SiPMs.

paolafer:
In this PR, the old `SiPMpetVUV` class is being modified to make it configurable in size and with a protective window with a configurable refractive index. Moreover, the size of the active area has been made slightly smaller than the SiPM size, so that no photons can be detected coming from the sides.
It is used now as a default in the full-body PET geometry, instead of `SiPMpetFBK`, which is kept for now, but could be removed in the next future.

### DIFF
--- a/source/geometries/SiPMpetVUV.cc
+++ b/source/geometries/SiPMpetVUV.cc
@@ -32,7 +32,7 @@ namespace nexus {
 
   SiPMpetVUV::SiPMpetVUV(): BaseGeometry(),
 			    visibility_(0),
-			    refr_index_(1),
+			    refr_index_(1.6),
 			    eff_(1.),
                             time_binning_(1.*microsecond),
                             sensor_depth_(-1),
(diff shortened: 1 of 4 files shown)


source/geometries/SiPMpetVUV.cc
```diff
@@ -36,19 +34,28 @@ namespace nexus {
 			    visibility_(0),
 			    refr_index_(1),
```
carmenromo: Since the quartz refraction index is ~1.6, I would either change this default value or either modify the macro with the reasonable value of this property so that when someone executes the macro in the examples, they get a value closer to the true one.

paolafer: I did the former.

carmenromo (approved):
This PR modifies the `SiPMpetVUV` class to use a more reasonable configuration previously studied and discussed, it adds the last conventions for a sensitive detector and changes it to default in the full-body PET geometry. Moreover the code is very clear, so I approved!

PR #91, opened 2021-03-24
Cosmetics

paolafer:
This PR fixes a few descriptions of files which were overlooked in the last convention review. It also improves an error message when reading an illegal parameter.


jmunozv (approved):
I approve this PR that basically implements some cosmetics in sw and some outputs.

PR #92, opened 2021-03-25
Change PMT IDs to make them equal to data.

paolafer:
This PR changes the IDs of DEMO++ PMTs from 0--2 to 2--4. This way, they are the same as in data.


jmunozv (approved):
I approve this PR that accommodates the DEMOpp PMT numbering to that used in IC database.

PR #93, opened 2021-03-28
Refactor Back2backGammas::GeneratePrimaryVertex

jacg:
This PR thoroughly cleans up code generating back-to-back gammas.

Caveat: As the installation/development/testing environment I contributed to Nexus is not present on this branch, I have not had the time to actually compile, run and test this in its full glory and context (though I did verify the general principle in my own environment). Therefore someone else will have to compile, test, and fix any mistakes that might have crept in.

I have a suspicion that the original (pointlessly noisy) code has been copy-pasted in numerous places in Nexus (I have seen at least one with my own eyes). Moral of the story:

+ Don't copy-paste: abstract!
+ If you absolutely *must* copy-paste, then *think carefully* about the meaning of the code, and adapt it to your specific use-case, rather than bringing in truckloads of irrelevant junk.

### DIFF
--- a/source/generators/Back2backGammas.cc
+++ b/source/generators/Back2backGammas.cc
@@ -81,7 +81,7 @@ namespace nexus {
     G4PrimaryVertex* vertex =
         new G4PrimaryVertex(position, time);
 
-    G4ParticleDefinition* particle_definition =
+    G4ParticleDefinition* gamma =
       G4ParticleTable::GetParticleTable()->FindParticle("gamma");
 
     if (costheta_min_ != -1. || costheta_max_ != 1. || phi_min_ != 0. || phi_max_ != 2.*pi) {
--- a/source/generators/Back2backGammas.cc
+++ b/source/generators/Back2backGammas.cc
@@ -83,16 +83,13 @@ namespace nexus {
 
     G4ParticleDefinition* particle_definition =
       G4ParticleTable::GetParticleTable()->FindParticle("gamma");
-    // Set masses to PDG values
-    G4double mass = particle_definition->GetPDGMass();
     // Set charges to PDG value
     G4double charge = particle_definition->GetPDGCharge();
 
     G4ThreeVector momentum_direction = G4RandomDirection();
 
     // Calculate cartesian components of momentum
-    G4double energy = 510.999*keV + mass;
-    G4double pmod = std::sqrt(energy*energy - mass*mass);
+    G4double pmod = 510.999*keV;
     G4double px = pmod * momentum_direction.x();
     G4double py = pmod * momentum_direction.y();
     G4double pz = pmod * momentum_direction.z();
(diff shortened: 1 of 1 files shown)


jacg:
It seems that the GHA tests *do* run on the `petalo` branch, and they pass.

gonzaponte:
I see that you use brackets instead of parentheses for data constructors. Is there a specific motivation for that?

jacg:
It's been the recommended way to construct in C++ for almost a decade, since the introduction of *uniform initialization* in C++ 11. As all things C++ it's a complex issue (with pitfalls (though fewer than the pitfalls it avoids)) but, very briefly:

1. It makes a number of things possible which were previously impossible.
2. It avoids falling into some very hard-to-debug C++ traps.

Joe HEP Programmer, who can't be bothered to think too much about C++, should simply remember: Thou shalt always use braces rather than parentheses when invoking constructors.

Joe HEP Programmer should stop reading now, and work on remembering the rule. If you want to understand a bit more, read on.

# Details

Following this rule will, in the vast majority of cases, either

+ work exactly the same way as using parentheses, or
+ avoid a C++ horror that would waste many hours of Joe's time. 

The latter occurs mostly in situations where something that looks like a construction to Joe, is parsed as a function declaration by C++:

``` c++
Banana a_lovely_fresh_banana(); # Declaration of Banana-returning function
Banana a_lovely_fresh_banana{}; # Default construction of Banana
Banana a lovely_fresh_banana;   # What Joe SHOULD have written, pre-C++ 11
```

Joe tends to write/read the first line as default-constructing an object of type `Banana` into a variable `a_lovely_fresh_banana`, but C++ parses it as a declaration of a function called `a_lovely_fresh_banana` which has no parameters and returns a `Banana`. Much more subtle and nasty variants of this problem exist.

The compilation error points, not to this line, but to the first *usage* of `a_lovely_fresh_banana` and says something completely incomprehensible to Joe. Because Joe doesn't follow best practices such as

+ declaring variables as close to the first point of need, as possible;
+ NOT writing functions which are 2000 lines long

this incomprehensible error message will, on average, be pointing at a line which is 283 lines below the one which caused the problem. Therefore it is in Joe's best interests to avoid causing this problem in the first place. Therefore Joe should develop the habit of always constructing with braces rather than parentheses.

C++ and Joe will agree on the meaning of the second an third lines. But Joe will forget these subtleties, therefore Joe should develop the habit of constructing with braces.

In rare cases, replacing parentheses with braces will break something. The most obvious example I can think of, off the top of my head is

```c++
std::vector<int> a(5,2); # [2, 2, 2, 2, 2]
std::vector<int> b{5,2}; # [5, 2]
```

So Joe should continue to use this particular overload of the `std::vector` constructor with parentheses.

But then, Joe probably isn't aware of the existence of of the `(size_t, T)` constructor of `std::vector`, so it's a problem that is unlikely to arise.

Having spent some time digging around Geant4 source code recently, I get the impression that even Guru HEP Programmer is unaware of the existence of `std::vector`, let alone its constructors, or the concept of a C++ iterator (or any other number of basic C++ features and best practices), so Joe is also unlikely to encounter said constructor in the wild.


# TLDR

Always construct with braces rather than parentheses.

gonzaponte:
I wasn't aware of all that. Thanks for the detailed explanation.

PR #94, opened 2021-03-29
Review optical properties

paolafer:
This PR does a review of the optical properties used in the petalo branch.
1. Unused properties, which are not foreseen to be used in the PET environment have been deleted.
2. The same minimum and maximum energy have been set for all the properties, and this range covers the full scintillation spectrum of LXe + the emission spectrum of TPB.
3. Some properties, such as TPB or PTFE, have been updated to the most recent implementation of the master branch.
4. The argon properties have been left as they are, since they imply a review to extend the optical properties of the materials to a larger range. We'll do it if we need to simulate argon.

### DIFF
--- a/source/materials/OpticalMaterialProperties.cc
+++ b/source/materials/OpticalMaterialProperties.cc
@@ -13,7 +13,7 @@
 
 #include <G4MaterialPropertiesTable.hh>
 
-#include <assert.h>
+#include <cassert>
 
 using namespace nexus;
 using namespace CLHEP;
@@ -122,7 +122,7 @@ G4MaterialPropertiesTable* OpticalMaterialProperties::EpoxyLXeRefr()
   // ABSORPTION LENGTH
 
   std::vector<G4double> abs_energy = {optPhotMinE_, optPhotMaxE_};
-  std::vector<G4double> abs_length  = {noAbsLength_, noAbsLength_};
+  std::vector<G4double> abs_length = {noAbsLength_, noAbsLength_};
 
   assert(abs_energy.size() == abs_length.size());
   mpt->AddProperty("ABSLENGTH", abs_energy.data(), abs_length.data(), abs_energy.size());
--- a/source/materials/OpticalMaterialProperties.cc
+++ b/source/materials/OpticalMaterialProperties.cc
@@ -708,7 +708,7 @@ G4MaterialPropertiesTable* OpticalMaterialProperties::Pyrex_vidrasa()
 }
 
 
-G4MaterialPropertiesTable* OpticalMaterialProperties::TPB(G4double decay)
+G4MaterialPropertiesTable* OpticalMaterialProperties::TPB(G4double decay_time)
 {
 
   /// This is the simulation of the optical properties of TPB (tetraphenyl butadiene)
@@ -795,7 +795,7 @@ G4MaterialPropertiesTable* OpticalMaterialProperties::TPB(G4double decay)
                    WLS_emiSpectrum.data(), WLS_emi_energy.size());
 
   // WLS Delay
-  mpt->AddConstProperty("WLSTIMECONSTANT", decay);
+  mpt->AddConstProperty("WLSTIMECONSTANT", decay_time);
 
   // WLS Quantum Efficiency
   // According to the paper, the QE of TPB depends on the incident wavelength.
@@ -808,7 +808,7 @@ G4MaterialPropertiesTable* OpticalMaterialProperties::TPB(G4double decay)
 }
 
 
-G4MaterialPropertiesTable* OpticalMaterialProperties::TPB_LXe(G4double decay)
+G4MaterialPropertiesTable* OpticalMaterialProperties::TPB_LXe(G4double decay_time)
 {
   /// This is the simulation of the optical properties of TPB (tetraphenyl butadiene)
   /// a wavelength shifter which allows to converts VUV photons to blue photons.
@@ -909,7 +909,7 @@ G4MaterialPropertiesTable* OpticalMaterialProperties::TPB_LXe(G4double decay)
                    WLS_emiSpectrum.data(), WLS_emi_energy.size());
 
   // WLS Delay
-  mpt->AddConstProperty("WLSTIMECONSTANT", decay);
+  mpt->AddConstProperty("WLSTIMECONSTANT", decay_time);
 
   // WLS Quantum Efficiency
   // According to the paper, the QE of TPB depends on the incident wavelength.
@@ -922,7 +922,7 @@ G4MaterialPropertiesTable* OpticalMaterialProperties::TPB_LXe(G4double decay)
 }
 
 
-G4MaterialPropertiesTable* OpticalMaterialProperties::TPB_LXe_nconst(G4double decay)
+G4MaterialPropertiesTable* OpticalMaterialProperties::TPB_LXe_nconst(G4double decay_time)
 {
     /// This is the simulation of the optical properties of TPB (tetraphenyl butadiene)
   /// a wavelength shifter which allows to converts VUV photons to blue photons.
@@ -1010,7 +1010,7 @@ G4MaterialPropertiesTable* OpticalMaterialProperties::TPB_LXe_nconst(G4double de
                    WLS_emiSpectrum.data(), WLS_emi_energy.size());
 
   // WLS Delay
-  mpt->AddConstProperty("WLSTIMECONSTANT", decay);
+  mpt->AddConstProperty("WLSTIMECONSTANT", decay_time);
 
   // WLS Quantum Efficiency
   // According to the paper, the QE of TPB depends on the incident wavelength.
--- a/source/materials/OpticalMaterialProperties.cc
+++ b/source/materials/OpticalMaterialProperties.cc
@@ -332,27 +332,32 @@ G4MaterialPropertiesTable* OpticalMaterialProperties::GAr(G4double sc_yield)
   // REFRACTIVE INDEX
   const G4int ri_entries = 200;
 
-  G4double ri_energy[ri_entries];
+  std::vector<G4double> ri_energy;
   for (int i=0; i<ri_entries; i++) {
-    ri_energy[i] = (1 + i*0.049)*eV;
+    ri_energy.push_back((1 + i*0.049)*eV);
   }
 
-  G4double rindex[ri_entries];
+  std::vector<G4double> ri_index;
 
   for (int i=0; i<ri_entries; i++) {
     G4double lambda = h_Planck*c_light/ri_energy[i]*1000; // in micron
-    rindex[i] = 1 + 0.012055*(0.2075*pow(lambda,2)/(91.012*pow(lambda,2)-1) +
-    0.0415*pow(lambda,2)/(87.892*pow(lambda,2)-1) + 4.3330*pow(lambda,2)/(214.02*pow(lambda,2)-1)); // From refractiveindex.info
+    ri_index.push_back(1 + 0.012055*(0.2075*pow(lambda,2)/(91.012*pow(lambda,2)-1) +
+                                     0.0415*pow(lambda,2)/(87.892*pow(lambda,2)-1) +
+                                     4.3330*pow(lambda,2)/(214.02*pow(lambda,2)-1)));
+    // From refractiveindex.info
     //    std::cout << "rindex = " << rindex[i] << std::endl;
   }
+
+  assert(ri_energy.size() == ri_index.size());
+  mpt->AddProperty("RINDEX", ri_energy.data(), ri_index.data(), ri_energy.size());
 
   // EMISSION SPECTRUM
   // Sampling from ~110 nm to 150 nm <----> from ~11.236 eV to 8.240 eV
   const G4int sc_entries = 380;
-  G4double sc_energy[sc_entries];
-  G4double intensity[sc_entries];
+  std::vector<G4double> sc_energy;
+  std::vector<G4double> intensity;
 
-  G4double Wavelength_peak = 128.*nm;
+  G4double Wavelength_peak  = 128.*nm;
   G4double Wavelength_sigma = 2.929*nm;
 
   G4double Energy_peak = (h_Planck*c_light/Wavelength_peak);
@@ -361,23 +366,24 @@ G4MaterialPropertiesTable* OpticalMaterialProperties::GAr(G4double sc_yield)
   //  std::cout << "Energy_peak = " << Energy_peak << std::endl;
 
   for (int j=0;j<sc_entries;j++){
-    sc_energy[j] = 8.240*eV + 0.008*j*eV;
-    intensity[j] = exp(-pow(Energy_peak/eV-sc_energy[j]/eV,2)/(2*pow(Energy_sigma/eV, 2)))/(Energy_sigma/eV*sqrt(pi*2.));
+    sc_energy.push_back(8.240*eV + 0.008*j*eV);
+    intensity.push_back(exp(-pow(Energy_peak/eV-sc_energy[j]/eV,2)/(2*pow(Energy_sigma/eV, 2)))/(Energy_sigma/eV*sqrt(pi*2.)));
     //    std::cout << "(energy, intensity) = (" << sc_energy[j] << "," << intensity[j] << ")" << std::endl;
   }
 
+  assert(sc_energy.size() == intensity.size());
+  mpt->AddProperty("SLOWCOMPONENT", sc_energy.data(), intensity.data(), sc_energy.size());
+  mpt->AddProperty("FASTCOMPONENT", sc_energy.data(), intensity.data(), sc_energy.size());
+  mpt->AddProperty("ELSPECTRUM", sc_energy.data(), intensity.data(), sc_energy.size());
+
   // ABSORTION LENGTH
-  G4double energy[2] = {0.01*eV, 100.*eV};
-  G4double abslen[2] = {1.e8*m, 1.e8*m};
-  mpt->AddProperty("ABSLENGTH", energy, abslen, 2);
+  std::vector<G4double> energy = {0.01*eV, 100.*eV};
+  std::vector<G4double> abslen = {1.e8*m, 1.e8*m};
+  mpt->AddProperty("ABSLENGTH", energy.data(), abslen.data(), energy.size());
 
 //An argon gas proportional scintillation counter with UV avalanche photodiode scintillation readout C.M.B. Monteiro, J.A.M. Lopes, P.C.P.S. Simoes, J.M.F. dos Santos, C.A.N. Conde
 
-  mpt->AddProperty("RINDEX", ri_energy, rindex, ri_entries);
-  mpt->AddProperty("FASTCOMPONENT", sc_energy, intensity, sc_entries);
-  mpt->AddProperty("SLOWCOMPONENT", sc_energy, intensity, sc_entries);
   mpt->AddConstProperty("SCINTILLATIONYIELD", sc_yield);
-  mpt->AddProperty("ELSPECTRUM", sc_energy, intensity, sc_entries);
   mpt->AddConstProperty("FASTTIMECONSTANT",6.*ns);
   mpt->AddConstProperty("SLOWTIMECONSTANT",37.*ns);
   mpt->AddConstProperty("YIELDRATIO",.52);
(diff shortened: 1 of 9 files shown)


source/materials/OpticalMaterialProperties.cc
```diff
@@ -74,10 +74,8 @@ G4MaterialPropertiesTable* OpticalMaterialProperties::EpoxyFixedRefr(G4double n)
 
   XenonLiquidProperties LXe_prop;
 
-  const G4int ri_entries = 18;
-  G4double ri_energy[ri_entries] = {optPhotMinE_, 2*eV, 3*eV, 4*eV, 5*eV, 6*eV, 6.2*eV,
-                                    6.4*eV, 6.6*eV, 6.8*eV, 7*eV, 7.2*eV, 7.4*eV,
-                                    7.6*eV, 7.8*eV, 8*eV, 8.2*eV, optPhotMaxE_};
+  const G4int ri_entries = 2;
+  G4double ri_energy[ri_entries] = {optPhotMinE_, optPhotMaxE_};
```
jacg: Please eradicate all use of C-arrays throughout the code base. In almost every case you'll want to use `std::vector` instead. This has been the case since before 1998 and even Geant4 has now understood this (in some places) in 10.7!

paolafer: Yes, this is something to be done. What about putting it off to a following PR, when we update nexus to GEANT4.10.07? This way we will be able to use directly the new signature of the `AddProperty()` function.

jacg: Are there any plans to move to 10.7?

I would say that it is reasonable to leave any C-array using lines that are not directly touched by this PR. But for any lines that *are* modified, I recommend that the C-arrays be changed to `std::vector` right now. This will require changing the declaration, and the call to `AddProperty`, so 2 or 3 other lines. If it turns out to require changing dozens of lines, then leave it, but if it's only a few, please do it now.

We can be pretty sure that 

+ A concerted effort to replace C-arrays with `std::vector` in a subsequent PR will never happen, because we'll always have something more important to do. (Repaying technical debt isn't very sexy.)
+ People will end up copy-pasting this. So let's give them a chance to copy-paste the correct solution, rather than the wrong one. Let's plant the seed of the correct solution. You know what the correct solution is, because I told you; everyone else on the project will think that the C-array is the correct solution, because nowhere is there any indication to the contrary: the same mistake is repeated *everywhere* in the project. Here is our chance to record, in at least one place in the project, what the correct approach is.

source/materials/OpticalMaterialProperties.cc
```diff
@@ -13,6 +13,8 @@
 
 #include <G4MaterialPropertiesTable.hh>
 
+#include <assert.h>
```
jacg: `<cassert>`

Generally speaking you should not include any standard headers that end in `.h`: these are *PRE-1998-STANDARD*! Their use has been deprecated for almost a quarter of a century.

source/materials/OpticalMaterialProperties.cc
```diff
@@ -102,27 +104,28 @@ G4MaterialPropertiesTable* OpticalMaterialProperties::EpoxyLXeRefr()
...
+  assert(ri_energy.size() == ri_index.size());
+  mpt->AddProperty("RINDEX", ri_energy.data(), ri_index.data(), ri_energy.size());
+
   // ABSORPTION LENGTH
-  const G4int abs_entries = 2;
 
-  G4double abs_energy[abs_entries] = {optPhotMinE_, optPhotMaxE_};
-  G4double abs_length[abs_entries]  = {noAbsLength_, noAbsLength_};
+  std::vector<G4double> abs_energy = {optPhotMinE_, optPhotMaxE_};
+  std::vector<G4double> abs_length  = {noAbsLength_, noAbsLength_};
```
jacg: Extra space before `=`.

source/materials/OpticalMaterialProperties.cc
```diff
@@ -21,16 +23,19 @@ G4MaterialPropertiesTable* OpticalMaterialProperties::Vacuum()
...
-  G4double energy[entries] = {optPhotMinE_, optPhotMaxE_};
+  std::vector<G4double> energy = {optPhotMinE_, optPhotMaxE_};
 
   // REFRACTIVE INDEX
-  G4double rindex[entries] = {1., 1.};
-  mpt->AddProperty("RINDEX", energy, rindex, entries);
+  std::vector<G4double> ri_index = {1., 1.};
+
+  assert(energy.size() == ri_index.size());
+  mpt->AddProperty("RINDEX", energy.data(), ri_index.data(), energy.size());
```
jacg, line 32: This pattern is repeated 54 times in this file! It's begging for a function which performs the check before forwarding to `AddProperty`. Give the function an interface resembling the 10.7's `AddProperty`, and you'll make the migration to 10.7 trivial.

The good news is that Geant4 will take care of performing the check for you in 10.7: https://gitlab.cern.ch/geant4/geant4/-/blob/master/source/materials/src/G4MaterialPropertiesTable.cc#L304

paolafer: If the check is included in the `AddProperty` function in 10.7, I don't see the point of writing a function now to optimize a piece of code that will become obsolete soon.

jacg: You wrote out this pattern by hand *54 times*!

After you wrote it out 3 or 4 times (ideally, after the second time) you should have said to yourself, "I *refuse* to type this out by hand again!" and you should have written the abstraction.

As it stands, there is a small but greater than zero chance that some of your asserts check different values than are passed to `AddProperty`. (*I* certainly couldn't be bothered to check them all! Even if I did, it's so repetitively tedious, that I'd be very likely to miss any mistakes.) By using the abstraction, you reduce the probability of such a mistake to strictly zero. As these are run-time rather than compile-time checks, they are more difficult to spot, and have consequences that are nastier to deal with, so it is especially important to avoid the mistake.

Using the abstraction will shorten the code by 54 lines, while writing the abstraction will cost 3 non-brace-only lines.

You are expecting this code to be updated for 10.7. Think of the poor soul who will have to do it, and will be faced with 54 instances of 

``` c++
assert(energy.size() == ri_index.size());
mpt->AddProperty("RINDEX", energy.data(), ri_index.data(), energy.size());
```

which will need to be changed to 

``` c++
mpt->AddProperty("RINDEX", energy, ri_index);
```

Why not be nice to that person (probably you) and leave 54 instances of 

``` c++
AddProperty(mpt, "RINDEX", energy, ri_index);
```

Better still, with the abstraction in place, when the time comes to switch to 10.7, you can change the *definition* of your abstraction (removing the assert, and switching to the `std::vector`-using overload) just once, and leave its 54 uses untouched, and you'll have all the benefits of the 10.7 version at the cost of changing 2 lines. As it stands, switching to the 10.7 version will require changing 108 lines.

You may argue that there is no point in changing 108 lines *now* to save changing 108 lines later. There are two answers to this:

1. It is now fresh in your mind and you know exactly what is happening. Leave it in, and you will have to re-establish the mental context, months down the line; or, worse still, someone *else* will have to acquire the mental context. Doing it now is cheaper than doing it in the future.

2. You shouldn't have written out 54 repetitions of something that will require 54 fixes, in the first place. It's our old friend copy-paste: just don't do it!

In summary, it's simply a very basic example of good programming practice that you should follow, because it is good for you and anyone else whose life is touched by the code. It is a *habit* that you should develop, because it will result in you writing more robust code, more quickly, that is more pleasant to work with both for you and for your collaborators.

Put another way, *I* don't see the point of writing bad code, when it's so easy to write better code.

Important point: you should have refused to write the assert out 54 times, in the first place!

Important point to *everyone*: please *refuse* to write out the same tedious stuff, over and over and over again. Think to yourselves: what interface would make my life easier? ... and then *create* that interface.

**Copy paste? JUST SAY NO!**

paolafer: > You may argue that there is no point in changing 108 lines now to save changing 108 lines later.

That was exactly my point :-).

I agree with you that I shouldn't have made the effort of writing 54 times the same pattern. In fact, I shouldn't have changed the code at all at this point and leave the implementation of vectors instead of C-arrays for the update to Geant4.10.7, but your point of being educational convinced me.

I'm not sure it is better to be left with 

```
AddProperty(mpt, "RINDEX", energy, ri_index);
```

once we pass to Geant4.10.7, rather than the current

```
mpt->AddProperty("RINDEX", energy, ri_index);
```

which seems more straightforward to me.

Anyway, I would finish the review of this PR on the physics content (@carmenromo is assigned) and I would leave for another PR any implementation you would like to do.

jacg: Fine. I'll decouple from this PR.

source/materials/OpticalMaterialProperties.cc
```diff
@@ -42,99 +42,86 @@ G4MaterialPropertiesTable* OpticalMaterialProperties::Epoxy()
...
 
   return mpt;
 }
 
 G4MaterialPropertiesTable* OpticalMaterialProperties::EpoxyFixedRefr(G4double n)
 {
-  // Optical properties adjusted in order not to modify the trajectories of photons and
-  // not to absorb them before they meet the photocathode. To be set to real properties when available
+  // Optical properties adjusted in order not to absorb photons
+  // before they meet the photocathode.
```
carmenromo: What do you mean by photocathode here?

source/materials/OpticalMaterialProperties.cc
```diff
@@ -20,16 +23,19 @@ G4MaterialPropertiesTable* OpticalMaterialProperties::Vacuum()
...
-  const G4int entries = 2;
-  G4double energy[entries] = {0.01*eV, 100.*eV};
+  std::vector<G4double> energy = {optPhotMinE_, optPhotMaxE_};
 
   // REFRACTIVE INDEX
-  G4double rindex[entries] = {1., 1.};
-  mpt->AddProperty("RINDEX", energy, rindex, entries);
+  std::vector<G4double> ri_index = {1., 1.};
+
+  assert(energy.size() == ri_index.size());
```
carmenromo, line 31: It is not clear to me the reason of these `asserts`. If both vectors are defined with 2 elements just above, in which case should them fail? Is this to be sure that `optPhotMinE_` and `optPhotMaxE_` are correctly set?

source/materials/OpticalMaterialProperties.cc
```diff
@@ -42,119 +48,105 @@ G4MaterialPropertiesTable* OpticalMaterialProperties::Epoxy()
...
      3.973*eV, 4.120*eV, 4.224*eV, 4.320*eV, 4.420*eV,
-     5.018*eV, 8.*eV};
-  G4double abslength[abs_entries] =
+     5.018*eV, 8.*eV, optPhotMaxE_};
+  std::vector<G4double> abs_length =
     {15000.*cm, 326.*mm, 117.68*mm, 85.89*mm, 50.93*mm,
      31.25*mm, 17.19*mm, 10.46*mm, 5.26*mm, 3.77*mm,
      2.69*mm, 1.94*mm, 1.94*mm, 1.94*mm, 1.94*mm,
-     1.94*mm, 1.94*mm};
+     1.94*mm, 1.94*mm, 1.94*mm};
```
carmenromo: If 7 of the values of `abs_length` are the same, it is not worthy to reduce the length of the vector and avoid repeating values?

source/materials/OpticalMaterialProperties.cc
```diff
@@ -270,240 +262,79 @@ G4MaterialPropertiesTable* OpticalMaterialProperties::FakeGenericMaterial(G4doub
...
-
-
-  return mpt;
-
-}
-
 
 G4MaterialPropertiesTable* OpticalMaterialProperties::GAr(G4double sc_yield)
 {
+  // WARNING: before using GAr properties, check that we know the
```
carmenromo, line 328: Is this misalignment on purpose?

source/materials/OpticalMaterialProperties.cc
```diff
@@ -649,37 +477,47 @@ G4MaterialPropertiesTable* OpticalMaterialProperties::GXe(G4double pressure,
...
-  const G4int ri_entries = 9;
-  G4double ri_energy[ri_entries]
-    = {1*eV, 2*eV, 3*eV, 4*eV, 5*eV, 6*eV, 7*eV, 8*eV, 9*eV};
-
-  // G4int ri_entries = 15;
-  // G4double ri_energy[ri_entries]
-  //   = {1*eV, 2*eV, 3*eV, 4*eV, 5*eV, 6*eV, 6.2*eV, 6.4*eV, 6.6*eV, 6.8*eV, 7*eV, 7.2*eV, 7.4*eV, 7.6*eV, 7.8*eV};
-
-  G4double rindex[ri_entries];
+  const G4int ri_entries = 200;
```
carmenromo, line 489: Why did you increase the number of entries so much?

source/materials/OpticalMaterialProperties.cc
```diff
@@ -736,300 +580,94 @@ G4MaterialPropertiesTable* OpticalMaterialProperties::LXe()
...
 {
   XenonLiquidProperties LXe_prop;
   G4MaterialPropertiesTable* LXe_mpt = new G4MaterialPropertiesTable();
 
-  const G4int ri_entries = 18;
-  G4double ri_energy[ri_entries]
-    = {1*eV, 2*eV, 3*eV, 4*eV, 5*eV, 6*eV, 6.2*eV, 6.4*eV, 6.6*eV, 6.8*eV,
-       7*eV, 7.2*eV, 7.4*eV, 7.6*eV, 7.8*eV, 8*eV, 8.2*eV, 8.21*eV};
+  std::vector<G4double> ri_energy = {optPhotMinE_, optPhotMaxE_};
+  std::vector<G4double> ri_index  = {1.7, 1.7};
```
carmenromo: Can you provide a reference where that `ri_index` value comes from? I thought it was 1.69...

source/materials/OpticalMaterialProperties.cc
```diff
@@ -1042,268 +680,349 @@ G4MaterialPropertiesTable* OpticalMaterialProperties::Pyrex_vidrasa()
...
-  pyrex_mpt->AddProperty("ABSLENGTH", ri_energy, abs_length, ri_entries);
+  assert(ri_energy.size() == abs_length.size());
+  pyrex_mpt->AddProperty("ABSLENGTH", ri_energy.data(), abs_length.data(), ri_energy.size());
 
   return pyrex_mpt;
 }
 
 
-G4MaterialPropertiesTable* OpticalMaterialProperties::TPB(G4double pressure, G4double temperature)
+G4MaterialPropertiesTable* OpticalMaterialProperties::TPB(G4double decay)
```
carmenromo: What this `decay` parameter stands for? It refers to time, right? I would specify it, but since this comes from the master perhaps this isn't worth it...

source/materials/XenonGasProperties.cc
```diff
@@ -239,6 +239,15 @@ namespace nexus {
 
 
 
+  void XenonGasProperties::Scintillation
```
carmenromo: Why are there two methods `void XenonGasProperties::Scintillation` and also `void XenonLiquidProperties::Scintillation`?

source/materials/XenonLiquidProperties.cc
```diff
@@ -124,5 +124,13 @@ namespace nexus {
 
 
 
+  void XenonLiquidProperties::Scintillation
+  (std::vector<G4double>& energy, std::vector<G4double>& intensity)
+  {
+    for (G4int i=0; i<energy.size(); i++)
```
carmenromo: When compiling, this warning shows up in both `XenonLiquidProperties.cc` and `XenonGasProperties.cc`:

`warning: comparison of integers of different signs: 'G4int' (aka 'int') and 'std::__1::vector<double, std::__1::allocator<double> >::size_type' (aka 'unsigned long') [-Wsign-compare]`

jacg: Replace `G4int` with `unsigned`.

source/materials/OpticalMaterialProperties.cc
paolafer: In this case, it won't fail, but if I had made a mistake in the size of the vectors, the compiler wouldn't complain, and Geant4 would not, either (it would just cut the longer vector to the same size of the other one) leading to an incorrect behaviour. Instead, with this assert, the simulation crashes on run time with an error message that points one here. It's a test also for future modifications of the code.

source/materials/OpticalMaterialProperties.cc
paolafer: To model the property better. It is more relevant in the case of the scintillation spectrum; for the refraction index I just followed what is done in the `master` branch.

source/materials/OpticalMaterialProperties.cc
paolafer: It's the decay constant of the TPB, the characteristic time of the re-emission of the photons. I changed it to `decay time`.

source/materials/XenonGasProperties.cc
paolafer: Because the scintillation spectrum is different in gaseous and liquid xenon.

source/materials/OpticalMaterialProperties.cc
paolafer: The refractive index at the scintillation peak is 1.69. This property was used years ago for some studies and `1.7` was used (a worse approximation, but still similar to the most probable value). I hadn't changed it, because this property is not used anymore, but I will, since I agree that `1.69` is more accurate.

source/materials/OpticalMaterialProperties.cc
paolafer: The choice of the word is not great, it should mean "active area". However, I've changed the description, since it doesn't reflect the current implementation of that property.

source/materials/XenonGasProperties.cc
carmenromo: No, but there are in total four methods:
- `void XenonGasProperties::Scintillation`
- `void XenonGasProperties::Scintillation`
- `void XenonLiquidProperties::Scintillation`
- `void XenonLiquidProperties::Scintillation`

paolafer: I see what you mean, that we should delete the old signature 

```
Scintillation
  (G4int entries, G4double* energy, G4double* intensity)
```

right? Will do.

paolafer: However, we still have `GAr` which has the old format. I either would keep both signatures or update the GAr property.

carmenromo: > right? Will do.

Yes, thanks!

> However, we still have GAr which has the old format. I either would keep both signatures or update the GAr property.

But will it work?

jacg: Hint: those are C-arrays in that signature!

paolafer: > But will it work?

Yes, in fact there are three functions with the same name, but they have different signatures and the correct one will be used in the right place, according to the arguments passed.

carmenromo: > Hint: those are C-arrays in that signature!

What do you mean @jacg ?

jacg: Signatures like 

``` c++
fn(int N, double* this, double* that);
```

(containing pointers to the start of contiguously-allocated chunks of homogeneous memory, and an accompanying int to tell us how much memory the pointer points to, AKA C-arrays) are what you would write in C. In C++ *you should not use C-arrays*! (For a plethora of well-known and oft-repeated reasons that I do not wish to rehash here.) C++ admits them (and makes `std::vector` trivially compatible with them), in case you want to communicate with a legacy C-API, but you should not write your own functions with such signatures (unless it's a callback you're going to pass to C).

The fact that Geant4 perpetuates interfaces that have been deprecated for more than a quarter of a century, is no excuse for us to do the same (rule of thumb: if G4 does it, it's probably a bad idea (and I'm only half-joking)) ... and even Geant4 has started to wake up to the fact in 10.7!

Remove that signature, hunt down and eradicate others like it that might be lurking in the code (hope and pray that no clients have already started using them), and don't be tempted to write any more like it!

carmenromo: I see!! Thanks a lot! :)

source/materials/XenonLiquidProperties.cc
```diff
@@ -127,7 +127,7 @@ namespace nexus {
   void XenonLiquidProperties::Scintillation
   (std::vector<G4double>& energy, std::vector<G4double>& intensity)
   {
-    for (G4int i=0; i<energy.size(); i++)
+    for (unsigned int i=0; i<energy.size(); i++)
```
carmenromo: I think you should delete `int` here

jacg: Only for aesthetic purposes; it makes no difference to the semantics: https://en.cppreference.com/w/cpp/language/types

carmenromo: Nice! I was not sure!

carmenromo (approved):
That's all from my side, this PR performs a great review of the optical properties: it updates the properties of some materials, deletes the unused ones and updates the code removing the C-arrays. Approved!

source/materials/XenonLiquidProperties.cc
jacg: Actually, I *should* have told you to use `size_t`, which is the type specifically designed for this purpose. Sorry, I wasn't fully concentrating.

PR #95, opened 2021-03-29
Refactor generators

paolafer:
This PR adds to the petalo branch a feature introduced recently in the master, namely, it extracts the functions to generate a random value of energy and momentum direction to a separate file. This way, they can be used in different parts of the code.
The code of two generators has been changed accordingly to use the new functions.

### DIFF
--- a/source/generators/Back2backGammas.cc
+++ b/source/generators/Back2backGammas.cc
@@ -79,9 +79,9 @@ void Back2backGammas::GeneratePrimaryVertex(G4Event* evt)
     G4ParticleTable::GetParticleTable()->FindParticle("gamma");
 
   auto p = 510.999 * keV * (
-                            (costheta_min_ != -1. || costheta_max_ != 1. || phi_min_ != 0. || phi_max_ != 2.*pi) ?
-                            Direction(costheta_min_, costheta_max_, phi_min_, phi_max_)                   :
-                            G4RandomDirection());
+           (costheta_min_ != -1. || costheta_max_ != 1. || phi_min_ != 0. || phi_max_ != 2.*pi) ?
+           Direction(costheta_min_, costheta_max_, phi_min_, phi_max_)                          :
+           G4RandomDirection());
 
   G4ThreeVector position = geom_->GenerateVertex(region_);
   G4double time = 0.;
--- a/source/generators/Back2backGammas.cc
+++ b/source/generators/Back2backGammas.cc
@@ -28,72 +28,70 @@
 #include <TFile.h>
 #include <TH1F.h>
 
-namespace nexus {
 
-  using namespace CLHEP;
+using namespace CLHEP;
+using namespace nexus;
 
-  Back2backGammas::Back2backGammas() : geom_(0), costheta_min_(-1.),
-                                       costheta_max_(1.),
-				       phi_min_(0.), phi_max_(2.*pi)
-  {
-    G4cout << "Limits = " << std::numeric_limits<unsigned int>::max() << G4endl;
-    /// For the moment, only random direction are allowed. To be fixes if needed
-     msg_ = new G4GenericMessenger(this, "/Generator/Back2back/",
-    "Control commands of 511-keV back to back gammas generator.");
+Back2backGammas::Back2backGammas() : geom_(0), costheta_min_(-1.),
+                                     costheta_max_(1.),
+                                     phi_min_(0.), phi_max_(2.*pi)
+{
+  G4cout << "Limits = " << std::numeric_limits<unsigned int>::max() << G4endl;
+  /// For the moment, only random direction are allowed. To be fixes if needed
+  msg_ = new G4GenericMessenger(this, "/Generator/Back2back/",
+                                "Control commands of 511-keV back to back gammas generator.");
 
-     msg_->DeclareProperty("region", region_,
-			   "Set the region of the geometry where the vertex will be generated.");
+  msg_->DeclareProperty("region", region_,
+                        "Set the region of the geometry where the vertex will be generated.");
 
-     msg_->DeclareProperty("min_costheta", costheta_min_,
-			   "Set minimum cosTheta for the direction of the particle.");
-     msg_->DeclareProperty("max_costheta", costheta_max_,
-			   "Set maximum cosTheta for the direction of the particle.");
-     msg_->DeclareProperty("min_phi", phi_min_,
-			   "Set minimum phi for the direction of the particle.");
-     msg_->DeclareProperty("max_phi", phi_max_,
-			   "Set maximum phi for the direction of the particle.");
+  msg_->DeclareProperty("min_costheta", costheta_min_,
+                        "Set minimum cosTheta for the direction of the particle.");
+  msg_->DeclareProperty("max_costheta", costheta_max_,
+                        "Set maximum cosTheta for the direction of the particle.");
+  msg_->DeclareProperty("min_phi", phi_min_,
+                        "Set minimum phi for the direction of the particle.");
+  msg_->DeclareProperty("max_phi", phi_max_,
+                        "Set maximum phi for the direction of the particle.");
 
-     DetectorConstruction* detconst = (DetectorConstruction*) G4RunManager::GetRunManager()->GetUserDetectorConstruction();
-    geom_ = detconst->GetGeometry();
+  DetectorConstruction* detconst = (DetectorConstruction*) G4RunManager::GetRunManager()->GetUserDetectorConstruction();
+  geom_ = detconst->GetGeometry();
 
-    theta_angle_ = new TH1F("CosTheta", "CosTheta", 180, -2, 2);
-    theta_angle_->GetXaxis()->SetTitle("Cos(theta)");
-    theta_angle_->GetYaxis()->SetTitle("Entries");
+  theta_angle_ = new TH1F("CosTheta", "CosTheta", 180, -2, 2);
+  theta_angle_->GetXaxis()->SetTitle("Cos(theta)");
+  theta_angle_->GetYaxis()->SetTitle("Entries");
 
-    phi_angle_ = new TH1F("Phi", "Phi", 360, -pi-2, pi+2);
-    phi_angle_->GetXaxis()->SetTitle("Phi (rad)");
-    phi_angle_->GetYaxis()->SetTitle("Entries");
-  }
+  phi_angle_ = new TH1F("Phi", "Phi", 360, -pi-2, pi+2);
+  phi_angle_->GetXaxis()->SetTitle("Phi (rad)");
+  phi_angle_->GetYaxis()->SetTitle("Entries");
+}
 
-  Back2backGammas::~Back2backGammas()
-  {
-    out_file_ = new TFile("GenerationAngles.root", "recreate");
-    theta_angle_->Write();
-    phi_angle_->Write();
-    out_file_->Close();
-  }
+Back2backGammas::~Back2backGammas()
+{
+  out_file_ = new TFile("GenerationAngles.root", "recreate");
+  theta_angle_->Write();
+  phi_angle_->Write();
+  out_file_->Close();
+}
 
-  void Back2backGammas::GeneratePrimaryVertex(G4Event* evt)
-  {
-    G4ParticleDefinition* gamma =
-      G4ParticleTable::GetParticleTable()->FindParticle("gamma");
+void Back2backGammas::GeneratePrimaryVertex(G4Event* evt)
+{
+  G4ParticleDefinition* gamma =
+    G4ParticleTable::GetParticleTable()->FindParticle("gamma");
 
-    auto p = 510.999 * keV * (
-      (costheta_min_ != -1. || costheta_max_ != 1. || phi_min_ != 0. || phi_max_ != 2.*pi) ?
-      nexus::Direction(costheta_min_, costheta_max_, phi_min_, phi_max_)                   :
-      G4RandomDirection());
+  auto p = 510.999 * keV * (
+                            (costheta_min_ != -1. || costheta_max_ != 1. || phi_min_ != 0. || phi_max_ != 2.*pi) ?
+                            Direction(costheta_min_, costheta_max_, phi_min_, phi_max_)                   :
+                            G4RandomDirection());
 
-    G4ThreeVector position = geom_->GenerateVertex(region_);
-    G4double time = 0.;
-    auto vertex = new G4PrimaryVertex(position, time);
+  G4ThreeVector position = geom_->GenerateVertex(region_);
+  G4double time = 0.;
+  auto vertex = new G4PrimaryVertex(position, time);
 
-    vertex->SetPrimary(new G4PrimaryParticle(gamma,  p.x(),  p.y(),  p.z()));
-    vertex->SetPrimary(new G4PrimaryParticle(gamma, -p.x(), -p.y(), -p.z()));
+  vertex->SetPrimary(new G4PrimaryParticle(gamma,  p.x(),  p.y(),  p.z()));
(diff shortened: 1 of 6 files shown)


source/generators/SingleParticle.cc
```diff
@@ -145,9 +145,9 @@ void SingleParticle::GeneratePrimaryVertex(G4Event* event)
     pz = pmod * momentum_Z_/mom_mod;
   } else if (costheta_min_ != -1. || costheta_max_ != 1. || phi_min_ != 0. || phi_max_ !=2.*pi) {
     G4ThreeVector p = nexus::Direction(costheta_min_, costheta_max_, phi_min_, phi_max_);
-    px = p.x();
-    py = p.y();
-    pz = p.z();
+    px = p.x() * pmod;
```
jacg: Why not use scalar-vector multiplication to scale the whole vector in one go, rather than doing it component by copmonent?

jacg:
Do I understand correctly that the implementation of `RandomEnergy` is being copy-pasted from `master` to `petalo`?

jacg:
Why, oh why, was the implementation on `master` implemented as `inline` in the header file, in spite of being 20 lines long and doing nontrivial computation?

source/generators/SingleParticle.cc
paolafer: Sure, that would be better! What do you suggest?

paolafer:
> Do I understand correctly that the implementation of `RandomEnergy` is being copy-pasted from `master` to `petalo`?

The idea is to use the same functions contained in RandomUtils in both branches, that's why I cherry picked the commit of the `master` branch that added that file.

paolafer:
> Why, oh why, was the implementation on `master` implemented as `inline` in the header file, in spite of being 20 lines long and doing nontrivial computation?

No particular reason. Do you have a suggestion of how to do it better?

jacg:
>> Do I understand correctly that the implementation of RandomEnergy is being copy-pasted from master to petalo?

> The idea is to use the same functions contained in RandomUtils in both branches, that's why I cherry picked the commit of the master branch that added that file.

You do understand that cherry picking is just a fancy Git trick to do a copy-paste (whose evils we are, right now, discussing elsewhere)?

>> Why, oh why, was the implementation on master implemented as inline in the header file, in spite of being 20 lines long and doing nontrivial computation?

> No particular reason. Do you have a suggestion of how to do it better?

I'm puzzled by the question. Do you understand the point/meaning/consequences of declaring a function `inline`?

But to answer your question explicitly: How about *the standard way to provide functions*? That is, declare its interface (signature) in a header file (`.h`/`.hh`/`.hpp`/etc.), and provide its implementation in an implementation file (`.cc`/`.cpp`/etc.).

paolafer:
> But to answer your question explicitly: How about _the standard way to provide functions_? That is, declare its interface (signature) in a header file (`.h`/`.hh`/`.hpp`/etc.), and provide its implementation in an implementation file (`.cc`/`.cpp`/etc.).

I think that the idea was to avoid writing an implementation file, but, as you point out, one of the functions is already quite complex, so it's probably better to write a `.cc` file.

source/generators/SingleParticle.cc
jacg: Something along the lines of what I submitted in the other PR:
https://github.com/next-exp/nexus/blob/6829fcfc397f405808f21e746339bd873290b9b1/source/generators/Back2backGammas.cc#L82-L97

Furthermore, once an acceptable `cosTheta` has been found, it's probably pointlessly wasteful to throw it away if `phi` is inadequate, so there should probably be a second, inner infinite loop to find `phi` without throwing away `cosTheta`.

paolafer: On the other hand, since we need the three components individually to be used in the instantiation of `G4PrimaryParticle`, maybe it's more convenient to leave it component by component.

jacg: That's why vectors have component accessors.

source/utils/RandomUtils.cc
```diff
@@ -33,11 +33,14 @@ namespace nexus {
         G4double sinTheta2 = 1. - cosTheta*cosTheta;
         if (sinTheta2 < 0.)  sinTheta2 = 0.;
         G4double sinTheta  = std::sqrt(sinTheta2);
-        G4double phi = CLHEP::twopi*G4UniformRand();
-        if (phi > phi_min && phi < phi_max){
-          return G4ThreeVector (sinTheta*std::cos(phi),
-                                sinTheta*std::sin(phi),
-                                cosTheta).unit();
+        G4bool phi_component = false;
+        while (phi_component == false) {
```
jacg: Is there any reason why these two lines can't be replaced with `while (true) {` ?

jacg: ... and the same goes for a similar situation in the outer loop and the `dir` flag.

paolafer: Right, done.

source/generators/Back2backGammas.cc
```diff
@@ -100,9 +100,10 @@ namespace nexus {
 
     if (costheta_min_ != -1. || costheta_max_ != 1. || phi_min_ != 0. || phi_max_ != 2.*pi) {
       G4ThreeVector p = nexus::Direction(costheta_min_, costheta_max_, phi_min_, phi_max_);
-      px = p.x() * pmod;
-      py = p.y() * pmod;
-      pz = p.z() * pmod;
+      p = p * pmod;
+      px = p.x();
+      py = p.y();
+      pz = p.z();
```
jacg: As I suggested in my PR:

https://github.com/next-exp/nexus/blob/6829fcfc397f405808f21e746339bd873290b9b1/source/generators/Back2backGammas.cc#L103-L119
Create the scaled vector in one go, and use the accessors at the point where you need the components.

If you don't like the ternary operator, default construct `momentum_direction` beforehand, and overwrite in an `if-then-else` that replaces the ` ? : `.

jacg: Given that we both seem to be at this right now, how about con-peras?

jacg: That code sample above contains too much stuff, so let's break it down. First: https://github.com/next-exp/nexus/blob/6829fcfc397f405808f21e746339bd873290b9b1/source/generators/Back2backGammas.cc#L107

Then: https://github.com/next-exp/nexus/blob/6829fcfc397f405808f21e746339bd873290b9b1/source/generators/Back2backGammas.cc#L113-L114

And also note that all the setting of particle properties (absent from my version) is pointless, because the particle *already has* those properties with those values!

source/generators/SingleParticle.cc
```diff
@@ -111,56 +108,51 @@ void SingleParticle::SetParticleDefinition(G4String particle_name)
...
-    pz = pmod * momentum_Z_/mom_mod;
-  } else if (costheta_min_ != -1. || costheta_max_ != 1. || phi_min_ != 0. || phi_max_ !=2.*pi) {
-    G4ThreeVector p = nexus::Direction(costheta_min_, costheta_max_, phi_min_, phi_max_);
-    px = p.x() * pmod;
-    py = p.y() * pmod;
-    pz = p.z() * pmod;
+
+  momentum_.setX(momentum_X_);
+  momentum_.setY(momentum_Y_);
+  momentum_.setZ(momentum_Z_);
```
jacg: This should not be here. You should 

+ *completely remove* the three members `momentum_X/Y/Z_` from the class, replacing them with a single `G4ThreeVector momentum_` member.

+ In the constructor, replace 

   ```c++
   msg_ -> DeclareProperty("...", momentum_X_ , "..."); # replace these with ...
   msg_ -> DeclareProperty("...", momentum_[0], "..."); # ... ones like these
   ```

source/generators/SingleParticle.cc
```diff
@@ -35,8 +35,8 @@ using namespace CLHEP;
 
 SingleParticle::SingleParticle():
 G4VPrimaryGenerator(), msg_(0), particle_definition_(0),
-energy_min_(0.), energy_max_(0.), geom_(0), momentum_X_(0.),
-momentum_Y_(0.), momentum_Z_(0.), costheta_min_(-1.), costheta_max_(1.),
+energy_min_(0.), energy_max_(0.), geom_(0),
+costheta_min_(-1.), costheta_max_(1.),
```
jacg: You should initialize `momentum_` here: Just add `momentum_{}`, or `momentum_{0,0,0}` if you want to be more explicit. Alternatively add `{}` or `{0,0,0}` at the point where you declare the `momentum_` member in the header.

paolafer: Done

source/generators/SingleParticle.cc
paolafer: Done.

source/generators/Back2backGammas.cc
paolafer: Done

source/generators/SingleParticle.cc
paolafer: Done.

source/generators/SingleParticle.cc
```diff
@@ -110,80 +107,48 @@ void SingleParticle::SetParticleDefinition(G4String particle_name)
...
+  } else {
+    p_dir = G4RandomDirection();
   }
 
+  G4ThreeVector p = pmod * p_dir;
+
   // Create the new primary particle and set it some properties
   G4PrimaryParticle* particle =
-    new G4PrimaryParticle(particle_definition_, px, py, pz);
+    new G4PrimaryParticle(particle_definition_, p.x(), p.y(), p.z());
```
jacg: Use `auto` to shorten this to a single line.

source/utils/RandomUtils.cc
```diff
@@ -0,0 +1,48 @@
...
+    }
+  }
+
+
+  G4ThreeVector Direction(G4double costheta_min, G4double costheta_max,
+                          G4double phi_min, G4double phi_max)
+  {
+    while (true) {
+      G4double cosTheta  = 2.*G4UniformRand()-1.;
+      if (cosTheta > costheta_min && cosTheta < costheta_max){
```
jacg: This loop-until-we-hit approach to restricting the range of random number is a bit perverse, and it's a waste of CPU cycles.

Is there any good reason why we can't just map `G4UniformRandom`'s `[0,1]` range to whatever `[mix, max]` range we need? Line 30 already maps it to `[-1,1]` so why not just do the correct mapping in the first place? As it stands, we generate random numbers in a range that is too large and keep trying until we hit our smaller target. If the target is much smaller than the range, it will, on average, take lots and lots of attempts.

Is there *any* point to doing it this bizarre way?

paolafer: I think that what you propose works. It would be replacing line 30 with 

```
G4double cosTheta  = (costheta_max - costheta_min) * G4UniformRand() + costheta_min;
```

And the same with `phi`. This way we can simply remove the while loop whatsoever, is that right?

source/generators/SingleParticle.cc
paolafer: Done.

jacg: I think that marking the conversation as resolved is probably enough, without having to add a "Done." comment: the arrival of the corresponding commit should be enough to attract attention.

source/utils/RandomUtils.cc
jacg: Yes, that's exactly right. (Removing *both* `while` loops.)

jacg: Hehe ... and when we're finished, of course, you should propagate (copy-paste) all we've learned here to the the other copy we have of this code on the `master` branch! (Sorry to bang on about it, but I must emphasize that the branch divergence is technical debt on which we're paying interest *all the time*.)

source/utils/RandomUtils.cc
```diff
@@ -26,22 +26,16 @@ namespace nexus {
...
-          G4double phi = CLHEP::twopi*G4UniformRand();
-          if (phi > phi_min && phi < phi_max){
-            return G4ThreeVector (sinTheta*std::cos(phi),
-                                  sinTheta*std::sin(phi),
-                                  cosTheta).unit();
-          }
-        }
-      }
-    }
+    G4double cosTheta  = (costheta_max - costheta_min) * G4UniformRand() + costheta_min;
```
jacg: Good.

However, now that we have to go through this same song-and-dance for `phi`, 5 lines lower down, we should say to ourselves: "Hmm, we've written the same pattern out by hand twice. Is it a useful pattern? Should we abstract it? Are we the first people in the history of programming to have come across the need for this? Is this the last time it will be needed in our project? Does `G4UniformRand` already cater for this option?"

Disappointingly, the authors of `G4UniformRand` appear not to have anticipated this need (I was hoping there were two optional parameters with default values 0 and 1).

I would argue that we should add `uniform_rand_in_range` to our collection of utilities, and hide the details in there.

source/generators/SingleParticle.cc
```diff
@@ -110,80 +107,47 @@ void SingleParticle::SetParticleDefinition(G4String particle_name)
...
-
-  // Create a new vertex
-  G4PrimaryVertex* vertex = new G4PrimaryVertex(position, time);
-
   // Generate uniform random energy in [E_min, E_max]
-  G4double kinetic_energy = RandomEnergy();
-
-  // Generate random direction by default
-  G4ThreeVector momentum_direction_ = G4RandomDirection();
+  G4double kinetic_energy = nexus::RandomEnergy(energy_max_, energy_min_);
```
jacg: There is already a `using namespace nexus;` near the top of this file, so the the `nexus::` isn't needed here, though you could leave it in, if you think that it adds some useful clarity.

source/generators/Back2backGammas.cc
```diff
@@ -27,116 +28,70 @@
...
+
+void Back2backGammas::GeneratePrimaryVertex(G4Event* evt)
+{
+  G4ParticleDefinition* gamma =
+    G4ParticleTable::GetParticleTable()->FindParticle("gamma");
+
+  auto p = 510.999 * keV * (
+                            (costheta_min_ != -1. || costheta_max_ != 1. || phi_min_ != 0. || phi_max_ != 2.*pi) ?
+                            Direction(costheta_min_, costheta_max_, phi_min_, phi_max_)                   :
+                            G4RandomDirection());
```
jacg: Ooops, the indentation got messed up here, presumably when you moved everything out of the `nexus` namespace.

Please,  detent the ternary operator and align `?` and `:`

source/utils/RandomUtils.cc
```diff
@@ -0,0 +1,42 @@
...
+// ----------------------------------------------------------------------------
+#include "RandomUtils.h"
+
+#include <Randomize.hh>
+
+#include "CLHEP/Units/SystemOfUnits.h"
+
+namespace nexus {
+
+  G4double UniformRandomInRange(G4double max_value,G4double min_value)
```
jacg: Please add the missing space in the parameter list. I think that similar minor problems like this remain in other places, but it's probably not reasonable to point them all out, here.

jacg (approved):
This PR contains some unrelated changes

1. Thorough cleaning up and simplification of `Back2backGammas::GeneratePrimaryVertex`
2. Addition of a couple of random data generation utilities:
   + `Direction`: generates uniformly distributed direction vectors subject to theta-phi angle restrictions
   + `UniformRandomInRange`: generates uniformly distributed `G4double`s within a specified range
3. Removal of `Back2backGammas` from `nexus` namespace.

I am not qualified to comment on no. 3.

1 & 2 have been reviewed con peras and I approve them, subject to one important code formatting fix (and some minor formatting niggles).

paolafer:
> 3. Removal of `Back2backGammas` from `nexus` namespace.
> 
> I am not qualified to comment on no. 3.

This commit doesn't belong to this PR, it slipped out of my fingers. The consistent use of the `nexus` namespace across the code will be reviewed in another PR.

jacg:
LGTM

PR #96, opened 2021-04-19
Add DEMOPP psf generator

gondiaz:
This PR adds a PSF generator, consisting of a vertex generator from random points inside a cilinder in the EL_GAP region. The PR also solves the bug in the exception raised when zmin>zmax.

The generator is tested by comparing the DEMOPP-Run8 psf generated from AD_HOC emission points with this new generator. The results shows a matching between PSFs, but effects can be seen at low distances probably because the AD_HOC method is affected by border effects.

Captura de pantalla 2021-04-19 a las 9 22 34

### DIFF
--- a/source/geometries/NextDemoFieldCage.cc
+++ b/source/geometries/NextDemoFieldCage.cc
@@ -85,7 +85,11 @@ namespace nexus {
     ELtransv_diff_ (1. * mm / sqrt(cm)),
     ELlong_diff_ (0.5 * mm / sqrt(cm)),
     elfield_(0),
-    ELelectric_field_ (23.2857 * kilovolt / cm)
+    ELelectric_field_ (23.2857 * kilovolt / cm),
+    // EL gap generation disk parameters
+    el_gap_gen_disk_diam_(0.),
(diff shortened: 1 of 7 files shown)


source/geometries/NextDemoFieldCage.cc
```diff
@@ -85,7 +85,11 @@ namespace nexus {
     ELtransv_diff_ (1. * mm / sqrt(cm)),
     ELlong_diff_ (0.5 * mm / sqrt(cm)),
     elfield_(0),
-    ELelectric_field_ (23.2857 * kilovolt / cm)
+    ELelectric_field_ (23.2857 * kilovolt / cm),
+    // EL gap generation disk parameters
+    el_gap_gen_disk_diam_(0.),
+    el_gap_gen_disk_x_(0.), el_gap_gen_disk_y_(0.),
+    el_gap_gen_disk_zmin_(0.), el_gap_gen_disk_zmax_(1.)
   {
```
jmunozv, line 93: I think it is clearer if we only initialize one variable per line

jmunozv (approved):
This PR does what it claims, so I approve it.

PR #97, opened 2021-04-19
Change nexus interactive mode

gondiaz:
This change allows a more user-friendly interactive nexus visualization. In particular, it enables a Qt-based GUI (see Figure). The GUI is composed by a 3D mouse-interactive OGL viewer (up panel), a listing with the possible input commands (left panel) and a terminal where commands are launched (down panel).
To activate this GUI, Geant4 must be compiler with Qt (see G4 installation guide). The Figure shows an example of 100 photons generated in the EL_GAP for DEMOPP using the generator implemented in PR #96.

Captura de pantalla 2021-04-19 a las 9 43 24

### DIFF
--- a/source/nexus.cc
+++ b/source/nexus.cc
@@ -109,7 +109,7 @@ G4int main(int argc, char** argv)
 
   // visual mode
   if (!batch) {
-    std::unique_ptr<G4UIExecutive> ui{new G4UIExecutive{argc, argv}};
+    std::unique_ptr<G4UIExecutive> ui{new G4UIExecutive{1, argv}};
     std::unique_ptr<G4VisManager> visManager{new G4VisExecutive};
     visManager->Initialize();
     UI->ApplyCommand("/control/execute macros/vis.mac");
--- a/source/nexus.cc
+++ b/source/nexus.cc
@@ -8,17 +8,13 @@
 
 #include "NexusApp.h"
 
-#include <G4UImanager.hh>   
-#include <G4UIterminal.hh>  
-#include <G4UItcsh.hh> 
+#include <G4UImanager.hh>
+#include <G4UIExecutive.hh>
 #include <G4VisExecutive.hh>
 
 #include <getopt.h>
 
 using namespace nexus;
-
-
-
 
 
 void PrintUsage()
@@ -33,15 +29,14 @@ void PrintUsage()
 }
 
 
-
-G4int main(int argc, char** argv) 
+G4int main(int argc, char** argv)
 {
   ////////////////////////////////////////////////////////////////////
   // PARSE COMMAND-LINE OPTIONS
 
   // Abort if no command-line argument is provided. The user must
   // specify at least the name of a configuration macro.
-  if (argc < 2) PrintUsage(); 
+  if (argc < 2) PrintUsage();
 
   G4bool batch = true;
   G4int nevents = 0;
@@ -61,7 +56,7 @@ G4int main(int argc, char** argv)
     //  int option_index = 0;
     opterr = 0;
     c = getopt_long(argc, argv, "bin:", long_options, 0);
-    
+
     if (c==-1) break; // Exit if we are done reading options
 
     switch (c) {
@@ -89,8 +84,8 @@ G4int main(int argc, char** argv)
   // If there is no other command-line argument to be processed, abort
   // because the user has not provided a configuration macro.
   // (The variable optind is set by getopt_long to the index of the next
-  // element of the argv array to be processed. Once getopt has found all 
-  // of the option arguments, this variable can be used to determine where 
+  // element of the argv array to be processed. Once getopt has found all
+  // of the option arguments, this variable can be used to determine where
   // the remaining non-option arguments begin.)
   if (optind == argc) PrintUsage();
 
@@ -98,13 +93,10 @@ G4int main(int argc, char** argv)
   // command-line parameters that is not a GNU option.
   G4String macro_filename = argv[optind];
 
-  if (macro_filename == "") PrintUsage(); 
-    
-
+  if (macro_filename == "") PrintUsage();
 
   ////////////////////////////////////////////////////////////////////
 
-  
   NexusApp* app = new NexusApp(macro_filename);
   app->Initialize();
 
@@ -112,20 +104,18 @@ G4int main(int argc, char** argv)
 
   // if (seed < 0) CLHEP::HepRandom::setTheSeed(time(0));
   // else CLHEP::HepRandom::setTheSeed(seed);
-  
+
   // CLHEP::HepRandom::showEngineStatus();
 
   // visual mode
   if (!batch) {
-    G4VisManager* vismgr = new G4VisExecutive();
-    vismgr->Initialize();
-
-    G4UIsession* session = new G4UIterminal(new G4UItcsh);
+    G4UIExecutive* ui = new G4UIExecutive(argc, argv);
+    G4VisManager* visManager = new G4VisExecutive;
+    visManager->Initialize();
     UI->ApplyCommand("/control/execute macros/vis.mac");
-    session->SessionStart();
-
-    delete session;
-    delete vismgr;
+    ui->SessionStart();
+    delete ui;
+    delete visManager;
   }
   else {
     app->BeamOn(nevents);
@@ -133,4 +123,4 @@ G4int main(int argc, char** argv)
 
   delete app;
   return EXIT_SUCCESS;
-} 
+}
--- a/source/nexus.cc
+++ b/source/nexus.cc
@@ -109,13 +109,11 @@ G4int main(int argc, char** argv)
 
   // visual mode
   if (!batch) {
-    G4UIExecutive* ui = new G4UIExecutive(argc, argv);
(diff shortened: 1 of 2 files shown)


source/nexus.cc
```diff
@@ -89,48 +84,43 @@ G4int main(int argc, char** argv)
...
+    G4VisManager* visManager = new G4VisExecutive;
+    visManager->Initialize();
     UI->ApplyCommand("/control/execute macros/vis.mac");
-    session->SessionStart();
-
-    delete session;
-    delete vismgr;
+    ui->SessionStart();
+    delete ui;
+    delete visManager;
```
jacg: Given that you're touching this code anyway, and that it's of such limited scope that it's impossible not to understand that the change is 100% equivalent ... why not take this opportunity to enter the world of *Modern C++ Programming* (est. 2011) and manage these lifetimes with RAII, specifically `std::unique_ptr`!

# TLDR

``` c++
{
  Banana* my_banana = new Banana(colour, shape, taste);
  // do stuff with my_banana
  delete my_banana;
}
```
should be written as
``` c++
{
  std::unique_ptr<Banana> my_banana = std::make_unique<Banana>(colour, shape, taste);
  // do stuff with my_banana
} // deleted automatically when unique_ptr goes out of scope
```
or rather
``` c++
{
  auto my_banana = std::make_unique<Banana>(colour, shape, taste);
  // do stuff with my_banana
} // deleted automatically when unique_ptr goes out of scope
```

Caveat: The C++ standards committee forgot to include `make_unique` in C++11, and I believe that Nexus still uses C++11 (that's, like, the middle ages, guys!) So you'll have to write that first line as one of these:

``` c++
auto my_banana = std::unique_ptr<Banana>{new my_banana{colour, shape, taste}}; // version 1
std::unique_ptr<Banana> my_banana{new my_banana{colour, shape, taste}};        // version 2
```

Yes, I used braces! Parentheses will work too, in this case, but that's very passé. c.f. [*The Tale of Joe HEP Programmer*](https://github.com/next-exp/nexus/pull/93#issuecomment-809274190).

Lots of goodness follows from using smart pointers, I have proof of this, but there is not enough space in this margin.

# Background

Since C++11 you don't want to see `new` and `delete` *anywhere* in your code, unless you are writing some very low-level utility such as a container, smart pointer or allocator ... which this emphatically is *not*.

Caveats:

+ It's not possible to make `new` disappear completely, but the modern C++ programmer doesn't want to see it unless it is in the argument list of some lifetime-managing consumer.

+ Geant4 makes it *really* hard to make `new`s disappear: it insists you drop all sorts of `new`ed objects on the floor because it takes control of them *implicitly*. Ugh!

+ The C++ committee itself didn't help, by forgetting to include `std::make_unique` until C++14.

gondiaz: Thanks for the explanation. But it seems that `unique_ptr` is not valid in this case or I could be missing something...

```
source/nexus.cc:112:36: error: no matching constructor for initialization of 'std::unique_ptr<G4UIExecutive>'
    std::unique_ptr<G4UIExecutive> ui{argc, argv};
                                   ^ ~~~~~~~~~~~~
/Library/Developer/CommandLineTools/usr/bin/../include/c++/v1/memory:2500:3: note: candidate constructor template not
      viable: no known conversion from 'int' to 'std::__1::unique_ptr<G4UIExecutive,
      std::__1::default_delete<G4UIExecutive> >::pointer' (aka 'G4UIExecutive *') for 1st argument
  unique_ptr(pointer __p, _LValRefType<_Dummy> __d) _NOEXCEPT
```

jacg: This doesn't make sense:
``` c++
std::unique_ptr<G4UIExecutive> ui{argc, argv};
```
In C++ 11 you should write
``` c++
std::unique_ptr<G4UIExecutive> ui{new G4UIExecutive(argc, argv)};
```
whereas in C++ 14 you should prefer
``` c++
auto ui = std::make_unique<G4UIExecutive>(argc, argv);
```

(And adding `using std::make_unique` and/or `using std::unique_ptr` somewhere near the top of the file, is probably a good idea, especially if it is not a header file.)

The first version says (or would, if it were meaningful) something like

+ `ui` should be a unique pointer to `G4UIExecutive`

+ It should take charge of the lifetime of the two `G4UIExecutive`s pointed to by `argc` and `argv`


The C++ 11 version says

+ `ui` should be a unique pointer to `G4UIExecutive`

+ It should take charge of the lifetime of the `G4UIExecutive` pointed to by `new G4UIExecutive(argc, argv)`


The C++ 14 version says

+ Construct a G4UIExecutive with the construction arguments `argc` and `argv`, and manage its lifetime with a `unique_ptr`.

+ Store this unique pointer in the variable `ui` and figure out what its type is yourself.

jmalbos:
> Caveat: The C++ standards committee forgot to include make_unique in C++11, and I believe that Nexus still uses C++11 (that's, like, the middle ages, guys!).

This is mostly an accident. There is no good reason to stay in C++11 besides lack of time to test that everything works with a more modern version of the standard. We'll try to address this in coming weeks.

source/nexus.cc
gondiaz: Ok, I was just following your previous explanation
`std::unique_ptr<Banana> my_banana{colour, shape, taste};`
must be
`std::unique_ptr<Banana> my_banana{new Banana(colour, shape, taste)};`
or maybe, from Joe's tale
`std::unique_ptr<Banana> my_banana{new Banana{colour, shape, taste}};`

jacg: Hmm, I'm not sure whether you're pointing out an error in my original explanation (if so, I haven't found it yet) or whether you're drawing parallels to my original explanation ... hmm, maybe that's it ... in any case: you are currently stuck in C++11, so let's forget that `std::make_unique` exists (because for you, it won't exist for another 3 years: until 2014).

Let's start with (already using braces for construction wherever possible)

``` c++
G4UIExecutive* ui = new G4UIExecutive{argc, argv};
```

which means

> Create a variable `ui` of type pointer-to-`G4UIExecutive` and initialize it to `new G4UIExecutive{argc, argv}`, and make managing its lifetime be my problem.

First, let's get rid of the annoying and pointless repetition of `G4UIExecutive`; `auto` is your friend:

``` c++
auto ui = new G4UIExecutive{argc, argv};
```

Now we're saying

> Create a variable `ui` and figure out what its type must be, given that we're going to initialize it to `new G4UIExecutive{argc, argv}`, and make managing its lifetime be my problem.

We want to change this to 

> Create a variable `ui` and figure out what its type must be, given that we're going to initialize it to `new G4UIExecutive{argc, argv}`, and let `unique_ptr` manage its lifetime.

The 'let `unique_ptr` manage its lifetime' part is achieved by using it as a construction argument to `unique_ptr`. So we should end up with

``` c++
auto ui = std::unique_ptr<G4UIExecutive>{new std::G4UIExecutive{argc, argv}};
```

Replacing *copy-initialization* with *direct-initialization* we can also write this as

``` c++
std::unique_ptr<G4UIExecutive> ui{new std::G4UIExecutive{argc, argv}};
```

And comparing it to what you wrote in the code that caused the error

``` c++
std::unique_ptr<G4UIExecutive> ui{                       argc, argv };
```

we see that the latter means (very roughly) something like `T* = a, b;` while `T* = new T(a,b)` was needed.

Make sense? Any use to you?

(The nice thing about `make_unique` is that it's a function template, and therefore it can deduce its template argument from its function argument, which spares us the need of repeating `G4UIExecutive`.)

source/nexus.cc
```diff
@@ -109,13 +109,11 @@ G4int main(int argc, char** argv)
 
   // visual mode
   if (!batch) {
-    G4UIExecutive* ui = new G4UIExecutive(argc, argv);
-    G4VisManager* visManager = new G4VisExecutive;
+    std::unique_ptr<G4UIExecutive> ui{new G4UIExecutive{argc, argv}};
+    std::unique_ptr<G4VisManager> visManager{new G4VisExecutive};
```
jacg: Spot on!

source/nexus.cc
gondiaz: Yes, I think it is clear. I was pointing to this, from your first explanation:

> Caveat: The C++ standards committee forgot to include make_unique in C++11, and I believe that Nexus still uses C++11 (that's, like, the middle ages, guys!) So you'll have to write that first line as:
> 
> `std::unique_ptr<Banana> my_banana{colour, shape, taste};`

jacg: Hmm ... dunno what that's doing there: it's complete nonsense! Sorry. I'll edit it, lest someone else be misled.

source/nexus.cc
jmalbos: We are not using the command-line parameters for whatever use `G4UIExecutive` gives them. Maybe we should pass a 1 as `argc`, as if the user had not provided any argument.

jmalbos:
This PR updates the main function of nexus so that it makes use of `UIExecutive`. This, in turn, enables the use of all the different visualisation drivers (including, for example, Qt) available in the local Geant4 distribution. I approve it.

paolafer (approved):
The code has been tested on several platforms. It gives some problems in certain operative systems, but, since the old visualization keeps working, we approve this PR, while we debug the individual problems.

PR #99, opened 2021-04-21
Demo run9 masks

gondiaz:
Implement rectangular mask holes in demo geometry for Run 9.

### DIFF
--- a/source/geometries/NextDemoSiPMBoard.cc
+++ b/source/geometries/NextDemoSiPMBoard.cc
@@ -87,7 +87,7 @@ void NextDemoSiPMBoard::Construct()
 
   G4double sipm_z_dim = 0.;
   G4RotationMatrix* sipm_rot = new G4RotationMatrix();
-  
+
   if (sipm_type_ == "sensl"){
     SiPMSensl* sipm = new SiPMSensl();
 
@@ -118,9 +118,6 @@ void NextDemoSiPMBoard::Construct()
     sipm_z_dim = sipm->GetDimensions().z();
     sipm_rot->rotateY(0.0);
   }
-  else
-    G4Exception("[NextDemoSiPMBoard]", "Construct()",
-                FatalException, "Unknown sipm type.");
 
   /// Make sure the mother physical volume is actually valid
   if (!mother_phys_)
--- a/source/geometries/NextDemoSiPMBoard.cc
+++ b/source/geometries/NextDemoSiPMBoard.cc
@@ -15,6 +15,7 @@
 
 #include "MaterialsList.h"
 #include "SiPMSensl.h"
+#include "Next100SiPM.h"
 #include "OpticalMaterialProperties.h"
 #include "BoxPointSampler.h"
 #include "Visibilities.h"
@@ -48,10 +49,11 @@ NextDemoSiPMBoard::NextDemoSiPMBoard():
   mask_thickn_     (2.0  * mm),
   membrane_thickn_ (0.),
   coating_thickn_  (0.),
-  hole_type_       ("rounded"),
+  hole_type_       (""),
   hole_diam_       (3.5  * mm),
   hole_x_          (0.0  * mm),
   hole_y_          (0.0  * mm),
+  sipm_type_       (""),
   mother_phys_     (nullptr),
   kapton_gen_      (nullptr)
 {
@@ -67,7 +69,7 @@ NextDemoSiPMBoard::NextDemoSiPMBoard():
   msg_->DeclareProperty("sipm_board_vis", visibility_,
                         "NextDemoSiPMBoard visibility.");
 
-  sipm_ = new SiPMSensl;
+  // sipm_ = new SiPMSensl;
 }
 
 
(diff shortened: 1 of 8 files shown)


source/geometries/NextDemoTrackingPlane.cc
```diff
@@ -86,31 +86,52 @@ void NextDemoTrackingPlane::Construct()
...
+    if(verbosity_) G4cout << "run9 ..." << G4endl;
+    gate_board_dist = 19.66 * mm;
+    mask_thickn     = 6.0 * mm;
+    membrane_thickn = 0.2 * mm;
+    coating_thickn  = 2.0 * micrometer;
+    hole_type       = "rectangular";
+    hole_x          = 6.0 * mm;
+    hole_y          = 5.0 * mm;
+    sipm_type       = "next100";
   }
```
paolafer, line 130: I would add a last `else` which raises an exception if `config_` takes none of the above values.

paolafer: In fact, I would move the exception at line 81 here.

paolafer: Looking at the rest of the code, I see that the check that the configuration is a legal one is done in `NextDemoInnerElements`. In that case, I would say that it is superfluous to add any exception here, so I would remove that at line 81, too. The same should be done in `NextDemoFieldCage`. Do you agree?

source/geometries/NextDemoSiPMBoard.cc
```diff
@@ -48,23 +51,26 @@ NextDemoSiPMBoard::NextDemoSiPMBoard():
...
 {
   msg_ = new G4GenericMessenger(this, "/Geometry/NextDemo/",
                                 "Control commands of the NextDemo geometry.");
 
-  msg_->DeclareProperty("sipm_board_verbosity", verbosity_,
-                        "NextDemoSiPMBoard verbosity");
+  msg_->DeclareProperty("sipm_board_verbosity",       verbosity_, "NextDemoSiPMBoard verbosity");
+  msg_->DeclareProperty("sipm_verbosity"      ,  sipm_verbosity_, "NextDemo SiPMs verbosity");
+  msg_->DeclareProperty("sipm_board_vis"      ,      visibility_, "NextDemoSiPMBoard visibility.");
+  msg_->DeclareProperty("sipm_visibility"     , sipm_visibility_, "SiPMs Visibility");
```
paolafer: For uniformity, I would use the same type of string as the other three parameters, namely: "NextDemo SiPMs visibility".

source/geometries/NextDemoSiPMBoard.cc
```diff
@@ -79,6 +85,40 @@ NextDemoSiPMBoard::~NextDemoSiPMBoard()
...
+    sipm->SetMotherDepth(4);
+    sipm->SetNamingOrder(1000);
+    sipm->Construct();
+
+    sipm_ = sipm;
+
+    sipm_z_dim = sipm->GetDimensions().z();
+  }
+  else
+    G4Exception("[NextDemoSiPMBoard]", "Construct()",
```
paolafer: Same comment as before: I'm not sure we need this exception, since the value of `sipm_type_` is set automatically when the configuration parameter for the run is chosen.

source/geometries/NextDemoSiPMBoard.cc
```diff
@@ -167,31 +207,29 @@ void NextDemoSiPMBoard::Construct()
...
-  sipm_->SetNamingOrder(1000);
-  sipm_->Construct();
-
   // Generate SiPM positions
   GenerateSiPMPositions();
 
   // SiPM placement inside the hole
   G4RotationMatrix* sipm_rot = new G4RotationMatrix();
-  sipm_rot->rotateY(pi);
+  if (sipm_type_ == "sensl")
```
paolafer: I would put this rotation above, where the other properties of the SiPM are set, so we can use the same `if` condition.

source/geometries/NextDemoTrackingPlane.cc
gondiaz: Agree. We are redundantly doing the same check in three different places now.

source/geometries/NextDemoSiPMBoard.cc
gondiaz: I added this exception together with the hole equivalent in
https://github.com/gondiaz/nexus/blob/8e8a377656b9fb50d0fc14c0116cefeeb9b8a5a8/source/geometries/NextDemoTrackingPlane.cc#L158

Maybe is useful if someone is about to change the parameters for some run (which is not a good idea by the way) or to add another one with different parameters. Both checks are only made once.

paolafer: As I see it, we want to have the configurations frozen, with `runN` having all its correct parameters. Checking that the sipm type is one of those doesn't really ensure that it is the correct type, just that is one of the possible types. Therefore, I'm not sure this check is useful, although is not harmful, either.

gondiaz: Agree, to me sounds better to keep those parameters constant and correct for each specific run configuration. I remove the exceptions.

paolafer:
I think that testing all the configurations of DEMO is good, even if it adds time for the tests to run. One thing you could do is to try to use a lower electric field and scintillation yield to see how much we gain in time, while having a high enough sensor response.

paolafer (approved):
This PR adds a new configuration to the DEMO++ with different masks and SiPMs. It adapts the code to use different shapes of masks and sensors. Approved!

PR #100, opened 2021-04-22
Remove noise when running automatic tests

paolafer:
This PR sets to false the parameters that control verbosity in NextFlex, so that less noise is produced when automatic tests run. The parameters are kept in the macro to show their usage.


jmunozv (approved):
Approved

PR #101, opened 2021-04-26
Create nexus library in compilation

paolafer:
This PR modifies the building script to create a nexus library and to copy the headers files to a separate directory.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -321,7 +321,7 @@ if env['PREFIX'] == DEFAULT_PATH:
 
 WriteNexusConfig(w_prefix_dir)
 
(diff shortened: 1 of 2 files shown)


SConstruct
```diff
@@ -74,6 +77,53 @@ def AssertG4Version(path):
             Abort(msg)
 
 
+## write nexus-config file
+def WriteNexusConfig(dir_prefix):
```
jmalbos, line 81: I was trying to find out if SCons provides any standard tool to generate such configuration scripts, but the documentation doesn't mention it at all.

jmalbos (approved):
This PR modifies the `SConstruct` building script so that it generates a library and a `pkg-config`-like script for **nexus**. Approved!

PR #102, opened 2021-04-27
Update RandomUtils and SingleParticleGenerator to the latest changes in the petalo branch

paolafer:
This PR adds to the nexus core the latest changes carried out in the `petalo` branch regarding the random generation of momentum and energy of an individual particle. The code itself has already passed the review, it only needs to be checked that the code compiles and runs (in the case mistakes had been introduced in the copy), which is done automatically.

### DIFF
--- a/source/generators/SingleParticleGenerator.cc
+++ b/source/generators/SingleParticleGenerator.cc
@@ -35,9 +35,8 @@ using namespace CLHEP;
 
 SingleParticleGenerator::SingleParticleGenerator():
 G4VPrimaryGenerator(), msg_(0), particle_definition_(0),
-energy_min_(0.), energy_max_(0.), geom_(0), momentum_X_(0.),
-momentum_Y_(0.), momentum_Z_(0.), costheta_min_(-1.),
-costheta_max_(1.), phi_min_(0.), phi_max_(2.*pi)
+energy_min_(0.), energy_max_(0.), geom_(0), momentum_{},
+costheta_min_(-1.), costheta_max_(1.), phi_min_(0.), phi_max_(2.*pi)
 {
   msg_ = new G4GenericMessenger(this, "/Generator/SingleParticle/",
     "Control commands of single-particle generator.");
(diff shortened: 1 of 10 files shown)


source/generators/SingleParticleGenerator.cc
```diff
@@ -62,11 +61,11 @@ costheta_max_(1.), phi_min_(0.), phi_max_(2.*pi)
   msg_->DeclareProperty("region", region_,
     "Set the region of the geometry where the vertex will be generated.");
 
-  msg_->DeclareProperty("momentum_X", momentum_X_,
+  msg_->DeclareProperty("momentum_X", momentum_[0],
```
jmalbos: We could simplify this using the function `DeclarePropertyWithUnit`, which accepts 3-vectors (see https://geant4.kek.jp/lxr/source/intercoms/src/G4GenericMessenger.cc#L106).

source/utils/RandomUtils.cc
```diff
@@ -0,0 +1,42 @@
...
+  {
+    if (max_value == min_value){
+      return min_value;
+    } else {
+      return (G4UniformRand()*(max_value - min_value) + min_value);
+    }
+  }
+
+
+  G4ThreeVector Direction(G4double costheta_min, G4double costheta_max,
```
jmalbos: `Direction` is not a very descriptive name. Why not `UniformRandomDirection` or something similar?

paolafer: What about `RandomDirection`?

paolafer: Actually, it should be `RandomDirectionInRange`.

jmalbos: Yes, that's fine too.

jmalbos:
This PR updates the `SingleParticleGenerator` cleaning up the code and making use of new random-generation functions (added to `utils`). The new code compiles and runs. I approve it.

PR #104, opened 2021-04-30
FactoryBase

jmalbos:
First version of the new `FactoryBase`


paolafer:
The code of the example works and is a good solution to the factory problem. I can complete it pushing to your branch, if that's ok, @jmalbos.

jmalbos:
>  I can complete it pushing to your branch, if that's ok, @jmalbos.

Yes, of course, go ahead.

jmalbos:
This PR gets rid of the old if-else-based factories and replaces them with a mechanism similar to that of the `G4GenericMessenger`. I approve it.

PR #105, opened 2021-05-06
Add  new class describing the geometry of the NEXT-100 SiPMs

jmalbos:
This pull request adds the geometry of the Hamamatsu silicon photomultipliers (MPPC S13372-1350TE) selected for the NEXT-100 detector.

Data sheet: [K30-I50426_specification_sheet_S13372-1350TE_.pdf](https://github.com/next-exp/nexus/files/6432978/K30-I50426_specification_sheet_S13372-1350TE_.pdf)


halmamol:
What about changing the name of the geometry to 'SiPMHamamatsu' (in the same manner we have 'SiPMSensl')?

jmalbos:
> What about changing the name of the geometry to 'SiPMHamamatsu' (in the same manner we have 'SiPMSensl')?

I don't like it. Hamamatsu has many SiPMs in its catalog, and this aren't the only ones we've used in NEXT from them. I'd either keep it this way or use the actual sensor name.

(I don't like the name 'SiPMSensl' either.)

macros/SIPM_QE_setup.config.mac
```diff
@@ -0,0 +1,25 @@
+## ----------------------------------------------------------------------------
+## nexus | pmtQE.mac
```
paolafer: Please, change the description from PMT to SiPM.

macros/SIPM_QE_setup.init.mac
```diff
@@ -0,0 +1,22 @@
...
+## nexus | SIPM_QE_setup.init.mac
+##
+## Initialization macro to simulate photons and measure them
+## in a SIPM
+##
+## The NEXT Collaboration
+## ----------------------------------------------------------------------------
+
+/PhysicsList/RegisterPhysics G4EmStandardPhysics_option4
+/PhysicsList/RegisterPhysics NexusPhysics
```
paolafer: In this case it doesn't matter, but in general NexusPhysics must be declared *after* the other G4 physics lists, because it uses processes or particles defined elsewhere. Therefore, I would change it here, too, in case somebody takes this macro as an example for other kinds of simulations.

source/geometries/Next100SiPM.cc
```diff
@@ -0,0 +1,191 @@
...
+
+  if (!sdmgr->FindSensitiveDetector(sdname, false)) {
+    PmtSD* sensdet = new PmtSD(sdname);
+
+    if (sensor_depth_ == -1)
+      G4Exception("[Next100SiPM]", "Construct()", FatalException,
+                  "Sensor Depth must be set before constructing");
+    sensdet->SetDetectorVolumeDepth(sensor_depth_);
+
+    if (mother_depth_ == -1)
```
paolafer: This `if` clause and the following one is not necessary, because the default value is `0`, not `-1`. We set them to `0` because one could use this SiPM directly inside the gas or a support, and not in a dice board, which is replicated. `mother_depth` and `naming_order` are only needed in the latter kind of situation, otherwise they can be left at their default value (0).

paolafer (changes requested):

jmalbos:
Do we want to include the files of the QE setup in this PR?

source/geometries/SIPM_QE_setup.cc
```diff
@@ -0,0 +1,126 @@
...
+    // G4Material* gxe = MaterialsList::GXe(gxe_pressure);
+    // gxe->SetMaterialPropertiesTable(OpticalMaterialProperties::GXe(gxe_pressure));
+    // G4LogicalVolume* gas_logic = new G4LogicalVolume(gas_solid, gxe, "GAS");
+
+    // air
+    G4Material* air = G4NistManager::Instance()->FindOrBuildMaterial("G4_AIR");
+    air->SetMaterialPropertiesTable(OpticalMaterialProperties::Vacuum());
+    G4LogicalVolume* gas_logic = new G4LogicalVolume(gas_solid, air, "GAS");
+    new G4PVPlacement(0, G4ThreeVector(0,0,0), gas_logic, "GAS",
+		      chamber_logic, false, 0, true);
```
paolafer: Please, set the variable that checks volume overlap to `false`. We want to avoid the check and the output every time we run a geometry.

source/geometries/SIPM_QE_setup.cc
```diff
@@ -0,0 +1,126 @@
...
+    // SIPM /////////////////////////////////////////
+    sipm_.SetSensorDepth(2);
+    sipm_.Construct();
+
+    G4LogicalVolume* sipm_logic = sipm_.GetLogicalVolume();
+    sipm_length_ = 1.3*mm;
+
+    new G4PVPlacement(0, G4ThreeVector(0.,0.,-length_/2.+sipm_length_/2.),
+		      sipm_logic, "SIPM",
+		      gas_logic, false, 0, true);
```
paolafer: Same as above.

paolafer:
> Do we want to include the files of the QE setup in this PR?

I don't know. I don't even know if we need to have a geometry for the PMT class and another one for the SiPM class. It looks like duplication of code.

paolafer:
We can give it a thought about which setup geometry we want to keep, or how to generalize it parametrizing the actual sensor. In this case, I would decouple it from this PR, and just leave the Next100SiPM geometry.

source/geometries/Next100SiPM.cc
halmamol: You're right. I started setting `mother_depth` and `naming_order` to `-1`, but later on I set them to `0`. I'll remove them.

paolafer:
One more thing: the value of the `visibility_` variable cannot be changed right now. If you want to leave to the user the possibility of making the class transparent or visible, you should add another setter method.

halmamol:
> One more thing: the value of the `visibility_` variable cannot be changed right now. If you want to leave to the user the possibility of making the class transparent or visible, you should add another setter method.

You're right. I've added it on the header file.

paolafer (approved):
This PR adds the geometry of the Hamamatsu SiPMs that are going to be used in  NEXT-100. The class has all the properties required for a SiPMs. Approved.

PR #107, opened 2021-05-20
Exclude optical photons and ie- from tracking to save memory.

paolafer:
This PR reverts a change of one of the last commits, starting from which the trajectories of all the particles in the simulation (including optical photons and ionization electrons) were saved, causing a huge consumption of memory.


jmalbos:
Simple PR fixing an accidental recent change in `DefaultTrackingAction`. I approve it.

jmalbos (approved):

PR #109, opened 2021-05-21
Remove variable which caused compilation warning.

paolafer:
Ditto.


gondiaz (approved):
This PR removes an unused declared variable, which is highlighted as a warning by the compiler. It does what it claims, I approve it.

PR #110, opened 2021-05-22
New 3d vectors

paolafer:
This PR adapts the code to using a 3-D vector instead of three different coordinates, when a specific vertex or momentum is passed by the user via configuration parameter.

It also changes the way the `AD_HOC` vertex is implemented in the `NEW` geometry, to make it similar to the rest of the geometries.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -304,7 +304,6 @@ for d in SRCDIR:
 env['CXXCOMSTR']  = "Compiling $SOURCE"
 env['LINKCOMSTR'] = "Linking $TARGET"
 
-
 LIBDIR = env['PREFIX'] + '/lib/nexus'
 libnexus = env.SharedLibrary(LIBDIR, src)
 
@@ -325,7 +324,7 @@ env.Execute(Chmod(w_prefix_dir+'/bin/nexus-config', 0o755))
 nexus = env.Program('bin/nexus', ['source/nexus.cc']+src)
 
 TSTDIR = ['utils',
-	  'example']
+	      'example']
 TSTDIR = ['source/tests/' + dir for dir in TSTDIR]
 
 tst = []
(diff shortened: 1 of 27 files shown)


macros/NEXT_options.config.mac
```diff
@@ -11,7 +11,7 @@
 ##### VERBOSITIES #####
 /run/verbose 0
 /event/verbose 0
-/tracking/verbose 0
+/tracking/verbose 1
```
jmalbos: Is this change on purpose?

macros/NEW_geantino.config.mac
```diff
@@ -15,9 +15,7 @@
 /Generator/SingleParticle/min_energy 1  eV
 /Generator/SingleParticle/max_energy 1 keV
 /Generator/SingleParticle/region AD_HOC
-/Geometry/NextNew/specific_vertex_X   0 mm
-/Geometry/NextNew/specific_vertex_Y   0 mm
-/Geometry/NextNew/specific_vertex_Z 100 mm
+/Geometry/NextNew/specific_vertex   0 0 100 mm
```
jmalbos: Remove the extra whitespaces between `specific_vertex` and the first `0`.

SConstruct
```diff
@@ -325,7 +324,7 @@ env.Execute(Chmod(w_prefix_dir+'/bin/nexus-config', 0o755))
 nexus = env.Program('bin/nexus', ['source/nexus.cc']+src)
 
 TSTDIR = ['utils',
-	  'example']
+	      'example']
```
jmalbos: The alignment has been changed.

paolafer: I'm not sure how to fix it, because in my editor it appears as aligned.

macros/NEXT_options.config.mac
paolafer: No, fixed.

SConstruct
paolafer: I fixed it using another editor.

jmalbos (approved):
This PR introduces the use of 3-vectors (plus unit) in the configuration macros. The code compiles and runs as intended. I approve it.

PR #111, opened 2021-05-24
Next100 shielding revision

gondiaz:
This PR revisits and updates the shielding geometry and generators. The dimensions and components were gently provided by Jordi Torrent and are shown in

[Shielding.pdf](https://github.com/next-exp/nexus/files/6554546/101-Castle.and.new.dimensions.pdf)
[pedastal.pdf](https://github.com/next-exp/nexus/files/6602464/pedastal.pdf)

Besides the relevant change of the new dimensions, two extra changes are made in this revision:
- the steel shielding in the laterals is enlarged, as shown in the drawings. The steel roof is also positioned above the inner steel box.
- solves a bug in the distribution of the vertex generator in the steel shielding structure composed by the roof and the top beams.
I have also tryed to rewrite the code to be more easily readable.

### DIFF
--- a/source/geometries/Next100Shielding.cc
+++ b/source/geometries/Next100Shielding.cc
@@ -185,7 +185,7 @@ namespace nexus {
     G4Box* air_box_solid = new G4Box("INNER_AIR", shield_x_/2., shield_y_/2., shield_z_/2.);
 
     air_box_logic_ = new G4LogicalVolume(air_box_solid,
-                                			   G4NistManager::Instance()->FindOrBuildMaterial("G4_AIR"),
+                                         G4NistManager::Instance()->FindOrBuildMaterial("G4_AIR"),
                                 			   "INNER_AIR");
 
     ////Limit the uStepMax=Maximum step length, uTrakMax=Maximum total track length,
@@ -222,36 +222,37 @@ namespace nexus {
     // Creating the vertex generators   //////////
     //lead_gen_  = new BoxPointSampler(steel_x, steel_y, steel_z, lead_thickness_, G4ThreeVector(0.,0.,0.), 0);
     // Only shooting from the innest 5 cm.
-    lead_gen_  = new BoxPointSampler(steel_x, steel_y, steel_z, 5.*cm, G4ThreeVector(0.,0.,0.), 0);
+    lead_gen_  = new BoxPointSampler(steel_x, steel_y, steel_z, 5.*cm,
+                                     G4ThreeVector(0.,0.,0.), 0);
 
     G4double ext_offset = 1. * cm;
-    external_gen_ = new BoxPointSampler(lead_x_ + ext_offset, lead_y_ + ext_offset, lead_z_ + ext_offset,
-                                        1. * mm, G4ThreeVector(0.,0.,0.), 0);
-
+    external_gen_ = new BoxPointSampler(lead_x_ + ext_offset, lead_y_ + ext_offset, lead_z_ + ext_offset, 1. * mm,
+                                        G4ThreeVector(0.,0.,0.), 0);
 
     steel_gen_ = new BoxPointSampler(shield_x_, shield_y_, shield_z_, steel_thickness_,
                                      G4ThreeVector(0.,0.,0.), 0);
 
     G4double inn_offset = .5 * cm;
-    inner_air_gen_ = new BoxPointSampler(shield_x_ - inn_offset, shield_y_ - inn_offset, shield_z_ - inn_offset,
-                                         1. * mm, G4ThreeVector(0.,0.,0.), 0);
+    inner_air_gen_ = new BoxPointSampler(shield_x_ - inn_offset, shield_y_ - inn_offset, shield_z_ - inn_offset, 1. * mm,
+                                         G4ThreeVector(0.,0.,0.), 0);
 
+    lat_roof_gen_ = new BoxPointSampler(lead_thickness_, beam_base_thickness_, shield_z_ + 2.*steel_thickness_, 0.,
+                                        G4ThreeVector(0., shield_y_/2. + steel_thickness_ + beam_base_thickness_/2., 0.), 0);
 
-    lat_roof_gen_ =
-      new BoxPointSampler(lead_thickness_,beam_base_thickness_,shield_z_,0.,
-			  G4ThreeVector(0.,shield_y_/2.+steel_thickness_/2.,0.),0);
-    front_roof_gen_ =
-      new BoxPointSampler(lead_x_,beam_base_thickness_,lead_thickness_,0.,
-			  G4ThreeVector(0.,shield_y_/2.+steel_thickness_/2.,0.),0);
-    // struct_gen_=;
-    struct_x_gen_ = new BoxPointSampler((shield_x_+2*lead_thickness_+2*steel_thickness_), lead_thickness_, beam_base_thickness_,0.,
-					G4ThreeVector(0.,top_beam_y,roof_z_separation_+lateral_z_separation_/2),0);
-    struct_z_gen_ = new BoxPointSampler( beam_base_thickness_, lead_thickness_ -1.*mm, shield_z_+2*lead_thickness_+2*steel_thickness_,0.,
-					 G4ThreeVector(-front_x_separation_/2.,top_beam_y, 0.), 0);
-    lat_beam_gen_ = new BoxPointSampler(lead_thickness_, shield_y_+steel_thickness_,beam_base_thickness_,0.,
-					G4ThreeVector(lat_beam_x,-beam_base_thickness_/2.,lateral_z_separation_/2.),0);
-    front_beam_gen_ = new BoxPointSampler(beam_base_thickness_, shield_y_+steel_thickness_,lead_thickness_,0.,
-					G4ThreeVector(-front_x_separation_/2.,-beam_base_thickness_/2.,front_beam_z),0);
+    front_roof_gen_ = new BoxPointSampler(lead_x_, beam_base_thickness_, lead_thickness_, 0.,
+                                          G4ThreeVector(0., shield_y_/2. + steel_thickness_ + beam_base_thickness_/2., 0.), 0);
+
+    struct_x_gen_ = new BoxPointSampler(shield_x_ + 2.*lead_thickness_ + 2.*steel_thickness_, lead_thickness_ - beam_base_thickness_, beam_base_thickness_, 0.,
+                                        G4ThreeVector(0., top_beam_y, roof_z_separation_+lateral_z_separation_/2.), 0);
+
+    struct_z_gen_ = new BoxPointSampler(beam_base_thickness_, lead_thickness_ - beam_base_thickness_, shield_z_ + 2.*lead_thickness_ + 2.*steel_thickness_, 0.,
+                                        G4ThreeVector(-front_x_separation_/2., top_beam_y, 0.), 0);
+
+    lat_beam_gen_ = new BoxPointSampler(lead_thickness_, shield_y_ + 2. * steel_thickness_+ lead_thickness_, beam_base_thickness_, 0.,
+                                        G4ThreeVector(lat_beam_x, -lead_thickness_/2., lateral_z_separation_/2.), 0);
+
+    front_beam_gen_ = new BoxPointSampler(beam_base_thickness_, shield_y_ + 2. * steel_thickness_+ lead_thickness_, lead_thickness_, 0.,
+                                          G4ThreeVector(-front_x_separation_/2., -lead_thickness_/2., front_beam_z), 0);
 
 
     // Calculating some probs
@@ -261,13 +262,15 @@ namespace nexus {
     //std::cout<<"TOP STRUCT VOLUME "<<struct_top_vol<<std::endl;
     G4double lateral_vol = lat_beam_solid->GetCubicVolume();
     //std::cout<<"LAT BEAM STRUCT VOLUME "<<lateral_vol<<"\t TOTAL LATERAL BEAMS VOL "<<8*lateral_vol<<std::endl;
-    G4double total_vol = roof_vol+struct_top_vol+(8*lateral_vol);
+    G4double total_vol = roof_vol + struct_top_vol + (8*lateral_vol);
     //std::cout<<"TOTAL STRUCTURE VOLUME "<<total_vol<<std::endl;
 
-    perc_roof_vol_ = roof_vol/total_vol;
-    perc_front_roof_vol_ = 2*(lead_x_*beam_base_thickness_*lead_thickness_) /roof_vol;
+    perc_roof_vol_       = roof_vol/total_vol;
+    perc_front_roof_vol_ = 2*(lead_x_*beam_base_thickness_*lead_thickness_)/roof_vol;
     perc_top_struct_vol_ = struct_top_vol /total_vol;
-    perc_struc_x_vol_ = 4*((shield_x_+2*lead_thickness_+2*steel_thickness_)*lead_thickness_*beam_base_thickness_)/struct_top_vol;
+
+    G4double lat_beam_x_vol = (shield_x_ + 2.*lead_thickness_ + 2.*steel_thickness_)*(lead_thickness_ - beam_base_thickness_)*beam_base_thickness_;
+    perc_struc_x_vol_    = 4*lat_beam_x_vol/struct_top_vol;
 
     // std::cout<<"SHIELDING LEAD VOLUME:\t"<<lead_box_solid->GetCubicVolume()<<std::endl;
     // std::cout<<"SHIELDING STEEL VOLUME:\t"<<steel_box_solid->GetCubicVolume()<<std::endl;
@@ -306,17 +309,17 @@ namespace nexus {
     G4ThreeVector vertex(0., 0., 0.);
 
     if (region == "SHIELDING_LEAD") {
-      G4VPhysicalVolume *VertexVolume;
-      do {
-	vertex = lead_gen_->GenerateVertex("WHOLE_VOL");
-	// To check its volume, one needs to rotate and shift the vertex
-	// because the check is done using global coordinates
-	G4ThreeVector glob_vtx(vertex);
-	// First rotate, then shift
-	glob_vtx.rotate(pi, G4ThreeVector(0., 1., 0.));
-	glob_vtx = glob_vtx + G4ThreeVector(0, 0, GetELzCoord());
-	VertexVolume = geom_navigator_->LocateGlobalPointAndSetup(glob_vtx, 0, false);
-      } while (VertexVolume->GetName() != "LEAD_BOX");
+        G4VPhysicalVolume *VertexVolume;
+        do {
+          	vertex = lead_gen_->GenerateVertex("WHOLE_VOL");
+          	// To check its volume, one needs to rotate and shift the vertex
+          	// because the check is done using global coordinates
+          	G4ThreeVector glob_vtx(vertex);
+          	// First rotate, then shift
+          	glob_vtx.rotate(pi, G4ThreeVector(0., 1., 0.));
+          	glob_vtx = glob_vtx + G4ThreeVector(0, 0, GetELzCoord());
+          	VertexVolume = geom_navigator_->LocateGlobalPointAndSetup(glob_vtx, 0, false);
+        } while (VertexVolume->GetName() != "LEAD_BOX");
     }
 
     else if (region == "SHIELDING_STEEL") {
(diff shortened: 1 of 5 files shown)


source/geometries/Next100Shielding.cc
```diff
@@ -36,27 +36,25 @@ namespace nexus {
...
+    front_x_separation_  (156. * mm), //distance between the two front beams
     // Box thickness
     lead_thickness_ (20. * cm),
-    steel_thickness_ (6. * mm),
+    steel_thickness_(2.0 * mm),
     visibility_ (0)
 
   {
-
     /// Shielding is compound by two boxes, the external made of lead,
```
paolafer: I would change the description to something like "The shielding is made of two boxes. The external one is made of lead and the internal one is made of a mix of steel and titanium. The steel beam structure is placed inside the lead." Is it correct that the internal box is made of steel and titanium?

I would also add a brief explanation about how we simulate the beam structure.

source/geometries/Next100Shielding.cc
```diff
@@ -273,13 +262,15 @@ namespace nexus {
...
-    perc_front_roof_vol_ = 2*(lead_x_*beam_base_thickness_*lead_thickness_) /roof_vol;
+    perc_roof_vol_       = roof_vol/total_vol;
+    perc_front_roof_vol_ = 2*(lead_x_*beam_base_thickness_*lead_thickness_)/roof_vol;
     perc_top_struct_vol_ = struct_top_vol /total_vol;
-    perc_struc_x_vol_ = 4*((shield_x_+2*lead_thickness_+2*steel_thickness_)*lead_thickness_*beam_base_thickness_)/struct_top_vol;
+
+    G4double lat_beam_x_vol = (shield_x_ + 2.*lead_thickness_ + 2.*steel_thickness_)*(lead_thickness_ - beam_base_thickness_)*beam_base_thickness_;
+    perc_struc_x_vol_    = 4*lat_beam_x_vol/struct_top_vol;
 
     // std::cout<<"SHIELDING LEAD VOLUME:\t"<<lead_box_solid->GetCubicVolume()<<std::endl;
```
paolafer: I would start using a verbosity configuration parameters in the geometry classes (following what @jmunozv already did with NextFlex), which, if set to `true` prints out these comments. I would add units and a bit more verbosity to explain what is printed.

source/geometries/Next100Shielding.cc
```diff
@@ -70,8 +68,6 @@ namespace nexus {
 
   void Next100Shielding::Construct()
   {
-    // Auxiliary solids
-    //   G4Box* shielding_box_solid = new G4Box("SHIELD_BOX", shield_x_/2., shield_y_/2., shield_z_/2.);
 
     // LEAD BOX   ///////////
     lead_x_ = shield_x_ + 2. * steel_thickness_ + 2. * lead_thickness_;
```
paolafer, line 93: The `y` dimension has to take into account the thickness of the beam base.

source/geometries/Next100Shielding.cc
```diff
@@ -36,27 +36,25 @@ namespace nexus {
...
+    shield_y_ (166.  * cm),
+    shield_z_ (259.4 * cm),
 
     //Steel Structure
-    beam_base_thickness_ (4. *mm),
-    lateral_z_separation_ (1006. *mm), //distance between the two lateral beams
-    roof_z_separation_ (755. *mm), //distance between 1st and 2nd roof beams
-    front_x_separation_ (154.*mm), //distance between the two front beams
+    beam_base_thickness_ (4.   * mm),
+    lateral_z_separation_(1010.* mm), //distance between the two lateral beams
```
paolafer: Are these the distances between the centres of the beams or between the sides? Please, specify. If it's the second case, the union of solids below needs to be reviewed.

source/geometries/Next100Shielding.cc
```diff
@@ -86,127 +82,120 @@ namespace nexus {
...
-    G4SubtractionSolid* roof_beam_solid = new G4SubtractionSolid("STRUCT_BEAM",roof_beam,aux_box,0,G4ThreeVector(0.,0.,0.));
-
-    G4Box* lat_beam_solid = new G4Box("STRUCT_BEAM",
-				      lead_thickness_/2.,
-				      shield_y_/2.+steel_thickness_/2.,
-				      beam_base_thickness_/2.);
+    // auxiliar positions used in translations
+    G4double lat_beam_x   = shield_x_/2. + steel_thickness_ + lead_thickness_/2.;
+    G4double front_beam_z = shield_z_/2. + steel_thickness_ + lead_thickness_/2.;
+    G4double top_beam_y   = shield_y_/2. + steel_thickness_ + (lead_thickness_ + beam_base_thickness_)/2.;
```
paolafer: I would say that `beam_base_thickness_` has to be added, not half.

paolafer: The same should be reviewed in all the dimensions and placements regarding the `y` coordinate.

paolafer (changes requested):

source/geometries/Next100Shielding.cc
gondiaz: Yes, following Jordi's drawing the internal steel box is made of "Stainless steel AISI-316Ti". Shoudl we inplement the box with this material? Not sure if it exists in G4. On the other hand, the castle structure is made of steel S-275.

source/geometries/Next100Shielding.cc
gondiaz: From Jordi's drawings I think it is between the beam centers. If it were not the case, the total distance (see lateral view in slide 5) would not sum up, since we had to add two times the beam thickness.

source/geometries/Next100Shielding.cc
```diff
@@ -36,46 +36,45 @@ namespace nexus {
     GeometryBase(),
 
     // Shielding internal dimensions
-    shield_x_ (155.  * cm),
-    shield_y_ (225.6 * cm),
-    shield_z_ (258.5 * cm), // 253.0 * cm before May 12, 2017
+    shield_x_ (158.  * cm),
+    shield_y_ (166.  * cm),
```
jmalbos: According to the drawings, this should be 166.6 cm, right?

source/geometries/Next100Shielding.cc
```diff
@@ -36,46 +36,45 @@ namespace nexus {
     GeometryBase(),
 
     // Shielding internal dimensions
-    shield_x_ (155.  * cm),
-    shield_y_ (225.6 * cm),
-    shield_z_ (258.5 * cm), // 253.0 * cm before May 12, 2017
+    shield_x_ (158.  * cm),
```
jmalbos: I thought we had decided to start using braces for initialisation: `shield_x_ {158. * cm}`

source/geometries/Next100Shielding.cc
gondiaz: Right

paolafer:
Still to be implemented: different thickness of the beams (4 and 6 mm), two beams below the vessel, which support it, and should be simulated, since they are not shielded by the lead bricks.

source/geometries/Next100Shielding.cc
```diff
@@ -86,127 +85,122 @@ namespace nexus {
...
-    G4Box* lat_beam_solid = new G4Box("STRUCT_BEAM",
-				      lead_thickness_/2.,
-				      shield_y_/2.+steel_thickness_/2.,
-				      beam_base_thickness_/2.);
+    // auxiliar positions used in translations
+    G4double lat_beam_x   = shield_x_/2. + steel_thickness_ + lead_thickness_/2.;
+    G4double front_beam_z = shield_z_/2. + steel_thickness_ + lead_thickness_/2.;
+    G4double top_beam_y   = (shield_y_-beam_thickness_)/2. + steel_thickness_ + beam_thickness_ + lead_thickness_/2.;
+    G4double lat_beam_y   = -(lead_thickness_ + beam_thickness_)/2.;
+    G4double roof_y       = (shield_y_-beam_thickness_)/2. + steel_thickness_ + beam_thickness_/2.;
```
paolafer: Isn't it `roof_y       = shield_y_/2. + steel_thickness_ + beam_thickness_/2.`?

source/geometries/Next100Shielding.cc
```diff
@@ -86,127 +85,122 @@ namespace nexus {
...
-    G4SubtractionSolid* roof_beam_solid = new G4SubtractionSolid("STRUCT_BEAM",roof_beam,aux_box,0,G4ThreeVector(0.,0.,0.));
-
-    G4Box* lat_beam_solid = new G4Box("STRUCT_BEAM",
-				      lead_thickness_/2.,
-				      shield_y_/2.+steel_thickness_/2.,
-				      beam_base_thickness_/2.);
+    // auxiliar positions used in translations
+    G4double lat_beam_x   = shield_x_/2. + steel_thickness_ + lead_thickness_/2.;
+    G4double front_beam_z = shield_z_/2. + steel_thickness_ + lead_thickness_/2.;
+    G4double top_beam_y   = (shield_y_-beam_thickness_)/2. + steel_thickness_ + beam_thickness_ + lead_thickness_/2.;
```
paolafer: Isn't it `top_beam_y   = shield_y_/2. + steel_thickness_ + beam_thickness_ + lead_thickness_/2.` ?

paolafer (changes requested):

source/geometries/Next100Shielding.cc
gondiaz: No, because now the lead box center is "displaced" by `beam_thickness_/2` because the `lead_y_` dimension has and extra term of ` beam_thickness_`

paolafer: You are right, it's correct.

source/geometries/Next100Shielding.cc
```diff
@@ -86,128 +93,182 @@ namespace nexus {
...
+                      pedestal_support_beam_top_logic, "PEDESTAL_SUPPORT_BEAM_TOP_2", air_box_logic_, false, 0);
+
+    new G4PVPlacement(0, G4ThreeVector(0., pedestal_y_pos, support_beam_dist_/2. + support_lat_dist_),
+                      pedestal_beam_lat_logic, "PEDESTAL_BEAM_LAT_1", lead_box_logic, false, 0);
+
+    new G4PVPlacement(0, G4ThreeVector(0., pedestal_y_pos, -support_beam_dist_/2. - support_lat_dist_),
+                      pedestal_beam_lat_logic, "PEDESTAL_BEAM_LAT_2", lead_box_logic, false, 0);
 
     // SETTING VISIBILITIES   //////////
     if (visibility_) {
```
paolafer, line 340: I would add the visibility behaviour of the new pedestal volumes. We can set the same one as the rest of the structure, for instance, and make them invisible if visibility is zero.

source/geometries/Next100Shielding.cc
```diff
@@ -217,74 +311,135 @@ namespace nexus {
       steel_box_logic->SetVisAttributes(grey_col);
       G4VisAttributes antiox_col = nexus::BloodRed();
       //  antiox.SetForceSolid(true);
-      roof_logic->SetVisAttributes(antiox_col);
+      roof_logic    ->SetVisAttributes(antiox_col);
       lat_beam_logic->SetVisAttributes(antiox_col);
-      struct_logic->SetVisAttributes(antiox_col);
+      struct_logic  ->SetVisAttributes(antiox_col);
```
paolafer: `front_beam_logic` should be also added here: in the current version of the code, the front beams are visible even when `visibility == 0`.

source/geometries/Next100Shielding.cc
```diff
@@ -217,74 +311,135 @@ namespace nexus {
...
     }
     else {
-      lead_box_logic->SetVisAttributes(G4VisAttributes::Invisible);
+      lead_box_logic ->SetVisAttributes(G4VisAttributes::Invisible);
       steel_box_logic->SetVisAttributes(G4VisAttributes::Invisible);
-      roof_logic->SetVisAttributes(G4VisAttributes::Invisible);
-      struct_logic->SetVisAttributes(G4VisAttributes::Invisible);
-      lat_beam_logic->SetVisAttributes(G4VisAttributes::Invisible);
+      roof_logic     ->SetVisAttributes(G4VisAttributes::Invisible);
```
paolafer: Same here: add the same color for the front beams.

paolafer:
Regarding the vertices: we could probably drop the `INNER_AIR` region, since we are not generating background from the air inside the castle, right @jmunozv, @msorel? On the other hand, the `SHIELDING_STEEL` and `PEDESTAL` regions must be reviewed, because the vertices turn out to be generated in different volumes, so there must be some mismatch of dimensions/positions.

source/geometries/Next100Shielding.cc
```diff
@@ -217,74 +311,135 @@ namespace nexus {
...
+    if (verbosity_){
+      std::cout<<"STEEL STRUCTURE VOLUME (m3)" <<std::endl;
+      std::cout<<"ROOF BEAM     "<< roof_vol      /1.e9 <<std::endl;
+      std::cout<<"TOP STRUCT    "<< struct_top_vol/1.e9 <<std::endl;
+      std::cout<<"LATERAL BEAMS "<< 8*lateral_vol /1.e9 <<std::endl;
+      std::cout<<"TOTAL         "<< total_vol     /1.e9 <<std::endl;
+
+      G4double inner_vol = air_box_solid  ->GetCubicVolume();
+      G4double steel_vol = steel_box_solid->GetCubicVolume() - inner_vol;
+      G4double lead_vol  = lead_box_solid ->GetCubicVolume() - steel_vol;
```
paolafer: Fix the definition of `lead_vol`.

msorel:
> Regarding the vertices: we could probably drop the `INNER_AIR` region, since we are not generating background from the air inside the castle, right @jmunozv, @msorel? On the other hand, the `SHIELDING_STEEL` and `PEDESTAL` regions must be reviewed, because the vertices turn out to be generated in different volumes, so there must be some mismatch of dimensions/positions.

Concerning INNER_AIR, if the vertex generator does not do any harm, perhaps we can keep it. In normal operations, we will take NEXT-100 low-bgr data always with Rn abatement system on, so we won't generate radon in the air there. But perhaps one day we will want to estimate the contributions of Rn system being off, you never know... If it is not a problem, we could leave this capability in, unless I am missing something.

source/geometries/Next100Shielding.cc
```diff
@@ -33,49 +33,65 @@ namespace nexus {
...
-    visibility_ (0)
+    lead_thickness_ {20.* cm},
+    steel_thickness_{2. * mm},
+
+    // Pedestal
+    pedestal_x_        {1518. * mm},
+    pedestal_top_x_    {1384. * mm},
+    support_beam_dist_ {1027. * mm},
+    support_front_dist_{732.  * mm},
+    pedestal_lateral_beam_thickness {8. * mm},
```
paolafer: Add trailing underscore in this line and the following, according to the nexus conventions.

source/geometries/Next100Shielding.cc
```diff
@@ -217,74 +315,165 @@ namespace nexus {
...
+                                               G4ThreeVector(0., ped_gen_y_, support_beam_dist_/2.), 0);
+
+    ped_front_gen_ = new BoxPointSampler(pedestal_x_, lead_thickness_, pedestal_front_beam_thickness, 0.,
+                                       G4ThreeVector(0., pedestal_y_pos, support_beam_dist_/2. + support_front_dist_), 0);
+
+    ped_lateral_gen_ = new BoxPointSampler(pedestal_lateral_beam_thickness, lead_thickness_, pedestal_lateral_length, 0.,
+                                           G4ThreeVector(pedestal_x_pos, pedestal_y_pos, 0.), 0);
+
+    // pedestal roof
+    G4double ped_roof_gen_x_ = pedestal_top_x_/2. + pedestal_roof_thickness/2.;
```
paolafer: Remove trailing underscore from the variable defined in lines 404, 405, 406, according to nexus conventions.

source/geometries/Next100Shielding.cc
```diff
@@ -217,74 +315,165 @@ namespace nexus {
...
+
+
+    // BUBBLE SEAL
+    bubble_seal_front_gen_ = new BoxPointSampler(pedestal_x_ + 2*pedestal_lateral_beam_thickness, bubble_seal_thickness, bubble_seal_thickness, 0.,
+                                                 G4ThreeVector(0., ped_gen_y_, 0.), 0);
+
+    bubble_seal_lateral_gen_ = new BoxPointSampler(bubble_seal_thickness, bubble_seal_thickness, pedestal_lateral_length, 0.,
+                                                   G4ThreeVector(0., ped_gen_y_, 0.), 0);
+
+    G4double bubble_front_vol_   = 2*(pedestal_x_ + 2*pedestal_lateral_beam_thickness)*(bubble_seal_thickness)*(bubble_seal_thickness);
```
paolafer: Remove trailing underscores, as above.

source/geometries/Next100Shielding.cc
```diff
@@ -217,74 +315,165 @@ namespace nexus {
...
+    perc_buble_lateral_vol_ = bubble_lateral_vol_/(bubble_front_vol_ + bubble_lateral_vol_);
+
+    // EDPM SEAL
+    edpm_seal_front_gen_ = new BoxPointSampler(edpm_seal_thickness, steel_y, edpm_seal_thickness, 0.,
+                                               G4ThreeVector(0., -beam_thickness_2/2., 0.), 0);
+
+    edpm_seal_lateral_gen_ = new BoxPointSampler(edpm_seal_thickness, edpm_seal_thickness, steel_z, 0.,
+                                                G4ThreeVector(0., -beam_thickness_2/2., 0.), 0);
+
+    G4double edpm_front_vol_   = 2*edpm_seal_thickness * steel_y * edpm_seal_thickness;
```
paolafer: Same as above.

source/geometries/Next100Shielding.cc
```diff
@@ -343,123 +550,224 @@ namespace nexus {
...
-      	    vertex = front_roof_gen_->GenerateVertex("INSIDE");
-      	    vertex.setZ(vertex.z() + (shield_z_/2.+steel_thickness_+lead_thickness_/2.));
-	    // std::cout<<"frontal +"<<std::endl;
+      	    vertex = ped_support_bottom_gen_->GenerateVertex("INSIDE");
       	  }
       	  else{
-      	    vertex = front_roof_gen_->GenerateVertex("INSIDE");
-      	    vertex.setZ(vertex.z() - (shield_z_/2.+steel_thickness_+lead_thickness_/2.));
-      	    // std::cout<<"frontal -"<<std::endl;
+      	    vertex = ped_support_bottom_gen_->GenerateVertex("INSIDE");
```
paolafer: Instead of generating the vertex twice, couldn't you just use the same vertex and shift it if the random number is less than 0.5? The same for the rest of the code and also in the `SHIELDING_STRUCT` generation. Please, tell me if I'm not seeing something here.

paolafer (changes requested):

source/geometries/Next100Shielding.cc
```diff
@@ -343,123 +550,224 @@ namespace nexus {
...
-	  vertex.setZ(vertex.z() -(shield_z_+2*steel_thickness_+lead_thickness_));
-	}
       }
 
+    else if(region=="EDPM_SEAL"){
+      G4double rand = G4UniformRand();
+      if (rand<perc_edpm_front_vol_){ // front
+        if (G4UniformRand() < 0.5){
+          vertex = edpm_seal_front_gen_->GenerateVertex("INSIDE");
+          vertex.setZ(vertex.z() + (shield_z_/2. + steel_thickness_ + edpm_seal_thickness/2.));
```
paolafer: The vertices are being generated into the `LEAD_BOX` volume, but this is not the intended behaviour, right? They should be generated just on the surface of the steel box, because they face the inner air.  I think the displacement should be 
`vertex.setZ(vertex.z() + (shield_z_/2. - edpm_seal_thickness/2.));`, do you agree?

source/geometries/Next100Shielding.cc
```diff
@@ -343,123 +550,224 @@ namespace nexus {
...
-	    vertex = struct_z_gen_->GenerateVertex("INSIDE");
-	    vertex.setX(vertex.x()+front_x_separation_);
-	  }
-	}
+    else if(region=="BUBBLE_SEAL"){
+      G4double rand = G4UniformRand();
+      if (rand<perc_bubble_front_vol_){ // front
+        if (G4UniformRand() < 0.5){
+          vertex = bubble_seal_front_gen_->GenerateVertex("INSIDE");
+          vertex.setZ(vertex.z() + (support_beam_dist_/2. + support_front_dist_ + pedestal_front_beam_thickness/2. + bubble_seal_thickness/2.));
```
paolafer: To simplify the code further and make it more readable, you can put the value to add or subtract in a different variable outside the `if (G4UniformRand() < 0.5)`. The same in `EDPM_SEAL`.

paolafer:
In line with the [coding conventions](https://github.com/next-exp/nexus/wiki/Coding-conventions-and-style-guide) of nexus, try to keep all the lines between 80 and 100 characters.

source/geometries/Next100Shielding.cc
gondiaz: It is not still clear, since from the slides it seems that the steel is above the seal. Let's wait for Jordi's answer to finally set the exact position.

gondiaz: After Jordi's confirmation, we move the EDPM vertex to be generated from inside the INNER_AIR.

paolafer (approved):
This PR makes a thorough revision of the geometry of the external shielding, fixing some inaccuracies and bugs and checking all the dimensions with the drawings from Jordi. The code works as intended and has been reviewed in detail, so I approve it. Excellent job, @gondiaz!

PR #112, opened 2021-05-28
Fix demopp anode grid

gondiaz:
This PR solves a pernicious bug in which the anode grid was only constructed for run7 and run8. I have replaced the `else if` statement by an `else` to avoid having the same issue in the future.


halmamol (approved):
The PR fixes the bug not building the ANODE for run9, and it avoids having the same issue with future configurations. I approve it.

PR #113, opened 2021-05-31
Update CMake building system

jmalbos:
This pull request updates the CMake building scripts.


jmalbos:
@paolafer, can you please test this in your laptop?

paolafer:
> @paolafer, can you please test this in your laptop?

Yes, it gives me an error in the configuration step, it can't find HDF5. I'm using `-DHDF5_ROOT=/path/to/hdf5/installation/directory`

jmalbos:
I should have provided some instructions, sorry. After sourcing `geant4.sh` and `thisroot.sh`, do the following from the _nexus_ folder:
```
mkdir build
cmake -DGSL_ROOT_DIR=/path/to/gsl -DHDF5_ROOT=/path/to/hdf5 -DCMAKE_INSTALL_PREFIX=. -S . -B build
cmake --build build 
```
Alternatively, CMake will find GSL and HDF5 if they are in a common location or if you have their `bin` folder added to `PATH`.

paolafer:
Yes, that's what I did (although installing it in a different path), but apparently CMake can't locate HDF5. I tried again with exactly your instructions and the same problem appears.

paolafer:
The building fails if a version of CMake older than 3.12 is used, without giving any hint about the reason. It would be useful to add a meaningful message error.

paolafer:
Also, the headers files are copied to the `include/nexus/include` folder, while it should be `include/nexus`.

jmalbos:
> The building fails if a version of CMake older than 3.12 is used, without giving any hint about the reason. It would be useful to add a meaningful message error.

There is already a command in the script for that:
`cmake_minimum_required(VERSION 3.12)`

According to CMake's documentation, "[if] the running version of CMake is lower than the <min> required version it will stop processing the project and report an error", so I don't understand why you would not be getting an error message.

paolafer:
> According to CMake's documentation, "[if] the running version of CMake is lower than the required version it will stop processing the project and report an error", so I don't understand why you would not be getting an error message.

Can you try on a machine with an older CMake version? I'm pretty sure neutrinos doesn't have a recent one.

jmalbos:
I think I know what the problem is: Only recent versions of CMake understand the `-S` and `-B` flags I indicated in my instructions, so in your system it is not even reading the script. For older versions, you have to do this from the nexus top folder:
```
mkdir build; cd build;
cmake -DGSL_ROOT_DIR=/path/to/gsl -DHDF5_ROOT=/path/to/hdf5 -DCMAKE_INSTALL_PREFIX=.. ..
cmake --build .
```

paolafer:
Ok, so, which instructions do you think it is better to write in the wiki? The old style ones?

jmalbos:
Those flags were added, apparently, in version 3.13. I guess the simplest thing is to bump up the minimum version to that one and indicate it in the installation instructions of the wiki.

paolafer (approved):
This PR updates the CMake building system and adds the creation of a nexus library to be used by external software. The building works well, provided a recent enough version of CMake is used, I approve it!

jacg:
It's worth reminding ourselves that Nix avoids all the above problems by design. With Nix you never have to test anything on any specific developer's laptop (OpenGL dependencies being the only exception I have found to this), wonder what will happen if some funny version of `cmake` is installed or give instructions on how to configure paths so that HDF5 libraries ore some development headers are found, for the simple reason that Nix provides the *precise* build environment needed to build any given version of the project. Consequently

+ it doesn't matter if `cmake` is installed at all on the target machine, let alone what version;
+ no instructions need to be given in any wiki, besides how to install Nix itself, which will be identical for *all* projects;
+ and so on.

You only have to explain the environment to Nix just the once (well, twice: once for Linux, a second time for MacOS), and you're done. Guaranteed to work on everyone's machine, subject to Nix being installed there. When the build instructions change, the developers don't even notice.

A major blocker for using Nix in this project in particular (besides the availability of Nix on HPC machines not administered by us), is that y'all use Macs, and, last time I looked, Geant4 wasn't packaged up in Nix for MacOS. Unfortunately I have neither the time nor the expertise to do this for you.

These shortcomings should be taken by the community as motivation to overcome them, rather than a reason to avoid Nix altogether.

I have no doubt that the future of system configuration is, if not Nix itself, then something based on the ideas that Nix brings to the table. It would good to be ahead of the curve.

PR #114, opened 2021-06-03
Prblackbox

MiryamMV:
More recent and improved (hopefully cleaner) version of the BlackBox geometry. It includes the options for SiPM boards with no mask, mask with circular holes and mask for rectangular holes.


source/geometries/BlackBox.cc
```diff
@@ -0,0 +1,212 @@
...
+#include <G4VisAttributes.hh>
+#include <G4Colour.hh>
+#include <G4SDManager.hh>
+#include <G4UserLimits.hh>
+#include <Randomize.hh>
+#include <G4LogicalSkinSurface.hh>
+#include <G4OpticalSurface.hh>
+#include <G4RotationMatrix.hh>
+
+#include <CLHEP/Units/SystemOfUnits.h>
```
jmalbos: Replace this header by `G4SystemOfUnits.hh`.

source/geometries/BlackBox.cc
```diff
@@ -0,0 +1,212 @@
...
+
+namespace nexus {
+
+  REGISTER_CLASS(BlackBox, GeometryBase)
+  using namespace CLHEP;
```
jmalbos: If you include Geant4's header (see the comment above), you won't need this.

source/geometries/BlackBoxSiPMBoard.cc
```diff
@@ -0,0 +1,344 @@
...
+#include <G4SubtractionSolid.hh>
+#include <G4PVPlacement.hh>
+#include <G4RotationMatrix.hh>
+#include <G4Material.hh>
+#include <G4NistManager.hh>
+#include <G4OpticalSurface.hh>
+#include <G4LogicalSkinSurface.hh>
+#include <G4LogicalBorderSurface.hh>
+
+#include <CLHEP/Units/SystemOfUnits.h>
```
jmalbos: Ditto.

paolafer:
What is the status of this PR @MiryamMV, @jmalbos?

source/geometries/BlackBox.cc
```diff
@@ -0,0 +1,210 @@
...
+    box_z_ (2. * m),
+    box_xy_ (1. *m),
+
+    visibility_(0)
+  {
+    msg_ = new G4GenericMessenger(this, "/Geometry/BlackBox/",
+				  "Control commands of BlackBox.");
+    msg_->DeclareProperty("visibility", visibility_, "Giant detectors visibility");
+
+    G4GenericMessenger::Command&  specific_vertex_X_cmd =
```
paolafer: Please, change these 3 parameters to one, using `DeclarePropertyWithUnit` (you can look at the `NextNew` class as an example).

macros/black_box.config.mac
```diff
@@ -0,0 +1,38 @@
...
+### This init macro simulates the response of a SensL SiPM DICE
+### board in the black box.
+### -----------------------------------------------------------
+
+/Generator/SingleParticle/particle opticalphoton
+#/Generator/SingleParticle/particle geantino
+/Generator/SingleParticle/min_energy 2.9 eV
+/Generator/SingleParticle/max_energy 2.9 eV
+/Generator/SingleParticle/region AD_HOC
+#/Generator/SingleParticle/momentum_X 0.
```
paolafer: Remove commented lines.

macros/configs/black_box_single.config.mac
```diff
@@ -0,0 +1,27 @@
+### -----------------------------------------------------------
+### This init macro simulates the response of a SensL SiPM
+### in the black box.
+### -----------------------------------------------------------
+
+/Generator/SingleParticle/particle opticalphoton
+/Generator/SingleParticle/min_energy 2.9 eV
+/Generator/SingleParticle/max_energy 2.9 eV
+/Generator/SingleParticle/region AD_HOC
+/Generator/SingleParticle/momentum_X 0.
```
paolafer: Replace these 3 lines with `/Generator/SingleParticle/momentum`.

macros/configs/black_box_single.config.mac
```diff
@@ -0,0 +1,27 @@
+### -----------------------------------------------------------
```
paolafer: What's the purpose of this `configs` folder? Is this macro used? I don't see the correspondent `init` macro.

source/geometries/BlackBox.cc
MiryamMV: Done!

macros/black_box.config.mac
MiryamMV: Done!

macros/configs/black_box_single.config.mac
MiryamMV: Done!

macros/configs/black_box_single.config.mac
MiryamMV: This folder is just a residue of the first stages of the PR. I'll remove it since it has no purpose now.

paolafer:
Some tests are failing, can you have a look at this?

source/geometries/BlackBox.cc
```diff
@@ -0,0 +1,296 @@
...
+  // SAPPHIRE //////////////////////////////////////////////////
+
+  G4String sapphire_name = "SAPPHIRE";
+
+  G4Material* sapphire_mat = materials::Sapphire();
+
+  sapphire_mat->SetMaterialPropertiesTable(opticalprops::Sapphire());
+
+  //G4double pedot_thickn  = 1. *um;
+  G4double tpb_thickn    = 1. *um;
```
paolafer: This variable is not used, please remove it if it is not needed, otherwise, review your code to use it.

macros/test.init.mac
```diff
@@ -0,0 +1,25 @@
+## ----------------------------------------------------------------------------
```
paolafer: This macro is fine, but I would use a name that explains what it does. Maybe `vertex_test`?

source/geometries/BlackBox.cc
```diff
@@ -0,0 +1,296 @@
+// ----------------------------------------------------------------------------
+//  nexus | BlackBox.cc
```
paolafer: Please, use the same format for the description, as for the other classes, namely a short, but not so short description of the class and `The NEXT Collaboration`. Same for the header file.

source/geometries/BlackBox.cc
```diff
@@ -0,0 +1,296 @@
...
+    world_z_ (3. * m),
+    world_xy_ (2. *m),
+    box_z_ (2. * m),
+    box_xy_ (1. *m),
+
+    visibility_(0)
+  {
+    msg_ = new G4GenericMessenger(this, "/Geometry/BlackBox/",
+				  "Control commands of BlackBox.");
+    msg_->DeclareProperty("visibility", visibility_, "Giant detectors visibility");
```
paolafer: Why `Giant detectors`?

source/geometries/BlackBox.cc
```diff
@@ -0,0 +1,296 @@
...
+
+    G4GenericMessenger::Command&  hole_y_cmd =
+      msg_->DeclareProperty("hole_y", hole_y_,
+                            "Hole Y");
+    hole_y_cmd.SetParameterName("hole_y", true);
+    hole_y_cmd.SetUnitCategory("Length");
+
+    msg_->DeclareProperty("membrane_hole", membrane_hole_,
+                          "True if membrane has holes, false if not.");
+    //G4GenericMessenger::Command& membrane_hole_cmd =msg_->DeclareProperty("membrane_hole", membrane_hole_,
```
paolafer: Remove the commented lines.

source/geometries/BlackBox.cc
```diff
@@ -0,0 +1,296 @@
...
+  G4LogicalVolume* world_logic_vol =
+    new G4LogicalVolume(world_solid_vol, world_mat, world_name);
+  world_logic_vol->SetVisAttributes(G4VisAttributes::GetInvisible());
+  GeometryBase::SetLogicalVolume(world_logic_vol);
+
+  // BLACK BOX /////////////////////////////////////////////////
+
+  G4String box_name = "BLACK BOX";
+
+  G4Material* box_mat = G4NistManager::Instance()->FindOrBuildMaterial("G4_AIR");
```
paolafer: You are defining the same material twice (`G4AIR`): remove the second instance and use the one you have defined above.

source/geometries/BlackBox.cc
```diff
@@ -0,0 +1,296 @@
...
+				  "Control commands of BlackBox.");
+    msg_->DeclareProperty("visibility", visibility_, "Giant detectors visibility");
+
+    msg_->DeclarePropertyWithUnit("specific_vertex", "mm",  specific_vertex_,
+      "Set generation vertex.");
+
+    G4GenericMessenger::Command&  dice_board_x_pos_cmd =
+      msg_->DeclareProperty("dice_board_x_pos", dice_board_x_pos_,
+                            "X position of the dice");
+    dice_board_x_pos_cmd.SetParameterName("dice_board_x_pos", true);
```
paolafer: All the parameters should have a default value. If the user doesn't specify them in the configuration file, they take a value which may be wrong. On the other hand, if you want to force the user to specify a custom value, you should make the program fail if that is not done.

source/geometries/BlackBox.cc
```diff
@@ -0,0 +1,296 @@
...
+
+  // SAPPHIRE //////////////////////////////////////////////////
+
+  G4String sapphire_name = "SAPPHIRE";
+
+  G4Material* sapphire_mat = materials::Sapphire();
+
+  sapphire_mat->SetMaterialPropertiesTable(opticalprops::Sapphire());
+
+  //G4double pedot_thickn  = 1. *um;
```
paolafer: Remove this line if it's not needed.

source/geometries/BlackBox.cc
```diff
@@ -0,0 +1,296 @@
...
+                     sapphire_logic_vol, sapphire_name, box_logic_vol,
+                     false, 0, false);
+
+  // PEDOT ////////////////////////////////////////////////////////
+
+  G4String pedot_name = "PEDOT";
+
+  G4Material* pedot_mat = materials::TPB();
+
+  sapphire_mat->SetMaterialPropertiesTable(opticalprops::Sapphire());
```
paolafer: Why is PEDOT made of TPB and given the same properties as sapphire? Is it on purpose? Why?

source/geometries/BlackBox.cc
```diff
@@ -0,0 +1,296 @@
...
+
+  // BLACK BOX /////////////////////////////////////////////////
+
+  G4String box_name = "BLACK BOX";
+
+  G4Material* box_mat = G4NistManager::Instance()->FindOrBuildMaterial("G4_AIR");
+
+  box_mat->SetMaterialPropertiesTable(opticalprops::Vacuum());
+
+  G4Box* box_solid_vol =
```
paolafer, line 137: I don't understand why you are creating a volume of air inside another volume of air. What's the purpose of it?

source/geometries/BlackBox.cc
```diff
@@ -0,0 +1,296 @@
...
+  G4LogicalVolume* dice_board_logic = dice_->GetLogicalVolume();
+
+  ////Dice Board placement
+  G4ThreeVector post(dice_board_x_pos_,dice_board_y_pos_,dice_board_z_pos_);
+  G4RotationMatrix* rot = new G4RotationMatrix();
+  rot -> rotateY(rotation_);
+  new G4PVPlacement(rot, post, dice_board_logic,
+            	      dice_board_logic->GetName(), box_logic_vol, false, 1, false);
+
+  // SAPPHIRE //////////////////////////////////////////////////
```
paolafer, line 165: What's the purpose of a sapphire window and a PEDOT (or TPB) layer in a box made of air?

source/geometries/BlackBox.h
```diff
@@ -0,0 +1,77 @@
...
+    BlackBoxSiPMBoard* dice_;
+    G4ThreeVector kdb_dimensions_;
+
+    G4bool visibility_;
+
+    //Messenger for configuration parameters
+    G4GenericMessenger* msg_;
+
+    G4ThreeVector specific_vertex_;
+    //G4double specific_vertex_X_;
```
paolafer: Remove these lines.

source/geometries/BlackBoxSiPMBoard.cc
```diff
@@ -0,0 +1,390 @@
+// -----------------------------------------------------------------------------
+// nexus | BlackBoxSiPMBoard.cc
+//
+// SiPM Board used in BlackBox.
+// -----------------------------------------------------------------------------
```
paolafer, line 5: Use the same convention as the rest of the code in the class descriptions.

source/geometries/BlackBoxSiPMBoard.cc
```diff
@@ -0,0 +1,390 @@
...
+
+  G4Material* mother_gas = mother_phys_->GetLogicalVolume()->GetMaterial();
+
+
+  /// Board configuration checks
+  // Coating require membranes
+  if (coating_thickn_ > 0.)
+    if (membrane_thickn_ == 0.)
+      G4Exception("[BlackBoxSiPMBoard]", "Construct()", FatalException,
+      "Coating require membranes");
```
paolafer: typo: require --> requires.

source/geometries/BlackBoxSiPMBoard.cc
```diff
@@ -0,0 +1,390 @@
...
+  G4String kapton_name = "KAPTON_BOARD";
+
+  G4double kapton_posz = - board_size_z/2. + kapton_thickn_/2.;
+
+  G4Box* kapton_solid = new G4Box(kapton_name, board_size_x/2.,
+                                  board_size_y/2., kapton_thickn_/2.);
+
+  G4LogicalVolume* kapton_logic =
+    new G4LogicalVolume(kapton_solid,
+                        G4NistManager::Instance()->FindOrBuildMaterial("G4_KAPTON"),
```
paolafer: We have seen that G4 has a bug that causes the jobs to crash sometimes when an optical photon enters a region with no material properties defined. Maybe you should add an empty instance of material properties to kapton here and to the rest of materials as well.

source/geometries/BlackBoxSiPMBoard.cc
```diff
@@ -0,0 +1,390 @@
...
+  //G4int central_sipms[] = {27, 28, 35, 36};
+
+  //for (G4int sipm_id:central_sipms) {
+  for (G4int sipm_id=0; sipm_id<num_sipms_; sipm_id++) {
+    //if (sipm_id==27 || sipm_id==28 || sipm_id==35 || sipm_id==36) {
+      //  continue;
+    //}
+       sipm_pos = sipm_positions_[sipm_id] + G4ThreeVector(0., 0., sipm_posz);
+       G4cout << "SiPM" << sipm_id << ":" << sipm_pos << G4endl;
+       //new G4PVPlacement(nullptr, sipm_pos, sipm_->GetLogicalVolume(),
```
paolafer: Unless there's a specific reason, remove all commented lines along the rest of the code.

source/geometries/BlackBox.cc
```diff
@@ -0,0 +1,296 @@
...
+
+  }
+
+    G4ThreeVector BlackBox::GenerateVertex(const G4String& region) const
+  {
+    G4ThreeVector vertex(0.,0.,0.);
+
+    // WORLD
+    if (region == "WORLD") {
+      vertex = G4ThreeVector(0.,0.,0.*mm);
```
paolafer: There's no need to multiply by the unit of measure if the value is zero. However, if you want to do it, you should do it with the three coordinates.

source/geometries/BlackBox.cc
```diff
@@ -0,0 +1,296 @@
...
+
+
+  }
+
+    G4ThreeVector BlackBox::GenerateVertex(const G4String& region) const
+  {
+    G4ThreeVector vertex(0.,0.,0.);
+
+    // WORLD
+    if (region == "WORLD") {
```
paolafer: It seems to me that this vertex is in the centre of the geometry. I suggest to call the region `CENTER`, instead of `WORLD`, which seems misleading to me.

source/geometries/BlackBox.cc
paolafer: Also, if you are returning the default value of the vertex, you don't need to redefine it. You can just do `return vertex`, right?

source/geometries/BlackBox.cc
```diff
@@ -0,0 +1,296 @@
...
+    else if (region == "AD_HOC") {
+      // AD_HOC does not need to be shifted because it is passed by the user
+      return specific_vertex_;
+    }
+    else {
+      G4Exception("[BlackBox]", "GenerateVertex()", FatalException,
+		  "Unknown vertex generation region!");
+    }
+
+    G4ThreeVector displacement = G4ThreeVector(0., 0., 0.);
```
paolafer: What's the purpose of this displacement by zero?

source/geometries/BlackBox.cc
```diff
@@ -0,0 +1,296 @@
...
+    G4ThreeVector BlackBox::GenerateVertex(const G4String& region) const
+  {
+    G4ThreeVector vertex(0.,0.,0.);
+
+    // WORLD
+    if (region == "WORLD") {
+      vertex = G4ThreeVector(0.,0.,0.*mm);
+    }
+    else if (region == "AD_HOC") {
+      // AD_HOC does not need to be shifted because it is passed by the user
```
paolafer: This comment doesn't belong to this class, I think. Why should the vertex be shifted?

source/geometries/SiPMSensl.cc
```diff
@@ -87,7 +87,8 @@ namespace nexus {
     G4double support_pos_z = - sipm_z/2. + support_z/2. + 0.21 * mm;
 
 
-    G4Material* plastic = G4NistManager::Instance()->FindOrBuildMaterial("G4_POLYCARBONATE");
+    //G4Material* plastic = G4NistManager::Instance()->FindOrBuildMaterial("G4_POLYCARBONATE");
+    G4Material* plastic = G4NistManager::Instance()->FindOrBuildMaterial("G4_AIR");
```
paolafer: Why are you changing the material of the SiPM package? This geometry is used in other detectors of the NEXT simulation, I don't think we should change it now, unless there is a good reason.

source/generators/LED450LWcone.cc
```diff
@@ -0,0 +1,160 @@
+// ----------------------------------------------------------------------------
+// nexus | LED450LWcone.cc
+//
+// This class is the primary generator for events consisting of
```
paolafer: Please, change the description as appropriate for this class.

source/generators/LED450LW.cc
```diff
@@ -0,0 +1,151 @@
+// ----------------------------------------------------------------------------
+// nexus | LED450LW.cc
+//
+// This class is the primary generator for events consisting of
```
paolafer: Is this class a copy of `SingleParticleGenerator` or is there any difference that I'm not spotting?

source/generators/LED450LWcone.cc
```diff
@@ -0,0 +1,160 @@
...
+  G4ThreeVector dir; // it will be set in the if branches below
+  if (fixed_momentum) { // if the user provides a momentum direction
+    dir = momentum_.unit();
+  } else if (restrict_angle) { // if the user provides a range of angles
+    dir = RandomDirectionInRange(costheta_min_, costheta_max_, phi_min_, phi_max_);
+  }
+
+  G4ThreeVector p_dir = G4LambertianRand(dir);
+
+  G4double coseno = -1*p_dir.dot(momentum_);
```
paolafer: We tend to use English for variable names, because often they are useful to understand the code. Please, stick to that.

source/generators/LED450LWcone.cc
```diff
@@ -0,0 +1,160 @@
...
+  bool restrict_angle = costheta_min_ != -1. || costheta_max_ != 1. || phi_min_ != 0. || phi_max_ !=2.*pi;
+
+  G4ThreeVector dir; // it will be set in the if branches below
+  if (fixed_momentum) { // if the user provides a momentum direction
+    dir = momentum_.unit();
+  } else if (restrict_angle) { // if the user provides a range of angles
+    dir = RandomDirectionInRange(costheta_min_, costheta_max_, phi_min_, phi_max_);
+  }
+
+  G4ThreeVector p_dir = G4LambertianRand(dir);
```
paolafer: The direction of the LED light should be lambertian around a fixed direction, which depends on the orientation of the LED. You could pass the specific direction through a configuration parameter.

macros/test.init.mac
MiryamMV: This wasn't intended to be in the PR. I'll remove it.

source/geometries/BlackBox.cc
MiryamMV: This must be a residue of another class. I'll delete it inmediately.

source/geometries/BlackBox.cc
MiryamMV: Done!

source/geometries/BlackBox.cc
MiryamMV: Done!

source/geometries/BlackBox.cc
MiryamMV: Done!

source/geometries/BlackBox.cc
MiryamMV: Done!

source/geometries/BlackBox.cc
MiryamMV: I just noticed the branch wasn't updated and it wasn't the last eversion of the code. Change this to the correct version.

macros/test.init.mac
MiryamMV: Deleted

source/geometries/BlackBox.cc
MiryamMV: This was implemented for the latest black box measurements in the lab (testing the reflectivity of the sapphire window with and without coating)

source/geometries/BlackBox.cc
MiryamMV: Done

source/geometries/BlackBox.cc
MiryamMV: Corrected

source/geometries/BlackBox.h
MiryamMV: Done

source/geometries/BlackBoxSiPMBoard.cc
MiryamMV: Done

source/geometries/BlackBox.cc
MiryamMV: Is it better to delete the box volume and rename the world volume to box?

paolafer: Since you don't simulate the walls of the box, the size doesn't matter, right? So you can just skip creating the `BOX` volume. I understand that, for the purpose of the simulation, the name of the volume is irrelevant, is that correct?

paolafer: Now that I think about it, maybe the name of the class is misleading 😅.

source/geometries/BlackBox.cc
paolafer: I see. Maybe you can make the placement of the window volume configurable, so you can use the code for different purposes.

source/geometries/BlackBox.cc
MiryamMV: If I do this how can I solve this situation: `dice_->SetMotherPhysicalVolume(box_phys_vol);`. There's no physical volume for the world volume.

paolafer: I see, so that was the purpose :-). If you need a physical volume, then I guess that what you did is the correct way.

macros/black_box.init.mac
```diff
@@ -0,0 +1,22 @@
...
+#/nexus/RegisterGenerator SingleParticleGenerator
+/nexus/RegisterGenerator LED450LWcone
+/nexus/RegisterPersistencyManager PersistencyManager
+
+/PhysicsList/RegisterPhysics G4OpticalPhysics
+
+### ACTIONS
+
+/nexus/RegisterRunAction DefaultRunAction
+/nexus/RegisterEventAction SaveAllEventAction
```
paolafer: This action doesn't exist anymore, that's why the tests are failing when trying to run this macro.

source/generators/LED450LWcone.cc
MiryamMV: OK, but I think we should still keep `costheta_min_, costheta_max_, phi_min_, phi_max_` in order to make the selection of events in the chosen cone (for memory purposes).

paolafer: Yes, agree!

MiryamMV: I noticed there's also a` LED450LW` class, I wonder if it would be useful to merge both classes into one called `LambertianGenerator`.

paolafer: That was another of my comments: it seems to me that the `LED450LW` class is just a copy of `SingleParticleGenerator`. If that's the case, I would simply eliminate it and modify the  `LED450LWcone` class as we talked.

source/geometries/SiPMSensl.cc
MiryamMV: This was implemented at some point to test the similarity of the MC to the data.  I've changed it to `G4_POLYCARBONATE` again.

source/generators/LED450LWcone.cc
MiryamMV: Done!

source/generators/LED450LWcone.cc
MiryamMV: Done!

source/generators/LED450LWcone.cc
MiryamMV: `LED450LW` removed and `LED450LWcone` renamed to `LambertianGenerator`.

macros/black_box.init.mac
MiryamMV: Removed

source/geometries/BlackBoxSiPMBoard.cc
```diff
@@ -231,87 +208,6 @@ void BlackBoxSiPMBoard::Construct()
     new G4PVPlacement(nullptr, G4ThreeVector(0., 0., mask_posz), mask_logic,
                       mask_name, board_logic, false, 0, true);
 
-    /// Membranes
-    if (membrane_thickn_ > 0. && mask_thickn_ >= sipm_->GetDimensions().z()){
-    //if (membrane_thickn_ > 0.){
```
MiryamMV: Membrane and coating features have been removed from the class since we're not using them in NEXT100 by the moment.

source/generators/LambertianGenerator.cc
```diff
@@ -0,0 +1,150 @@
...
+
+#include "CLHEP/Units/SystemOfUnits.h"
+
+using namespace nexus;
+using namespace CLHEP;
+
+REGISTER_CLASS(LED450LWcone, G4VPrimaryGenerator)
+
+
+LED450LWcone::LED450LWcone():
```
paolafer: You should replace `LED450LWcone` with `LambertianGenerator` everywhere in the class.

source/generators/LambertianGenerator.cc
```diff
@@ -0,0 +1,150 @@
...
+{
+  // Generate uniform random energy in [E_min, E_max]
+  G4double kinetic_energy = nexus::UniformRandomInRange(energy_max_, energy_min_);
+
+  // Calculate cartesian components of momentum
+  G4double mass   = particle_definition_->GetPDGMass();
+  G4double energy = kinetic_energy + mass;
+  G4double pmod = std::sqrt(energy*energy - mass*mass);
+
+  G4ThreeVector dir = momentum_;
```
paolafer: This assignment is not needed, you can just use the `momentum_` variable, right?

source/generators/LambertianGenerator.cc
```diff
@@ -0,0 +1,150 @@
...
+
+  G4ThreeVector p_dir = G4LambertianRand(dir);
+
+  G4double cosine = -1*p_dir.dot(momentum_);
+
+  G4ThreeVector p;
+
+  while((costheta_min_ > cosine) || (cosine > costheta_max_)){
```
paolafer, line 114: From this `while` statement I deduce that the range of `theta` angles that the user must provide is about the angle with the "axis" of the Lambertian distribution and not, for instance, the global reference system. I think that this can originate errors (in other generators the range is on the reference system), therefore you should document it very clearly. Moreover, the `phi` range, also provided by the user, is not used in the code. If you want to leave the possibility of limiting `phi`, as well, you have to include it in the `while` statement.

source/generators/LambertianGenerator.cc
```diff
@@ -0,0 +1,150 @@
...
+  G4double energy = kinetic_energy + mass;
+  G4double pmod = std::sqrt(energy*energy - mass*mass);
+
+  G4ThreeVector dir = momentum_;
+
+  G4ThreeVector p_dir = G4LambertianRand(dir);
+
+  G4double cosine = -1*p_dir.dot(momentum_);
+
+  G4ThreeVector p;
```
paolafer: You don't need to declare this variable here. Just declare it later at line 127, when you also assign it a value.

source/generators/LambertianGenerator.h
```diff
@@ -0,0 +1,71 @@
...
+    /// in the event.
+    void GeneratePrimaryVertex(G4Event*);
+
+  private:
+
+    void SetParticleDefinition(G4String);
+
+    /// Generate a random kinetic energy with flat probability in
+    //  the interval [energy_min, energy_max].
+    G4double RandomEnergy() const;
```
paolafer: This function is not defined neither used, just remove it.

source/geometries/BlackBox.cc
```diff
@@ -0,0 +1,251 @@
...
+    visibility_(0)
+  {
+    msg_ = new G4GenericMessenger(this, "/Geometry/BlackBox/",
+				  "Control commands of BlackBox.");
+    msg_->DeclareProperty("visibility", visibility_, "Black Box visibility");
+
+    msg_->DeclarePropertyWithUnit("specific_vertex", "mm",  specific_vertex_,
+      "Set generation vertex.");
+
+    G4GenericMessenger::Command&  dice_board_x_pos_cmd =
```
paolafer: Why don't you use a unique parameter for the position of dice? It's analogous to `specific_vertex_`.

source/geometries/BlackBox.cc
```diff
@@ -0,0 +1,251 @@
...
+    world_z_ (3. * m),
+    world_xy_ (2. *m),
+    box_z_ (2. * m),
+    box_xy_ (1. *m),
+    specific_vertex_(0,0,0),
+    dice_board_x_pos_(0),
+    dice_board_y_pos_(0),
+    dice_board_z_pos_(-49. *cm),
+    rotation_(0),
+    mask_thickn_(),
```
paolafer: Why do you leave an empty initialization? What happens if the user doesn't provide a value via configuration file?

test.config.mac
```diff
@@ -0,0 +1,32 @@
+## ----------------------------------------------------------------------------
```
paolafer: These two files should also be removed from the PR, right?

source/geometries/BlackBoxSiPMBoard.cc
MiryamMV: You mean a general empty instance for all materials without material properties?

paolafer: Yes, the same way it is done in the rest of the nexus geometries.

source/geometries/BlackBox.cc
```diff
@@ -0,0 +1,251 @@
...
+    hole_diameter_(1),
+    hole_x_(1),
+    hole_y_(1),
+
+    visibility_(0)
```
paolafer: Move the initialization of the `visibility_` variable up, to follow the same order as the declaration in the header file. This way, the warning that appears at compilation time will disappear.

source/geometries/BlackBoxSiPMBoard.cc
MiryamMV: Done!

paolafer: You should do it with every material that does not have already some specific optical properties, not just kapton.

source/geometries/BlackBox.cc
MiryamMV: Done!

source/geometries/BlackBox.cc
MiryamMV: Done

test.config.mac
MiryamMV: Removed!

source/geometries/BlackBox.cc
```diff
@@ -0,0 +1,234 @@
...
+    mask_thickn_(0),
+    hole_diameter_(1),
+    hole_x_(1),
+    hole_y_(1),
+
+    visibility_(0)
+  {
+    msg_ = new G4GenericMessenger(this, "/Geometry/BlackBox/",
+				  "Control commands of BlackBox.");
+    msg_->DeclareProperty("visibility", visibility_, "Black Box visibility");
```
paolafer: This property is not being used anywhere in the code. Either use it, or remove it.

MiryamMV: Deleted

source/generators/LambertianGenerator.cc
MiryamMV: I've attached a picture to make the explanation clearer. 
Also, regarding the negative sign, I'm not sure how to make it more general. I've also deleted the feature of limiting `phi`. I don't find it necessary right  now and I find it difficult to include in a general way.

angles

source/geometries/BlackBox.cc
MiryamMV: Comment removed!

source/generators/LambertianGenerator.cc
paolafer: I'm sorry, but I don't understand the explanation you've written. You are checking that `cosine` is in the defined range of angles and `cosine` is defined as the scalar product of the direction randomly generated and the direction specified by the user as the "axis" of the lambertian distribution. Therefore, you are selecting the angle between the momentum generated randomly and the "axis" of the lambertian distribution, which, in the black box geometry, is (0, 0, -1), but in another geometry could be any direction. Did I get it wrong? From your comment, it seems that the angle is always relative to the global reference system.

paolafer: Regarding the negative sign, it cannot depend on the geometry, because you are building a generic random lambertian generator, which can be used in any geometry. Given a direction, this generator should provide a momentum which follows a Lambertian distribution around that direction, independently of anything else.

source/geometries/BlackBox.cc
paolafer: I think that in the code right now the window and PEDOT are placed in the geometry, with no possibility of choice. Wouldn't it be useful to make these two placements configurable?

source/geometries/BlackBox.cc
```diff
@@ -156,53 +162,59 @@ namespace nexus {
...
-               window_thickn/2., 0, twopi);
+    G4double pedot_thickn  = 150. *nm;
+    G4double sapphire_thickn  = 6. *mm;
+    G4double sapphire_diam = 20 *cm;
+    G4double window_thickn  = sapphire_thickn;
 
-  G4LogicalVolume* sapphire_logic_vol =
-    new G4LogicalVolume(sapphire_solid_vol, sapphire_mat, sapphire_name);
+    if (pedot_coating_) {
+    G4double window_thickn  = sapphire_thickn + pedot_thickn;}
```
paolafer: You are declaring this variable twice, and the second time it lives only in the local scope of `if (pedot_coating_)`, therefore its value is not changed in the rest of the code. Please, fix it.

source/geometries/BlackBox.cc
```diff
@@ -156,53 +162,59 @@ namespace nexus {
...
 
-  new G4PVPlacement(0, pedot_pos,
-                    pedot_logic_vol, pedot_name, sapphire_logic_vol,
-                    false, 0, false);
+        G4ThreeVector pedot_pos(0,0,window_thickn/2.- pedot_thickn/2.);
+
+        new G4PVPlacement(0, pedot_pos,
+                          pedot_logic_vol, pedot_name, sapphire_logic_vol,
+                          false, 0, false);
+  }
```
paolafer: Try to indent these brackets in line with the open brackets.

source/geometries/BlackBox.cc
```diff
@@ -57,6 +57,12 @@ namespace nexus {
     msg_ = new G4GenericMessenger(this, "/Geometry/BlackBox/",
 				  "Control commands of BlackBox.");
 
+    // Boolean-type properties (true or false)
```
paolafer, line 62: Add a default value to these variables in the initialization list.

macros/black_box.config.mac
```diff
@@ -3,14 +3,18 @@
 ### board in the black box.
 ### -----------------------------------------------------------
 
-/Generator/LambertianGenerator/particle opticalphoton
+#/Generator/LambertianGenerator/particle opticalphoton
```
paolafer: I would leave `opticalphoton` as a particle in the example, since it is how we use this macro, most of the times.

paolafer (changes requested):

source/geometries/BlackBox.cc
```diff
@@ -0,0 +1,238 @@
...
+    G4String sapphire_name = "SAPPHIRE";
+
+    G4Material* sapphire_mat = materials::Sapphire();
+
+    sapphire_mat->SetMaterialPropertiesTable(opticalprops::Sapphire());
+
+    G4double pedot_thickn  = 150. *nm;
+    G4double sapphire_thickn  = 6. *mm;
+    G4double sapphire_diam = 20 *cm;
+    G4double window_thickn  = sapphire_thickn;
```
paolafer: I think that defining `sapphire_thickn`  is not necessary. You can simply do 

`G4double window_thickn = 6. * mm;`

and then modify it if PEDOT is added.

macros/black_box.config.mac
```diff
@@ -3,14 +3,18 @@
...
 /Generator/LambertianGenerator/min_energy 2.9 eV
 /Generator/LambertianGenerator/max_energy 2.9 eV
 /Generator/LambertianGenerator/region AD_HOC
 /Generator/LambertianGenerator/momentum 0 0 -1
 /Generator/LambertianGenerator/min_costheta 0.9935
 /Generator/LambertianGenerator/max_costheta 1.
 
+/Geometry/BlackBox/sapphire false
+/Geometry/BlackBox/pedot_coating false
+/Geometry/BlackBox/dice_board_pos 0 0 -50 cm
```
paolafer: This parameter is defined twice, remove one of the lines.

source/geometries/BlackBox.cc
MiryamMV: Done!

source/geometries/BlackBox.cc
MiryamMV: Done!

source/geometries/BlackBox.cc
MiryamMV: Done!

macros/black_box.config.mac
MiryamMV: Done!

source/geometries/BlackBox.cc
MiryamMV: Done!

macros/black_box.config.mac
MiryamMV: Done!

macros/black_box.config.mac
```diff
@@ -0,0 +1,32 @@
...
+/Geometry/BlackBox/rotation 0 deg
+/Geometry/BlackBox/mask_thickn 0. mm
+/Geometry/BlackBox/hole_diameter 0. mm #if set to 0 the holes will be rectangular.
+/Geometry/BlackBox/hole_x 6. mm
+/Geometry/BlackBox/hole_y 6. mm
+
+/run/verbose 0
+/event/verbose 0
+/tracking/verbose 0
+
```
paolafer, line 30: Add `/process/em/verbose 0` to eliminate part of the output. The verbosity may be annoying when running automatic tests, to identify where they fail.

paolafer:
Review the full geometry, including the SiPM board, and set to false the overlap checking. We want to remove as much verbosity as possible in the output.

source/geometries/BlackBoxSiPMBoard.cc
```diff
@@ -0,0 +1,249 @@
...
+  sipm_rot->rotateY(pi);
+  G4double sipm_posz = - board_size_z/2. + kapton_thickn_ + sipm_->GetDimensions().z()/2.;
+  G4ThreeVector sipm_pos;
+
+  //G4int central_sipms[] = {27, 28, 35, 36};
+
+  //for (G4int sipm_id:central_sipms) {
+  for (G4int sipm_id=0; sipm_id<num_sipms_; sipm_id++) {
+       sipm_pos = sipm_positions_[sipm_id] + G4ThreeVector(0., 0., sipm_posz);
+       G4cout << "SiPM" << sipm_id << ":" << sipm_pos << G4endl;
```
paolafer, line 139: Remove this `G4cout` or use it only if verbosity is true (in this case, set it to false in the example macro).

source/geometries/BlackBoxSiPMBoard.cc
```diff
@@ -0,0 +1,249 @@
...
+#include <G4LogicalBorderSurface.hh>
+#include <G4SystemOfUnits.hh>
+
+using namespace nexus;
+
+
+BlackBoxSiPMBoard::BlackBoxSiPMBoard():
+  GeometryBase     (),
+  verbosity_       (true),
+  sipm_verbosity_  (true),
```
MiryamMV: I think I should remove this variable as it is not used anywhere.

paolafer (approved):
This PR adds the geometry of the set-up used in the black box to study the response of a board of SiPMs. It also adds a generic Lambertian generator, which can be used for other purposes. Very good job!

PR #115, opened 2021-06-10
Add visibility parameter.

paolafer:
This PR adds a visibility parameter to the mini castle geometry, to make it invisible if not specified otherwise.


halmamol (approved):
This PR adds the visibility parameters for the smaller lead castle placed inside the main one. This visibility option was not included, so I approve this PR.

PR #116, opened 2021-06-23
DEMO Run 10

gondiaz:
This PR add DEMO Run 10 tracking plane configuration. The four modifications compared to Run 9 are the following:
- Internal sides of rectangular holes are TPB coated. A new seteable variable `hole_coated` is used.
- Dummy Xenon membranes are removed (setting their thickness to 0).
- Add opening holes in outer mask coating at SIPM positions.
- Add SiPM coating (still not implemented, waiting for a corresponding PR from @halmamol which adds this functionality in the Next100SiPM class).

### DIFF
--- a/source/geometries/NextDemoSiPMBoard.cc
+++ b/source/geometries/NextDemoSiPMBoard.cc
@@ -312,6 +312,10 @@ void NextDemoSiPMBoard::Construct()
                                coating_phys, coating_opsurf);
 
     // Coating holes
+    // If there are no membranes, the mask holes are open to the xe gas, namely the hole mask openings
+    // are not covered and exposed to the gas. There is no membrane to support the coating, therefore
+    // we have to open the corresponding holes in the coating volume at the hole (sipm) positions. This is done
+    // through a xe volume with hole dimensions placed inside the coating at the hole (sipm) positions.
     if (membrane_thickn_ == 0.){
       if (hole_type_ == "rectangular"){
 
--- a/source/geometries/NextDemoSiPMBoard.cc
+++ b/source/geometries/NextDemoSiPMBoard.cc
@@ -58,7 +58,7 @@ NextDemoSiPMBoard::NextDemoSiPMBoard():
   hole_diam_       (3.5  * mm),
   hole_x_          (0.0  * mm),
   hole_y_          (0.0  * mm),
-  hole_coated_     (true),
+  hole_coated_     (false),
   sipm_type_       (""),
   mother_phys_     (nullptr),
   kapton_gen_      (nullptr)
@@ -132,12 +132,12 @@ void NextDemoSiPMBoard::Construct()
   G4Material* mother_gas = mother_phys_->GetLogicalVolume()->GetMaterial();
 
 
-  // /// Board configuration checks
-  // // Coating require membranes
-  // if (coating_thickn_ > 0.)
-  //   if (membrane_thickn_ == 0.)
-  //     G4Exception("[NextDemoSiPMBoard]", "Construct()", FatalException,
-  //     "Coating require membranes");
+  /// Board configuration checks
+  // Coating require membranes
+  if ((!hole_coated_) && (coating_thickn_ > 0.))
+    if (membrane_thickn_ == 0.)
+      G4Exception("[NextDemoSiPMBoard]", "Construct()", FatalException,
+      "Coating require membranes");
 
   // Membranes require masks
   if (membrane_thickn_ > 0.)
--- a/source/geometries/NextDemoSiPMBoard.cc
+++ b/source/geometries/NextDemoSiPMBoard.cc
@@ -31,6 +31,7 @@
 #include <G4OpticalSurface.hh>
 #include <G4LogicalSkinSurface.hh>
 #include <G4LogicalBorderSurface.hh>
+#include <G4UnionSolid.hh>
 
 using namespace nexus;
 
@@ -56,6 +57,7 @@ NextDemoSiPMBoard::NextDemoSiPMBoard():
   hole_diam_       (3.5  * mm),
   hole_x_          (0.0  * mm),
   hole_y_          (0.0  * mm),
+  hole_coated_     (true),
   sipm_type_       (""),
   mother_phys_     (nullptr),
   kapton_gen_      (nullptr)
@@ -129,12 +131,12 @@ void NextDemoSiPMBoard::Construct()
   G4Material* mother_gas = mother_phys_->GetLogicalVolume()->GetMaterial();
 
 
-  /// Board configuration checks
-  // Coating require membranes
-  if (coating_thickn_ > 0.)
-    if (membrane_thickn_ == 0.)
-      G4Exception("[NextDemoSiPMBoard]", "Construct()", FatalException,
-      "Coating require membranes");
+  // /// Board configuration checks
+  // // Coating require membranes
+  // if (coating_thickn_ > 0.)
+  //   if (membrane_thickn_ == 0.)
+  //     G4Exception("[NextDemoSiPMBoard]", "Construct()", FatalException,
+  //     "Coating require membranes");
 
   // Membranes require masks
   if (membrane_thickn_ > 0.)
@@ -208,6 +210,13 @@ void NextDemoSiPMBoard::Construct()
 
 
   /// Mask Holes
+  G4String coating_name = "BOARD_COATING";
+  G4Material* tpb = MaterialsList::TPB();
+  tpb->SetMaterialPropertiesTable(OpticalMaterialProperties::TPB());
+  G4OpticalSurface* coating_opsurf =
+    new G4OpticalSurface(coating_name + "_OPSURF", glisur, ground,
+                         dielectric_dielectric, .01);
+
   G4VSolid* hole_solid = nullptr;
   G4String hole_name = "BOARD_MASK_HOLE";
   if (hole_type_ == "rounded"){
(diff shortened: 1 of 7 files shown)


source/geometries/NextDemoSiPMBoard.cc
```diff
@@ -207,25 +209,59 @@ void NextDemoSiPMBoard::Construct()
...
+  G4VPhysicalVolume* hole_phys_ = new G4PVPlacement(sipm_rot, G4ThreeVector(0., 0., sipm_posz), sipm_->GetLogicalVolume(),
+                                                    sipm_->GetLogicalVolume()->GetName(), hole_logic,
+                                                    false, 0, false);
+
+  if (hole_coated && (hole_type_ == "rectangular")){
+
+    G4Box* hole_coating_top = new G4Box("HOLE_COATING", hole_x_/2., coating_thickn_/2., mask_thickn_/2.);
+    G4Box* hole_coating_lat = new G4Box("HOLE_COATING", coating_thickn_/2., hole_y_/2., mask_thickn_/2.);
+
+    G4UnionSolid* hole_coating = new G4UnionSolid("HOLE_COATING", hole_coating_top, hole_coating_lat, 0,
```
halmamol, line 245: I've thinking about it, I'm not fully sure if you're properly locating each of the sides at the proper location with respect to the others. What if instead of using `G4UnionSolid` to construct the volume, we use `G4SubtractionSolid`? We could define 2 boxes: one the size of the hole, and the other corresponding to the free volume after adding the coating. At the end you would only have the walls around the hole that correspond to the coated area. It would be much simpler.

source/geometries/NextDemoSiPMBoard.cc
```diff
@@ -127,12 +129,12 @@ void NextDemoSiPMBoard::Construct()
   G4Material* mother_gas = mother_phys_->GetLogicalVolume()->GetMaterial();
 
 
-  /// Board configuration checks
-  // Coating require membranes
-  if (coating_thickn_ > 0.)
-    if (membrane_thickn_ == 0.)
-      G4Exception("[NextDemoSiPMBoard]", "Construct()", FatalException,
-      "Coating require membranes");
+  // /// Board configuration checks
```
halmamol: Is there any reason of leaving these lines commented?

source/geometries/NextDemoSiPMBoard.cc
gondiaz: I agree that your solution would also be valid and probably easier to implement, and also the proper way to implement the coating for rounded holes. Nevertheless you can check that the sides are properly placed. In the figure below I show the coating placed inside the hole for a coating thickness of 1 mm (to improve visualization).

coated sipm hole

source/geometries/NextDemoSiPMBoard.cc
gondiaz: I had to comment the lines to make the code compatible with Run 10. Since coating is placed over the membranes in Run 7, Run 8 and Run 9, this exception might be useful. I commented out this lines just to remember this fact.

In my opinion, since the membranes we are simulating are dummy ones (just a Xenon volume with no effect in the simulation) I don't even think they are useful in the geometry. But that is something @jmunozv might want to clarify.

halmamol: I see. If need to keep them (depending on what experts say), we could add lines to do this check depending on the run configuration used. I guess this will mean to add a config variable.

source/geometries/NextDemoSiPMBoard.cc
halmamol: Cool. I still think it would be more efficient and simple with the `G4SubtractionSolid`, it would be just two `G4Box` one inside the other. I'm not sure if you would have any issue in the joint areas.

gondiaz: What issue? It is a `G4UnionSolid` thus the overlapping is managed. There is no geometry overlap either.

source/geometries/NextDemoSiPMBoard.cc
```diff
@@ -55,6 +56,7 @@ NextDemoSiPMBoard::NextDemoSiPMBoard():
   hole_diam_       (3.5  * mm),
   hole_x_          (0.0  * mm),
   hole_y_          (0.0  * mm),
+  hole_coated      (true),
```
halmamol: By convention this variable needs underscore.

source/geometries/NextDemoSiPMBoard.cc
```diff
@@ -259,32 +293,40 @@ void NextDemoSiPMBoard::Construct()
...
     new G4LogicalBorderSurface("TEFLON_WLS_GAS_OPSURF", coating_phys,
                                mother_phys_, coating_opsurf);
     new G4LogicalBorderSurface("GAS_TEFLON_WLS_OPSURF", mother_phys_,
                                coating_phys, coating_opsurf);
 
+    // Coating holes
+    if (membrane_thickn_ == 0.){
+      if (hole_type_ == "rectangular"){
+
+        G4Box* board_coating_hole_solid = new G4Box("BOARD_COATING_HOLE", hole_x_/2., hole_y_/2., coating_thickn_/2.);
```
halmamol, line 327: I found a little bit missleading the name **BOARD_COATING_HOLE**. I would add some lines specifying the purpouse of this section.

halmamol:
I've realised that the `VERBOSITY` section of `NextDemoSiPMBoard.cc` it is not updated with the use of Next100SiPMs and rectangle mask holes. We could add these changes in this PR.

source/geometries/NextDemoSiPMBoard.cc
```diff
@@ -306,9 +352,18 @@ void NextDemoSiPMBoard::Construct()
 
     G4cout << "* Kapton thickness:   " << kapton_thickn_   << G4endl;
     G4cout << "* Mask thickness:     " << mask_thickn_     << G4endl;
-    G4cout << "* Mask hole diameter: " << hole_diam_       << G4endl;
     G4cout << "* Membrane thickness: " << membrane_thickn_ << G4endl;
-    G4cout << "* Coating thickness: "  << coating_thickn_  << G4endl;
+    G4cout << "* Coating thickness:  " << coating_thickn_  << G4endl;
+    G4cout << "* SiPM type:          " << sipm_type_       << G4endl;
```
halmamol, line 362: Maybe this line can be moved to line 346, or we could directly erase that one.

paolafer:
The description if the `NextDemoTrackingPlane` class needs to be updated with the new configurations.

paolafer:
The same applies to `NextDemoSiPMBoard`.

source/geometries/NextDemoSiPMBoard.cc
paolafer: To move here the discussion in Slack: the best way to do it is putting a `G4Box` inside another one, no subtractions neither unions involved, which may slow down the tracking. The bigger box will be made of TPB and the smaller one of the gas of the mother volume. Also, if we ever need to do the same with the rounded holes, it can be done the same way.

gondiaz: Since the SiPM resides in the hole, the two `G4Box` implementation you suggest would requiere to modify the volume were they are placed, and consequently the previous run configurations. The volumes are placed as intended and inner coating is only added if required, which  would not be the case with two boxes since one of them is made of tpb.

For external reader: we agreed via Slack discussion to maintain this implementation since is the minimal change possible. We would consider the change in the future if needed, for example, to implement rounded hole coatings.

source/geometries/NextDemoSiPMBoard.h
```diff
@@ -1,11 +1,12 @@
 // -----------------------------------------------------------------------------
-// nexus | NextDemoSiPMBoard.h
+// nexus | NextDemoSiPMBoard.cc
```
paolafer: Why this change? It should be `.h`.

source/geometries/NextDemoTrackingPlane.cc
```diff
@@ -2,9 +2,10 @@
 // nexus | NextDemoTrackingPlane.cc
 //
 // Tracking plane of the Demo++ geometry.
-// It implements 2 different TP implementations:
+// It describes 2 different TP implementations:
```
paolafer: The implementations are 3, I think.

paolafer:
Now we can add the `run10` configuration in the tests.

source/geometries/NextDemoSiPMBoard.h
gondiaz: Rigth, I just copied the description in the .cc. Amended.

source/geometries/NextDemoTrackingPlane.cc
gondiaz: Sure, amended.

source/geometries/NextDemoSiPMBoard.cc
```diff
@@ -208,25 +211,59 @@ void NextDemoSiPMBoard::Construct()
...
 
   // Generate SiPM positions
   GenerateSiPMPositions();
 
   G4double sipm_posz = - mask_thickn_/2. + sipm_z_dim/2.;
 
-  new G4PVPlacement(sipm_rot, G4ThreeVector(0., 0., sipm_posz), sipm_->GetLogicalVolume(),
-                    sipm_->GetLogicalVolume()->GetName(), hole_logic,
-                    false, 0, false);
+  G4VPhysicalVolume* hole_phys_ = new G4PVPlacement(sipm_rot, G4ThreeVector(0., 0., sipm_posz), sipm_->GetLogicalVolume(),
```
paolafer: Remove trailing underscore from the variable name in lines 235, 260 and 261.

paolafer: This placement is that of the SiPM inside the hole, right? If so it should be called `sipm_phys`, otherwise it's misleading.

source/geometries/NextDemoSiPMBoard.cc
```diff
@@ -208,25 +211,59 @@ void NextDemoSiPMBoard::Construct()
...
+    hole_coating = new G4UnionSolid("HOLE_COATING", hole_coating, hole_coating_top, 0,
+                                    G4ThreeVector(0., -hole_y_ + coating_thickn_, 0.));
+
+    G4LogicalVolume* hole_coating_logic = new G4LogicalVolume(hole_coating, tpb, "HOLE_COATING");
+
+    G4VPhysicalVolume* hole_coating_phys_ =
+          new G4PVPlacement(nullptr, G4ThreeVector(0., hole_y_/2. - coating_thickn_/2., 0.), hole_coating_logic,
+                            "HOLE_COATING", hole_logic, false, 0, false);
+
+    new G4LogicalBorderSurface("TEFLON_WLS_HOLE_GAS_OPSURF", hole_coating_phys_,
```
paolafer: If you use `hole_phys_` here, you're setting the surface between the SiPM and the coating, not between the gas of the hole and the coating, because `hole_phys_` is the physical volume of the SiPM in the current version of the code, right?

gondiaz: The sipm volume is smaller than the hole volume. The hole has dimensions of the mask holes and sipms are placed inside. That means that most of the hole is filled with Xe, except the SiPM volume, in the  figure the solid box represents the full SiPM encasing and the coated walls:
SiPM-inside-holes

gondiaz: Right, `hole_phys` should be `sipm_phys`. These `G4LogicalBorderSurface` should be defined for each hole in its positioning at line 289, no?

paolafer: I would say so. However, I'm looking at our own documentation of the optical surfaces (https://github.com/next-exp/nexus/wiki/Setting-optical-properties) and there's a caveat regarding the case when one of the volumes is the daughter of the other one. I think we should try and find a time to talk to @jmunozv to clarify this.

source/geometries/NextDemoSiPMBoard.cc
```diff
@@ -250,42 +281,62 @@ void NextDemoSiPMBoard::Construct()
 
 
   /// Placing the Holes with SiPMs & membranes inside
+  G4VPhysicalVolume* hole_phys[num_sipms_];
```
paolafer: It looks like this gives a compilation warning: `variable length arrays are a C99 feature`.
You could use an `std::vector` instead.

paolafer: After talking to @jmalbos, it seems that there is no need to define any container. You can simply assign the same pointer to the placement in each iteration of the loop.

source/geometries/NextDemoSiPMBoard.cc
halmamol: What are we going to do with these commented lines? Are we going to keep them like that or adapt them to the case without membrane? Maybe we could include `hole_coated_` in the condition, like `if (!hole_coated_ && (membrane_thickn_ == 0.) )`

halmamol:
This PR adds Run 10 in DEMO configuration. This run removes the membranes, and includes the possibility to coat with TPB the holes in the mask. In each of these holes, optical surfaces have been assigned to coating. It also sets the coating of the SiPMs from the Next100SiPM class. Tests have been included accordingly for Run 10. I approve this PR, but a final blessing would be expected from more experienced reviewers (@paolafer @jmalbos @jmunozv)

halmamol (approved):
This PR adds Run 10 in DEMO configuration. This run removes the membranes, and includes the possibility to coat with TPB the holes in the mask. In each of these holes, optical surfaces have been assigned to coating. It also sets the coating of the SiPMs from the Next100SiPM class. Tests have been included accordingly for Run 10. Future studies about the optical properties of daughter volumes would be required in order to improve the implementation of the coating of the holes. I approve this PR, but I would wait for a final blessing from more experienced reviewers (@paolafer @jmalbos @jmunozv)

PR #117, opened 2021-06-24
Add tpb coating into Next100SiPM 

halmamol:
This PR adds the possibiity to coat the Next100SiPMs with a TPB layer. The coating is added on top of the SiPMs windows, with a 1micrometer thickness. DEMO SiPM board and tracking plane geometries have been modified in order to add the coating option when this SiPMs are used.

### DIFF
--- a/source/geometries/Next100SiPM.cc
+++ b/source/geometries/Next100SiPM.cc
@@ -33,7 +33,6 @@ Next100SiPM::Next100SiPM():
   mother_depth_       (0),
   naming_order_       (0),
   time_binning_       (1.0 * us),
-  with_coating_       (false),
   coating_thickn_     (2. * micrometer),
   visibility_         (true)
 {
@@ -61,7 +60,6 @@ void Next100SiPM::Construct()
   G4double sipm_length = 4.0 * mm;
   G4double sipm_thickn = 1.3 * mm;
 
-  if(!with_coating_) coating_thickn_ = 0.;
   sipm_thickn = sipm_thickn + coating_thickn_;
 
   dimensions_.setX(sipm_width);
@@ -77,11 +75,8 @@ void Next100SiPM::Construct()
   GeometryBase::SetLogicalVolume(sipm_logic_vol);
 
   // COATING /////////////////////////////////////////////
-
-  G4LogicalVolume* coating_logic_vol;
-  if (with_coating_) {
+  if (coating_thickn_ > 0.) {
     G4String coating_name = sipm_name + "_WLS";
-
 
     G4Box* coating_solid_vol =
       new G4Box(coating_name, sipm_width/2., sipm_length/2., coating_thickn_/2.);
@@ -89,7 +84,7 @@ void Next100SiPM::Construct()
     G4Material* coating_mt = MaterialsList::TPB();
     coating_mt->SetMaterialPropertiesTable(OpticalMaterialProperties::TPB());
 
-    coating_logic_vol = new G4LogicalVolume(coating_solid_vol, coating_mt, coating_name);
+    G4LogicalVolume* coating_logic_vol = new G4LogicalVolume(coating_solid_vol, coating_mt, coating_name);
 
     G4double coating_zpos = sipm_thickn/2. - coating_thickn_/2;
 
--- a/source/geometries/Next100SiPM.cc
+++ b/source/geometries/Next100SiPM.cc
@@ -34,6 +34,7 @@ Next100SiPM::Next100SiPM():
   naming_order_       (0),
   time_binning_       (1.0 * us),
   with_coating_       (false),
+  coating_thickn_     (2. * micrometer), 
   visibility_         (true)
 {
 }
@@ -78,7 +79,6 @@ void Next100SiPM::Construct()
   if (with_coating_) {
     G4String coating_name = sipm_name + "_WLS";
 
-    G4double coating_thickn_  = 1. * micrometer;
 
     G4Box* coating_solid_vol =
       new G4Box(coating_name, sipm_width/2., sipm_length/2., coating_thickn_/2.);
--- a/source/geometries/Next100SiPM.cc
+++ b/source/geometries/Next100SiPM.cc
@@ -61,8 +61,8 @@ void Next100SiPM::Construct()
   G4double sipm_length = 4.0 * mm;
   G4double sipm_thickn = 1.3 * mm;
 
-  if(with_coating_) sipm_thickn = sipm_thickn + coating_thickn_;
-  else coating_thickn_ = 0.;
+  if(!with_coating_) coating_thickn_ = 0.;
+  sipm_thickn = sipm_thickn + coating_thickn_;
 
   dimensions_.setX(sipm_width);
   dimensions_.setY(sipm_length);
@@ -86,10 +86,10 @@ void Next100SiPM::Construct()
     G4Box* coating_solid_vol =
       new G4Box(coating_name, sipm_width/2., sipm_length/2., coating_thickn_/2.);
 
-    G4Material* coating_mt_ = MaterialsList::TPB();
-    coating_mt_->SetMaterialPropertiesTable(OpticalMaterialProperties::TPB());
+    G4Material* coating_mt = MaterialsList::TPB();
+    coating_mt->SetMaterialPropertiesTable(OpticalMaterialProperties::TPB());
 
-    coating_logic_vol = new G4LogicalVolume(coating_solid_vol, coating_mt_, coating_name);
+    coating_logic_vol = new G4LogicalVolume(coating_solid_vol, coating_mt, coating_name);
 
     G4double coating_zpos = sipm_thickn/2. - coating_thickn_/2;
 
--- a/source/geometries/Next100SiPM.cc
+++ b/source/geometries/Next100SiPM.cc
@@ -33,7 +33,8 @@ Next100SiPM::Next100SiPM():
   mother_depth_       (0),
   naming_order_       (0),
   time_binning_       (1.0 * us),
-  visibility_         (true)
+  visibility_         (true),
+  with_coating_       (false)
 {
 }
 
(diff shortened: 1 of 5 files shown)


source/geometries/Next100SiPM.cc
```diff
@@ -71,6 +72,35 @@ void Next100SiPM::Construct()
...
+
+    G4Box* coating_solid_vol =
+      new G4Box(coating_name, sipm_width/2., sipm_length/2., coating_thickn_/2.);
+
+    G4Material* coating_mt_ = MaterialsList::TPB();
+    coating_mt_->SetMaterialPropertiesTable(OpticalMaterialProperties::TPB());
+
+    coating_logic_vol = new G4LogicalVolume(coating_solid_vol, coating_mt_, coating_name);
+
+    G4double coating_zpos = sipm_thickn/2. + coating_thickn_/2.;
```
gondiaz: I have tested the coated sipms with demo Run 9. The coating volume overlaps with its mother sipm volume, you can check it with `/geometry/test/run`. I think that this comes from the fact that the coating volume is placed out of the sipm volume, since the sipm volume has `sipm_thickn` thickness. You should extend this `sipm_thickn` to `sipm_thickn + coating_thickn_`  if  `with_coating=true` in line 62.

halmamol: I see your point, I didn't realise. The thing is the parameter `sipm_thickn` defines the size of the encasing, that includes the sensor area and the window. I've located the coating on top of this encasing. If we extend `sipm_thickn` to `sipm_thickn` + `coating_thickn_`, we're making the encasing larger and we would need to locate the coating below this volumen (on top of the window). I'm not sure if this is realistic or not. Maybe a way to extend the mother sipm volumen (in case it's coated) is by using `sipm_thickn` + `coating_thickn_` instead of `sipm_thickn` in the `dimensions_` definition in line 66.  @paolafer, could you give us a hand here?

paolafer: Sure, I'll call you.

halmamol: I see now your point, @gondiaz. I've increased the `sipm_thickn` in case of coating, and I've corrected the position of the other volumes accordingly.

source/geometries/Next100SiPM.cc
```diff
@@ -71,6 +76,34 @@ void Next100SiPM::Construct()
...
+
+  G4LogicalVolume* coating_logic_vol;
+  if (with_coating_) {
+    G4String coating_name = sipm_name + "_WLS";
+
+
+    G4Box* coating_solid_vol =
+      new G4Box(coating_name, sipm_width/2., sipm_length/2., coating_thickn_/2.);
+
+    G4Material* coating_mt_ = MaterialsList::TPB();
```
paolafer: In nexus, the convention for local variables is not to use trailing underscore. We just use it for global variables.

source/geometries/Next100SiPM.cc
```diff
@@ -52,13 +54,16 @@ G4ThreeVector Next100SiPM::GetDimensions() const
...
   // ENCASING //////////////////////////////////////////////
-
+  //dimensions are increased in case of SiPM coating
   G4String sipm_name = "SIPM_S13372";
 
   G4double sipm_width  = 3.0 * mm;
   G4double sipm_length = 4.0 * mm;
   G4double sipm_thickn = 1.3 * mm;
 
+  if(with_coating_) sipm_thickn = sipm_thickn + coating_thickn_;
```
paolafer: I'm not sure I understand the purpose of this. If you do `if (!with_coating_) coating_thickn_ = 0;`, then you can use 

`sipm_thickn = sipm_thickn + coating_thickn_;`

for both cases. It looks strange to me an `if-else` condition that sets two different variables.

source/geometries/Next100SiPM.cc
jacg: Methinks you mean "the convention is to use trailing underscores for **member** variables":

``` c++
int this_is_a_global_variable = 0; // There's one for the whole program

void some_function() {
    int this_is_a_local_variable = 0; // It's only in scope inside this function
}

class some_class {
    int this_is_a_member_variable = 0; // Every instance of this class gets a separate one
};
```

source/geometries/NextDemoSiPMBoard.h
```diff
@@ -52,6 +54,8 @@ namespace nexus {
     G4bool sipm_verbosity_;
     G4bool visibility_;
     G4bool sipm_visibility_;
+    G4bool sipm_coating_;
+    G4bool sipm_coat_thick_;
```
gondiaz: This is not a boolean!

source/geometries/NextDemoSiPMBoard.h
```diff
@@ -104,6 +108,12 @@ namespace nexus {
   inline void NextDemoSiPMBoard::SetSiPMType(G4String type)
   { sipm_type_ = type; }
 
+  inline void NextDemoSiPMBoard::SetSiPMCoating(G4bool coating)
+  { sipm_coating_ = coating; }
+
+  inline void NextDemoSiPMBoard::SetSiPMCoatThick(G4bool thickn)
```
gondiaz: Same

source/geometries/Next100SiPM.cc
halmamol: I agree, I'll adapt it.

source/geometries/NextDemoSiPMBoard.h
halmamol: Good catch!

gondiaz: This is not solved in the last commit

source/geometries/Next100SiPM.h
```diff
@@ -31,11 +31,13 @@ namespace nexus {
...
-    void SetMotherDepth (G4int mother_depth);
-    void SetNamingOrder (G4int naming_order);
-    void SetTimeBinning (G4double time_binning);
-    void SetVisibility  (G4bool visibility);
+    void SetSensorDepth   (G4int sensor_depth);
+    void SetMotherDepth   (G4int mother_depth);
+    void SetNamingOrder   (G4int naming_order);
+    void SetTimeBinning   (G4double time_binning);
+    void SetSiPMCoating   (G4bool with_coating);
+    void SetSiPMCoatThick (G4bool coating_thickn);
```
gondiaz: Same

source/geometries/Next100SiPM.h
```diff
@@ -61,6 +65,12 @@ namespace nexus {
   inline void Next100SiPM::SetNamingOrder(G4int naming_order)
   { naming_order_ = naming_order; }
 
+  inline void Next100SiPM::SetSiPMCoating(G4bool with_coating)
+  { with_coating_ = with_coating; }
+
+  inline void Next100SiPM::SetSiPMCoatThick(G4bool coating_thickn)
```
gondiaz: Same

source/geometries/NextDemoSiPMBoard.h
halmamol: Something happend while doing the commit, because they're indeed changed in my local version. I'll do another one and push it. Sorry.

source/geometries/Next100SiPM.cc
```diff
@@ -71,6 +76,34 @@ void Next100SiPM::Construct()
 
   GeometryBase::SetLogicalVolume(sipm_logic_vol);
 
+  // COATING /////////////////////////////////////////////
+
+  G4LogicalVolume* coating_logic_vol;
+  if (with_coating_) {
+    G4String coating_name = sipm_name + "_WLS";
+
```
gondiaz, line 80: Remove empty line

source/geometries/Next100SiPM.cc
```diff
@@ -71,6 +76,34 @@ void Next100SiPM::Construct()
 
   GeometryBase::SetLogicalVolume(sipm_logic_vol);
 
+  // COATING /////////////////////////////////////////////
+
+  G4LogicalVolume* coating_logic_vol;
```
gondiaz: Is there any reason to declare this logical volume out of the `if` clause?

source/geometries/Next100SiPM.cc
```diff
@@ -33,6 +33,8 @@ Next100SiPM::Next100SiPM():
   mother_depth_       (0),
   naming_order_       (0),
   time_binning_       (1.0 * us),
+  with_coating_       (false),
+  coating_thickn_     (2. * micrometer),
```
gondiaz, line 36: To me the `with_coating` variable is redundant since `coating_thickn_` could mean no coating

gondiaz (changes requested):

source/geometries/Next100SiPM.cc
halmamol: I did it in the same manner it's declared in GenericPhotosensors. I thought it was a matter of convention.

source/geometries/Next100SiPM.cc
halmamol: I could remove it, and leave the SiPMs coated by default unless `coating_thickn_` is defined as 0.

source/geometries/Next100SiPM.cc
```diff
@@ -71,6 +74,33 @@ void Next100SiPM::Construct()
...
+
+    G4String coating_name = sipm_name + "_WLS";
+
+    G4Box* coating_solid_vol =
+      new G4Box(coating_name, sipm_width/2., sipm_length/2., coating_thickn_/2.);
+
+    G4Material* coating_mt = MaterialsList::TPB();
+    coating_mt->SetMaterialPropertiesTable(OpticalMaterialProperties::TPB());
+
+    coating_logic_vol = new G4LogicalVolume(coating_solid_vol, coating_mt, coating_name);
```
gondiaz: I meant to `G4LogicalVolume* coating_logic_vol = new blah blah`

gondiaz (changes requested):

source/geometries/Next100SiPM.cc
halmamol: Changed it!

gondiaz (approved):
This PR add a layer of TPB in front of Next100SiPM window. The layer thickness is seteable, and it is neglected if its value is zero. It adapts both the Next100SiPM class and Demo TrackingPlane and SiPMBoard to pass this parameter to the SiPMs. I approve this PR, but I would wait opinion from a more experienced revisor @paolafer @jmalbos @jmunozv.

paolafer:
Is the case with coating used in any of the tests? If not, I think we should test it.

paolafer:
The code looks correct to me, I would go ahead with the merging. Good job!

PR #118, opened 2021-06-30
Fieldcage

MiryamMV:
This PR completes the NEXT100 Field Cage according to the up-to-date detector drawings. It adds the copper rings, the ring holders, the cathode ring and other minor changes.

### DIFF
--- a/source/geometries/Next100FieldCage.cc
+++ b/source/geometries/Next100FieldCage.cc
@@ -45,7 +45,6 @@ Next100FieldCage::Next100FieldCage():
   GeometryBase(),
   // Dimensions
   active_diam_ (984 * mm), // distance between the centers of two opposite panels
-  gate_cathode_centre_dist_ (1205. * mm), // distance between gate and the centre of cathode grid
   active_cathode_dist_ (1178.*mm), //distance from the  beginning of the teflon drift panels to the first side of the cathode.
   gate_sapphire_wdw_dist_ (1460.5 * mm), // distance between gate and the surface of sapphire windows
   cathode_int_diam_ (960. * mm),
--- a/source/geometries/Next100FieldCage.cc
+++ b/source/geometries/Next100FieldCage.cc
@@ -80,8 +80,8 @@ Next100FieldCage::Next100FieldCage():
   // EL electric field
   elfield_ (0),
   ELelectric_field_ (34.5*kilovolt/cm),
-  cath_grid_transparency_ (.98), // to check
-  el_grid_transparency_ (.88), // to check
+  cath_grid_transparency_ (.95),
+  el_grid_transparency_ (cath_grid_transparency_), // to check
   max_step_size_ (1. * mm),
   visibility_ (0),
   verbosity_(0),
--- a/source/geometries/Next100FieldCage.cc
+++ b/source/geometries/Next100FieldCage.cc
@@ -70,7 +70,7 @@ Next100FieldCage::Next100FieldCage():
   gate_teflon_dist_ (2.*mm),    //preliminary
   gate_ext_diam_ (1042. * mm), //preliminary
   gate_int_diam_ (1009. * mm), //preliminary
-  gate_thickn_  (15. * mm), // same as anode thickness //preliminary
+  gate_ring_thickn_  (15. * mm), // same as anode thickness //preliminary
   overlap_(0.001*mm), //defined for G4UnionSolids to ensure a common volume within the two joined solids
   // Diffusion constants
   drift_transv_diff_ (1. * mm/sqrt(cm)),
@@ -192,7 +192,7 @@ void Next100FieldCage::Construct()
 {
   /// Calculate lengths of active and buffer regions
   active_length_ = (cathode_thickn_ - grid_thickn_)/2. + teflon_drift_length_ + gate_teflon_dist_
-                 + (gate_thickn_    - grid_thickn_);
+                 + (gate_ring_thickn_    - grid_thickn_);
   buffer_length_ = gate_sapphire_wdw_dist_ - active_length_ - grid_thickn_;
 
   /// Calculate length of teflon in the buffer region
@@ -205,12 +205,12 @@ void Next100FieldCage::Construct()
   gate_grid_zpos_  = GetELzCoord() - grid_thickn_/2.;
   active_zpos_     = gate_grid_zpos_ + grid_thickn_/2. + active_length_/2.;
   cathode_zpos_    = gate_grid_zpos_ + grid_thickn_/2. + active_length_ + grid_thickn_/2.;
-  gate_zpos_       = gate_grid_zpos_ + grid_thickn_/2. + gate_thickn_/2. - grid_thickn_;
+  gate_zpos_       = gate_grid_zpos_ + grid_thickn_/2. + gate_ring_thickn_/2. - grid_thickn_;
   el_gap_zpos_     = gate_grid_zpos_ - grid_thickn_/2. - el_gap_length_/2.;
-  anode_zpos_      = el_gap_zpos_ - el_gap_length_/2. - gate_thickn_/2.;
-  anode_grid_zpos_ = anode_zpos_ + gate_thickn_/2. - grid_thickn_/2.;
+  anode_zpos_      = el_gap_zpos_ - el_gap_length_/2. - gate_ring_thickn_/2.;
+  anode_grid_zpos_ = anode_zpos_ + gate_ring_thickn_/2. - grid_thickn_/2.;
 
-  teflon_drift_zpos_  = gate_grid_zpos_ + grid_thickn_/2. + (gate_thickn_ - grid_thickn_)
+  teflon_drift_zpos_  = gate_grid_zpos_ + grid_thickn_/2. + (gate_ring_thickn_ - grid_thickn_)
                       + gate_teflon_dist_ + teflon_drift_length_/2.;
   teflon_buffer_zpos_ = cathode_zpos_ + cathode_thickn_/2. + teflon_buffer_length_/2.;
 
@@ -264,7 +264,7 @@ void Next100FieldCage::DefineMaterials()
 void Next100FieldCage::BuildActive()
 {
   /// Position of z planes
-  G4double zplane[2] = {-active_length_/2.+(gate_thickn_   -grid_thickn_ + gate_teflon_dist_)-overlap_,
+  G4double zplane[2] = {-active_length_/2.+(gate_ring_thickn_   -grid_thickn_ + gate_teflon_dist_)-overlap_,
                          active_length_/2.-(cathode_thickn_-grid_thickn_)/2.};
   /// Inner radius
   G4double rinner[2] = {0., 0.};
@@ -287,10 +287,10 @@ void Next100FieldCage::BuildActive()
   //This volume is added as an extension of the active volume that reaches the gate grid.
   G4Tubs* active_gate_solid =
     new G4Tubs("ACT_GATE_GAS", 0, gate_int_diam_/2.,
-              (gate_teflon_dist_ + gate_thickn_- grid_thickn_)/2., 0, twopi);
+              (gate_teflon_dist_ + gate_ring_thickn_- grid_thickn_)/2., 0, twopi);
 
   G4ThreeVector act_gate_pos =
-  G4ThreeVector(0., 0., -active_length_/2.+(gate_teflon_dist_ + gate_thickn_- grid_thickn_)/2.);
+  G4ThreeVector(0., 0., -active_length_/2.+(gate_teflon_dist_ + gate_ring_thickn_- grid_thickn_)/2.);
 
   union_active =
     new G4UnionSolid ("ACTIVE", union_active, active_gate_solid, 0, act_gate_pos);
@@ -332,7 +332,13 @@ void Next100FieldCage::BuildActive()
 
 
   /// Visibilities
-  active_logic->SetVisAttributes(G4VisAttributes::Invisible);
+  if (visibility_) {
+    G4VisAttributes active_col = nexus::Yellow();
+    active_logic->SetVisAttributes(active_col);
+  } else {
+    active_logic->SetVisAttributes(G4VisAttributes::Invisible);
+  }
+
 
   /// Verbosity
   if (verbosity_) {
@@ -449,7 +455,12 @@ void Next100FieldCage::BuildBuffer()
                                             G4ThreeVector(0., 0., xenon_zpos));
 
   /// Visibilities
-  buffer_logic->SetVisAttributes(G4VisAttributes::Invisible);
+  if (visibility_) {
+    G4VisAttributes buffer_col = nexus::LightGreen();
+    buffer_logic->SetVisAttributes(buffer_col);
+  } else {
+    buffer_logic->SetVisAttributes(G4VisAttributes::Invisible);
+  }
 
   /// Verbosity
   if (verbosity_) {
@@ -464,7 +475,7 @@ void Next100FieldCage::BuildELRegion()
 {
   /// GATE ring.
   G4Tubs* gate_solid =
-    new G4Tubs("GATE_RING", gate_int_diam_/2., gate_ext_diam_/2., gate_thickn_/2., 0, twopi);
+    new G4Tubs("GATE_RING", gate_int_diam_/2., gate_ext_diam_/2., gate_ring_thickn_/2., 0, twopi);
 
   G4LogicalVolume* gate_logic =
     new G4LogicalVolume(gate_solid, steel_, "GATE_RING");
@@ -486,7 +497,7 @@ void Next100FieldCage::BuildELRegion()
 
   /// ANODE ring.
   G4Tubs* anode_solid =
-    new G4Tubs("ANODE_RING", gate_int_diam_/2., gate_ext_diam_/2., gate_thickn_/2., 0, twopi);
+    new G4Tubs("ANODE_RING", gate_int_diam_/2., gate_ext_diam_/2., gate_ring_thickn_/2., 0, twopi);
 
   G4LogicalVolume* anode_logic =
     new G4LogicalVolume(anode_solid, steel_, "ANODE_RING");
@@ -498,12 +509,12 @@ void Next100FieldCage::BuildELRegion()
   ///Gas under ANODE.
   G4Tubs* anode_gas_solid =
     new G4Tubs("ANODE_GAS", 0, gate_int_diam_/2.,
-              (gate_thickn_-grid_thickn_)/2., 0, twopi);
+              (gate_ring_thickn_-grid_thickn_)/2., 0, twopi);
 
   G4LogicalVolume* anode_gas_logic =
     new G4LogicalVolume(anode_gas_solid, gas_, "ANODE_GAS");
 
-  new G4PVPlacement(0, G4ThreeVector(0., 0., anode_grid_zpos_-gate_thickn_/2.),
+  new G4PVPlacement(0, G4ThreeVector(0., 0., anode_grid_zpos_-gate_ring_thickn_/2.),
                     anode_gas_logic, "ANODE_GAS", mother_logic_,
                     false, 0, false);
 
@@ -724,7 +735,7 @@ void Next100FieldCage::BuildFieldCage()
   G4int    num_drift_rings = 48;
   G4int    num_buffer_rings = 4;
   G4double posz;
-  G4double first_ring_drift_z_pos = gate_zpos_ + gate_thickn_/2. + gate_teflon_dist_ +
+  G4double first_ring_drift_z_pos = gate_zpos_ + gate_ring_thickn_/2. + gate_teflon_dist_ +
                                     drift_ring_dist_/2. + active_short_z/2.;
 
   G4double first_ring_buff_z_pos = first_ring_drift_z_pos + (num_drift_rings-1)*drift_ring_dist_ +
@@ -735,8 +746,6 @@ void Next100FieldCage::BuildFieldCage()
 
   G4LogicalVolume* ring_logic =
     new G4LogicalVolume(ring_solid, copper_, "RING");
-
-
 
   //Placement of the drift rings.
   for (G4int i=0; i<num_drift_rings; i++) {
@@ -834,9 +843,10 @@ void Next100FieldCage::BuildFieldCage()
     // CATHODE holders.
     G4double cathode_long_y = 29.*mm;
     G4double cathode_long_z = 61*mm;
+    G4double cathode_short_z = 24.5*mm;
     G4Box* cathode_large_solid =
       new G4Box("CATHODE_LARGE", holder_x_/2., cathode_long_y/2., cathode_long_z/2.);
-    G4double cathode_short_z = 24.5*mm;
+
     G4Box* cathode_short_solid =
       new G4Box("CATHODE_SHORT", holder_x_/2., holder_short_y_/2., cathode_short_z/2.);
 
--- a/source/geometries/Next100FieldCage.cc
+++ b/source/geometries/Next100FieldCage.cc
@@ -207,7 +207,7 @@ void Next100FieldCage::Construct()
   BuildCathodeGrid();
   BuildBuffer();
   BuildELRegion();
-  BuildFieldCage();
+  BuildLightTube();
 
   /// Calculate EL table vertices
   G4double z = el_gap_zpos_ + el_gap_length_/2.;
@@ -487,7 +487,7 @@ void Next100FieldCage::BuildELRegion()
 }
 
 
-void Next100FieldCage::BuildFieldCage()
+void Next100FieldCage::BuildLightTube()
 {
   /// DRIFT PART ///
   G4double teflon_drift_length = gate_cathode_centre_dist_ - gate_teflon_dist_ - cathode_gap_/2.;
--- a/source/geometries/Next100FieldCage.cc
+++ b/source/geometries/Next100FieldCage.cc
@@ -651,14 +651,14 @@ void Next100FieldCage::BuildFieldCage()/////////////////////////////////////////
                                      "HDPE_TUBE", mother_logic_, false, 0, true);
 
   G4double active_short_z = 13.5*mm; //Thickness for the small piece from the  holder in front of the ring.
-  G4double first_ring_drif_z_pos = drift_ring_dist_/2. + active_short_z/2. + gate_teflon_dist_ + GetELzCoord();
+  G4double first_ring_drift_z_pos = drift_ring_dist_/2. + active_short_z/2. + gate_teflon_dist_ + GetELzCoord();
   G4double buffer_short_z = 37.*mm;
 
   G4double ring_drift_buffer_dist = 72.*mm;
   G4int num_drift_rings = 48;
   G4int num_buffer_rings = 4;
   G4double posz;
-  G4double first_ring_buff_z_pos = first_ring_drif_z_pos + (num_drift_rings-1)*drift_ring_dist_ + ring_drift_buffer_dist;
+  G4double first_ring_buff_z_pos = first_ring_drift_z_pos + (num_drift_rings-1)*drift_ring_dist_ + ring_drift_buffer_dist;
 
   G4Tubs* ring_solid =
   new G4Tubs("RING", ring_int_diam_/2., ring_ext_diam_/2., ring_thickn_/2., 0, twopi);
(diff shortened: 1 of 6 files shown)


MiryamMV:
Here is a document with all the measurements provided by Sara . (Some may change in the code)
[fieldcage_measurements_medidas.pdf](https://github.com/next-exp/nexus/files/6790248/fieldcage_measurements_medidas.pdf)

source/geometries/Next100FieldCage.cc
```diff
@@ -186,12 +198,12 @@ void Next100FieldCage::SetMotherPhysicalVolume(G4VPhysicalVolume* mother_phys)
 void Next100FieldCage::Construct()
 {
   /// Calculate lengths of active and buffer regions
-  active_length_ = gate_cathode_centre_dist_ - grid_thickn_/2.;
+  active_length_ = GetELzCoord() + gate_teflon_dist_ + active_cathode_dist_ + cathode_thickn_/2. - grid_thickn_/2.;
```
paolafer: A length cannot depend on the relative position of the gate in its mother volume (`GetELzCoord()`). This will give you an in correct length of the active volume if it is not zero. I understand you debugged this geometry using the `Next100OpticalGeometry`, where that position is zero. Please, use the `Next100` geometry instead, and check all the dimensions and volume overlaps. You can comment the placement of the ICS for the moment and skip the overlaps of the vessel, since those volumes still need to be reviewed.

source/geometries/Next100FieldCage.cc
```diff
@@ -305,58 +328,79 @@ void Next100FieldCage::BuildActive()
...
   G4Polyhedra* buffer_solid =
-  new G4Polyhedra("BUFFER", 0., twopi, n_panels_, 2, zplane, rinner, router);
+  new G4Polyhedra("BUFFER_POLY", 0., twopi, n_panels_, 2, zplane, rinner, router);
+
+  G4Tubs* buffer_cathode_solid =
+  new G4Tubs("BUFF_CATHODE_RING", 0, cathode_int_diam_/2., (cathode_thickn_/2. - grid_thickn_/2.)/2., 0, twopi);
+
+  G4ThreeVector buff_cathode_pos = G4ThreeVector(0., 0., -buffer_length_/2.+(cathode_thickn_/2. - grid_thickn_/2.)/2.);
+
+  G4SubtractionSolid* subs_buffer =
```
gondiaz: In the BUFFER (white) side of the cathode (blue), there is an "empty space" which is not BUFFER. This region is filled with Xenon since it corresponds with the mother volume. Nevertheless remember that the hits of a simulated event contain a `label` which indicates the volume of the depositions. If we simulate a track starting from the cathode with direction to the BUFFER, we expect those hits to be completely on BUFFER, but that would not be the case now because of this "empty space".

I suggest to redefine the BUFFER with an `G4UnionSolid` instead with a `G4SubtractionSolid` as it is done with the ACTIVE above.

active-cathode-buffer

source/geometries/Next100FieldCage.cc
```diff
@@ -404,12 +448,12 @@ void Next100FieldCage::BuildELRegion()
 {
   /// EL gap
   G4Tubs* el_gap_solid =
-    new G4Tubs("EL_GAP", 0., el_gap_diam_/2., el_gap_length_/2., 0, twopi);
+  new G4Tubs("EL_GAP", 0., el_gap_diam_/2., el_gap_length_/2., 0, twopi);
```
gondiaz: What is the exact `el_gap_length_` ? Since we are placing the EL_GRID_GATE and EL_GRID_CATHODE inside the EL_GAP volume, the `el_gap_length_` is been reduced by `2*grid_thickn_`. Then, I think it would be more correct to this volume z-width to be `el_gap_length_ + 2*grid_thickn_`. 
It might be irrelevant since the grid thickness are very small. Are they? They are not included in the slides.

source/geometries/Next100FieldCage.cc
```diff
@@ -42,18 +45,27 @@ Next100FieldCage::Next100FieldCage():
...
+  teflon_total_length_ (1431. * mm),
   teflon_thickn_ (5 * mm),
   gate_teflon_dist_ (16. * mm),
   n_panels_ (18),
   tpb_thickn_ (1 * micrometer),
   el_gap_diam_ (1009. * mm), // internal diameter of EL meshes
   el_gap_length_ (1. * cm),
+  hdpe_tube_int_diam_ (1080. * mm),
+  hdpe_tube_ext_diam_ (1105.4 * mm),
+  hdpe_length_ (1192. * mm), //check this value
```
gondiaz, line 61: This value C3 is not inlcluded in the slides. Is it correct?

source/geometries/Next100FieldCage.cc
```diff
@@ -42,18 +45,27 @@ Next100FieldCage::Next100FieldCage():
   GeometryBase(),
   // Dimensions
   active_diam_ (984 * mm), // distance between the centers of two opposite panels
-  gate_cathode_centre_dist_ (1205. * mm), // distance between gate and the centre of cathode grid
+  active_cathode_dist_ (1178.*mm), //distance from the  beginning of the teflon drift panels to the first side of the cathode.
   gate_sapphire_wdw_dist_ (1460.5 * mm), // distance between gate and the surface of sapphire windows
```
gondiaz: I don't see these values in the slides, did you measure in the drawings or did you obtained them from another place?

source/geometries/Next100FieldCage.cc
```diff
@@ -69,7 +81,7 @@ Next100FieldCage::Next100FieldCage():
   el_table_binning_(1.*mm),
```
gondiaz: I think that this EL table generation is deprecated since we use now the cilindrical generator, should we get rid off this to clean the code? @jmalbos @paolafer

source/geometries/Next100FieldCage.cc
```diff
@@ -613,6 +656,149 @@ void Next100FieldCage::BuildFieldCage()
...
+                                     "HDPE_TUBE", mother_logic_, false, 0, true);
+
+  G4double active_short_z = 13.5*mm; //Thickness for the small piece from the  holder in front of the ring.
+  G4double first_ring_drif_z_pos = drift_ring_dist_/2. + active_short_z/2. + gate_teflon_dist_ + GetELzCoord();
+  G4double buffer_short_z = 37.*mm;
+
+  G4double teflon_drift_length = active_cathode_dist_;
+  G4double teflon_drift_zpos = GetELzCoord() + gate_teflon_dist_ + teflon_drift_length/2.;
+  G4double teflon_buffer_length = teflon_total_length_ - cathode_thickn_ - teflon_drift_length;
+  G4double teflon_buffer_zpos = teflon_drift_zpos + teflon_drift_length/2. + cathode_thickn_ + teflon_buffer_length/2;
```
gondiaz: These `teflon_*` variables are used in the previous functions, they should be global variables in the class to avoid future bugs.

source/geometries/Next100FieldCage.cc
```diff
@@ -613,6 +656,149 @@ void Next100FieldCage::BuildFieldCage()
...
+               hdpe_tube_ext_diam_/2., hdpe_length_/2., 0, twopi);
+
+  G4LogicalVolume* hdpe_tube_logic = new G4LogicalVolume(hdpe_tube_solid,
+                                                         hdpe_, "HDPE_TUBE");
+
+  new G4PVPlacement(0, G4ThreeVector(0., 0., hdpe_tube_z_pos), hdpe_tube_logic,
+                                     "HDPE_TUBE", mother_logic_, false, 0, true);
+
+  G4double active_short_z = 13.5*mm; //Thickness for the small piece from the  holder in front of the ring.
+  G4double first_ring_drif_z_pos = drift_ring_dist_/2. + active_short_z/2. + gate_teflon_dist_ + GetELzCoord();
```
gondiaz: Is this correct? I might be missing somethig, but from your slides I see that the first ring should be at A3 distance from the start of the teflon. Then its position from the EL gate would be
`first_ring_drif_z_pos = ring_thickn_/2. + active_short_z + gate_teflon_dist_ + GetELzCoord()`

source/geometries/Next100FieldCage.cc
```diff
@@ -613,6 +656,149 @@ void Next100FieldCage::BuildFieldCage()
...
+  G4double buffer_short_z = 37.*mm;
+
+  G4double teflon_drift_length = active_cathode_dist_;
+  G4double teflon_drift_zpos = GetELzCoord() + gate_teflon_dist_ + teflon_drift_length/2.;
+  G4double teflon_buffer_length = teflon_total_length_ - cathode_thickn_ - teflon_drift_length;
+  G4double teflon_buffer_zpos = teflon_drift_zpos + teflon_drift_length/2. + cathode_thickn_ + teflon_buffer_length/2;
+  G4double ring_drift_buffer_dist = 72.*mm;
+  G4double first_ring_buff_z_pos = first_ring_drif_z_pos + 47*drift_ring_dist_ + ring_drift_buffer_dist;
```
gondiaz: Avoid hardcoding the `47` by declaring the `num_drift_rings` above this line.

source/geometries/Next100FieldCage.cc
```diff
@@ -613,6 +656,149 @@ void Next100FieldCage::BuildFieldCage()
...
+  G4double first_act_short_z = -teflon_drift_length/2.+ active_short_z/2.;
+
+  G4Box* active_long_solid =
+    new G4Box("ACT_LONG", active_long_x/2., active_long_y/2., teflon_drift_length/2.);
+
+  G4UnionSolid* act_holder_solid =
+    new G4UnionSolid ("ACT_HOLDER", active_long_solid, active_short_solid, 0,
+                      G4ThreeVector(0.,active_long_y/2.+active_short_y/2.,first_act_short_z));
+
+  for (G4int j=1; j<48; j++) {
```
gondiaz: Avoid hardcoding

source/geometries/Next100FieldCage.cc
```diff
@@ -613,6 +656,149 @@ void Next100FieldCage::BuildFieldCage()
...
+      new G4Box("BUFF_SHORT", buffer_short_x/2., buffer_short_y/2., buffer_short_z/2.);
+
+    G4double first_buff_short_z = -teflon_buffer_length/2. + (ring_drift_buffer_dist/2. - cathode_thickn_/2.) + buffer_ring_dist_/2.;
+    G4Box* buffer_long_solid =
+      new G4Box("BUFF_LONG", buffer_long_x/2., buffer_long_y/2., teflon_buffer_length/2.);
+
+    G4UnionSolid* buff_holder_solid = new G4UnionSolid ("BUFF_HOLDER", buffer_long_solid, buffer_short_solid, 0,
+                                                G4ThreeVector(0.,buffer_long_y/2.+buffer_short_y/2.,first_buff_short_z));
+
+    for (G4int j=1; j<3; j++) {
```
gondiaz: Avoid hardcoding

source/geometries/Next100FieldCage.cc
```diff
@@ -613,6 +656,149 @@ void Next100FieldCage::BuildFieldCage()
...
+                                      G4ThreeVector(0.,buffer_long_y/2. + buffer_short_y/2.,posz));
+      }
+
+    G4double buffer_last_z  = 63.2 *mm;
+    G4Box* buffer_last_solid = new G4Box("ACT_LAST", buffer_short_x/2., buffer_short_y/2., buffer_last_z/2.);
+
+    buff_holder_solid = new G4UnionSolid("BUFF_HOLDER", buff_holder_solid, buffer_last_solid, 0,
+                                      G4ThreeVector(0.,buffer_long_y/2. + buffer_short_y/2., teflon_buffer_length/2. - buffer_last_z/2.));
+
+    G4LogicalVolume* buff_holder_logic = new G4LogicalVolume(buff_holder_solid,pe1000_, "ACT_HOLDER");
```
gondiaz: Rename ACT_HOLDER to BUFF_HOLDER, since you are creating the logical volume of the buffer holders.

gondiaz (commented):
Do we want a vertex generator for the copper rings and the holders? @paolafer @jmalbos @msorel @jmunozv

source/geometries/Next100FieldCage.cc
paolafer: Yes, I think we should.

source/geometries/Next100FieldCage.cc
```diff
@@ -42,18 +45,27 @@ Next100FieldCage::Next100FieldCage():
   GeometryBase(),
   // Dimensions
   active_diam_ (984 * mm), // distance between the centers of two opposite panels
-  gate_cathode_centre_dist_ (1205. * mm), // distance between gate and the centre of cathode grid
+  active_cathode_dist_ (1178.*mm), //distance from the  beginning of the teflon drift panels to the first side of the cathode.
```
paolafer: It's not clear to me what this distance represents. The beginning of the teflon drift is not the same as the beginning of the active region, right? Also, what is `the first side of the cathode`? In the past we used `beginning` and `end`, taking positive z as direction, for instance.

source/geometries/Next100FieldCage.cc
paolafer: In past geometries we used to place the grids inside the EL gap volume. Since they're very thin, it's irrelevant, as you say. I would keep it the same as the other geometries.

source/geometries/Next100FieldCage.cc
paolafer: This was checked with Alberto Martínez in a previous revision. It would be good to double check now.

source/geometries/Next100FieldCage.cc
MiryamMV: Changed! Now there are no overlaps!

source/geometries/Next100FieldCage.cc
MiryamMV: My bad! I don't know what led me to use a `G4SubtractionSolid` instead of `G4UnionSolid`. Anyway, it has already been fixed.

source/geometries/Next100FieldCage.cc
MiryamMV: Right, I had to measure it by myself with Autodesk Viewer.

source/geometries/Next100FieldCage.cc
MiryamMV: I just checked `active_cathode_dist_` and `gate_sapphire_wdw_dist_` with Sara and confirmed they're correct.

source/geometries/Next100FieldCage.cc
MiryamMV: I understand I should remove the use of `CalculateELTableVertices` function and replace it with the said cylindrical generator but I can't find any generator with that name or similar.

gondiaz: The cylindrical generator is already implemented, it is the `el_gap_gen_` and has the configuration parameters below the comment `// EL gap generation disk parameters` in the class constructor.

MiryamMV: Is it OK to remove lines from 869 to 883 then?

source/geometries/Next100FieldCage.cc
MiryamMV: Should I define a function as `DefineMaterials()` in` NextNewFieldCage` for all the materials used in the class?

source/geometries/Next100FieldCage.cc
MiryamMV: I defined `first_ring_drif_z_pos` like that to avoid possible infinitesimal overlaps because I wasn't sure if the space holding the rings (marked in yellow in the image) was exactly equal to the `ring_thickn_`
first_ring_drif_z_pos

source/geometries/Next100FieldCage.cc
gondiaz: It is OK, but you must also remove all the code related with this generator, namely lines 80-83, 144-151, 225-231,  and the `CalculateELTableVertices` function. I might be missing something, thus check that there is not unneccesary code left.

source/geometries/Next100FieldCage.cc
gondiaz: No, they are not material variables, they are distances. These variables `teflon_*`, defined in lines 678-681 are defined inside the function `BuildFieldCage`, but they are also defined in the previous function `BuildLightTube` in lines 538, 539, 574, 575. This is very bug prone since if you change one on these distances you must do it on two different places. The solution to avoid this duplication is to define them as global variables and define them once, as it is done for variables in lines 200-207 in the `Construct`. I suggest to place the `teflon_*` variables there also. Remember that global variable names finish with an underscore, like `active_length_`.

source/geometries/Next100FieldCage.cc
gondiaz: Ok, I see, it is clear now. I hadn't realised that the `ring_thickn_` is not equal to the spacing where they are positioned.

source/geometries/Next100FieldCage.cc
```diff
@@ -245,20 +248,28 @@ void Next100FieldCage::DefineMaterials()
...
-  new G4Polyhedra("ACTIVE", 0., twopi, n_panels_, 2, zplane, rinner, router);
+  new G4Polyhedra("ACTIVE_POLY", 0., twopi, n_panels_, 2, zplane, rinner, router);
+
+  G4Tubs* active_cathode_solid =
+  new G4Tubs("ACT_CATHODE_RING", 0, cathode_int_diam_/2., (cathode_thickn_/2. - grid_thickn_/2.)/2., 0, twopi);
+
+  G4ThreeVector act_cathode_pos = G4ThreeVector(0., 0., active_length_/2.-(cathode_thickn_/2. - grid_thickn_/2.)/2.);
+
+  G4UnionSolid* union_active =
+  new G4UnionSolid ("ACTIVE", active_solid, active_cathode_solid, 0, act_cathode_pos);
```
paolafer: With unions of solids it is better not to join a volume right at the side of the other, but to make one volume slightly larger so that there is actually a common volume between the two. This way border problems in the tracking of particles are avoided. The same holds for the buffer part and the hdpe holders. We can talk about the implementation online.

source/geometries/Next100FieldCage.cc
```diff
@@ -245,20 +248,28 @@ void Next100FieldCage::DefineMaterials()
...
 
   G4LogicalVolume* active_logic =
-  new G4LogicalVolume(active_solid, gas_, "ACTIVE");
+  new G4LogicalVolume(union_active, gas_, "ACTIVE");
 
   new G4PVPlacement(0, G4ThreeVector(0., 0., active_zpos_), active_logic,
-  "ACTIVE", mother_logic_, false, 0, false);
+  "ACTIVE", mother_logic_, false, 0, true);
```
paolafer: Put all the overlap variables to false, before submitting the PR. This way we avoid a lot of not needed verbosity in productions and in the run of tests.

source/geometries/Next100FieldCage.cc
```diff
@@ -290,12 +301,11 @@ void Next100FieldCage::BuildActive()
 
 
   /// Visibilities
+  //G4VisAttributes active_col = nexus::Yellow();
```
paolafer: Don't leave visibility lines commented. Either delete them or use them.

source/geometries/Next100FieldCage.cc
```diff
@@ -386,9 +417,9 @@ void Next100FieldCage::BuildBuffer()
   /// Visibilities
   buffer_logic->SetVisAttributes(G4VisAttributes::Invisible);
 
-  // G4VisAttributes active_col = nexus::Yellow();
-  // active_col.SetForceSolid(true);
-  // buffer_logic->SetVisAttributes(active_col);
+  //G4VisAttributes active_col = nexus::Yellow();
```
paolafer: Same as above: please, delete these lines or use them.

source/geometries/Next100FieldCage.cc
```diff
@@ -613,6 +636,146 @@ void Next100FieldCage::BuildFieldCage()
...
+               hdpe_tube_ext_diam_/2., hdpe_length_/2., 0, twopi);
+
+  G4LogicalVolume* hdpe_tube_logic = new G4LogicalVolume(hdpe_tube_solid,
+                                                         hdpe_, "HDPE_TUBE");
+
+  new G4PVPlacement(0, G4ThreeVector(0., 0., hdpe_tube_z_pos), hdpe_tube_logic,
+                                     "HDPE_TUBE", mother_logic_, false, 0, true);
+
+  G4double active_short_z = 13.5*mm; //Thickness for the small piece from the  holder in front of the ring.
+  G4double first_ring_drif_z_pos = drift_ring_dist_/2. + active_short_z/2. + gate_teflon_dist_ + GetELzCoord();
```
paolafer: I would use `drift` instead of `drif`, to be consistent across the class.

source/geometries/Next100FieldCage.cc
```diff
@@ -613,6 +636,146 @@ void Next100FieldCage::BuildFieldCage()
...
+    //G4cout << "i " << i << G4endl;
+    //G4cout << "90-i " << 90-i << G4endl;
+    G4RotationMatrix* rot = new G4RotationMatrix();
+    rot -> rotateZ((90-i) *deg);
+    new G4PVPlacement(rot, G4ThreeVector(r*cos(i*deg), r*sin(i*deg), teflon_drift_zpos_), act_holder_logic,
+                                       "ACT_HOLDER", mother_logic_, false, numbering, true);
+    numbering +=1;}
+
+    // BUFFER holders.
+    G4double buffer_long_x = active_long_x;
```
paolafer: What's the point of redefining the variables if they are exactly the same? You could use a generic name instead, which illustrates both uses, like `holder_long_x`, for instance. This holds for `r`, as well.

paolafer (changes requested):

source/geometries/Next100FieldCage.cc
MiryamMV: It is actually the length of the teflon drift panels. If it is okay I think I should change the name of the variable to `teflon_drift_length_` to make the nomenclature clearer.

source/geometries/Next100FieldCage.cc
MiryamMV: Is it OK if I add 1mm to the thickness of one of the solids that are going to be joined (and modify the position accordingly)?  Would that be "easily  understandable" for a future developer?

source/geometries/Next100FieldCage.cc
MiryamMV: Done!

source/geometries/Next100FieldCage.cc
MiryamMV: Done!

source/geometries/Next100FieldCage.cc
MiryamMV: Done!

source/geometries/Next100FieldCage.cc
MiryamMV: Done!

source/geometries/Next100FieldCage.cc
MiryamMV: Done!

source/geometries/Next100FieldCage.cc
MiryamMV: Done!

source/geometries/Next100FieldCage.cc
paolafer: Yes, that would be more straightforward. In the comment you can say `distance from the gate to the beginning of the cathode volume`

source/geometries/Next100FieldCage.cc
paolafer: Yes, 1 mm is enough. You can create a variable with that value and comment about it, if you think it is clearer.

source/geometries/Next100FieldCage.cc
MiryamMV: ok!

source/geometries/Next100FieldCage.cc
MiryamMV: ok!

source/geometries/Next100FieldCage.cc
MiryamMV: ok!

source/geometries/Next100FieldCage.cc
MiryamMV: done!

source/geometries/Next100FieldCage.cc
MiryamMV: done!

source/geometries/Next100FieldCage.cc
MiryamMV: done!

source/geometries/Next100FieldCage.cc
MiryamMV: done!

gondiaz:
Cosmetic revision:
- Use nexus conventions [nexus-coding-conventions](https://github.com/next-exp/nexus/wiki/Coding-conventions-and-style-guide). There are plenty of lines exceeding the 80-100 character convention.
- Indent lines 206-208. Besides, I think it is clearer if you define `el_gap_zpos_ = GetELzCoord() - el_gap_length_/2.`.
- Be consistent with vertical spacings. For example, set a fixed 2 line spacing between methods. Also, remove at least one of the lines between 667-669.
- Be consistent with horizontal spacings. For example, look the definitions at lines 329 and 342. I would prefer the 342 indentation since it is how is mostly used in nexus.
- Remove right slashed in line 627.
- Remove commented lines 707 and 708.

source/geometries/Next100FieldCage.cc
```diff
@@ -679,21 +830,6 @@ G4ThreeVector Next100FieldCage::GenerateVertex(const G4String& region) const
...
-      vertex = table_vertices_.at(i);
-      el_table_index_++;
-    }
-    catch (const std::out_of_range& oor) {
-      G4Exception("[Next100FieldCage]", "GenerateVertex()", FatalErrorInArgument,
-      "EL lookup table point out of range.");
-    }
-  }
 
   else if (region == "EL_GAP") {
```
gondiaz, line 964: This vertex is not working since it is not included in `Next100.cc`. Also, add a `do-while` protection to ensure that the vertex is always inside the EL_GAP volume.

gondiaz:
PE100 holders surrounding the cathode are not implemented. I would add them also.

source/geometries/Next100FieldCage.cc
```diff
@@ -612,6 +634,135 @@ void Next100FieldCage::BuildFieldCage()
...
+  /// Visibilities
+  if (visibility_) {
+    G4VisAttributes ring_col = nexus::CopperBrown();
+    ring_logic->SetVisAttributes(ring_col);
+    G4VisAttributes hdpe_col = nexus::White();
+    hdpe_tube_logic->SetVisAttributes(hdpe_col);
+    G4VisAttributes hold_col = nexus::LightGrey();
+    act_holder_logic->SetVisAttributes(hold_col);
+    buff_holder_logic->SetVisAttributes(hold_col);
+  } else {
```
gondiaz, line 887: Make hdpe tube invisible if visibility_ is false.

source/geometries/Next100FieldCage.cc
MiryamMV: Done!

source/geometries/Next100FieldCage.cc
MiryamMV: Done!

gondiaz (changes requested):
As we talked offline, a more natural definition of the optical surface would be between the volumes from which the photon enters and arrives to. In this case the photon enters from the ACTIVE (Xe) to the TPB (and viceversa). I suggets to change the current implementation that creates the interface between the mother and the TPB.

https://github.com/next-exp/nexus/blob/4a07b236a6b9b4100c212c200eae7f62793c7b21/source/geometries/Next100FieldCage.cc#L676

MiryamMV:
Here we have a list of magnitudes that are still to confirm:
- position of the cathode grid in the cathode ring.
- simulation of the photoetched grids.
- gate-TP distance
- gate-teflon panels distance
- gate/anode internal and external diameters
- gate_ring thickness
- cathode, anode and gate transparency

gondiaz:
This PR reviews the Next100 field cage elements, their positions and sides. There are still some parameters that must be reviewed as soon as the final design is provided by technicians. Particularly important is the gate-to-tracking-plane distance, still not confirmed. The implementation is complete and accurate, therefore I approve it. Great work!

gondiaz (approved):

PR #119, opened 2021-07-16
Next100 vessel

gondiaz:
This PR reviews the NEXT100 steel vessel which covers the inner elements. The geometry was almost rewritten, since based on the drawings, the dimensions and some elements did not correspond with the previous implementation. Overall the changes are the following:
- Use the correct dimensions based on drawings.
- Implement the calibration ports and its vertex generators.
- Remove Vacuum manifold: it does not exists.
- Nozzless are left commented, but I suggest to remove them if they are not neccessary.
- Adapt vessel placement inside the inner air, based in the new shielding implementation.

Issues and notes: 
- there are some volume overlaps that should be fixed in subsequent PRs.
- holes in the ICS (future PR) must be aligned with the calibration ports.
- dimensions and components are taken from the 2D drawings whenever possible. If not, they are meassured in the 3D drawings, which sometimes only provide approximate values.

Below, I append a summary of the vessel meassurements and parts, as well as the 2D drawing from which I took the information.
[next100vessel.pdf](https://github.com/next-exp/nexus/files/6900042/next100vessel.pdf)
[00-02.pdf](https://github.com/next-exp/nexus/files/6831307/00-02.pdf)
[000-90.pdf](https://github.com/next-exp/nexus/files/6831308/000-90.pdf)
[000-91.pdf](https://github.com/next-exp/nexus/files/6831309/000-91.pdf)
[000-92.pdf](https://github.com/next-exp/nexus/files/6831310/000-92.pdf)

### DIFF
--- a/source/geometries/Next100Vessel.cc
+++ b/source/geometries/Next100Vessel.cc
@@ -258,39 +258,6 @@ namespace nexus {
       new G4PVPlacement(0, G4ThreeVector(0.,0.,0.), vessel_gas_logic,
                         "VESSEL_GAS", vessel_logic, false, 0);
 
-
-    //// Vacuum Manifold
-    G4double vacuum_manifold_rad = nozzle_ext_diam_/2. - vessel_thickness_;
-    G4double vacuum_manifold_length = (large_nozzle_length_ - vessel_body_length_) / 2. + 9.4*cm; // 10.6 cm comes from Derek's drawings
-                                                                                                  // Switched to 9.4 to be aligned with Energy Plane
-    G4double vacuum_manifold_zpos = -1. * (large_nozzle_length_ - vacuum_manifold_length) / 2.;
-
-    G4Tubs* vacuum_manifold_solid = new G4Tubs("VACUUM_MANIFOLD", 0.*cm, vacuum_manifold_rad,
-					       vacuum_manifold_length/2., 0.*deg, 360.*deg);
-
-    G4LogicalVolume* vacuum_manifold_logic = new G4LogicalVolume(vacuum_manifold_solid,
-								 G4NistManager::Instance()->FindOrBuildMaterial("G4_Cu"),
-								 "VACUUM_MANIFOLD");
-
-    new G4PVPlacement(0, G4ThreeVector(0.,0.,vacuum_manifold_zpos), vacuum_manifold_logic,
-		      "VACUUM_MANIFOLD", vessel_gas_logic, false, 0);
-
-
-    G4double vacuum_manifold_gas_rad = vacuum_manifold_rad - vessel_thickness_;
-    G4double vacuum_manifold_gas_length = vacuum_manifold_length - 2. * vessel_thickness_;
-
-    G4Tubs* vacuum_manifold_gas_solid = new G4Tubs("VACUUM_MANIFOLD_GAS", 0.*cm, vacuum_manifold_gas_rad,
-						   vacuum_manifold_gas_length/2., 0.*deg, 360.*deg);
-
-    G4LogicalVolume* vacuum_manifold_gas_logic = new G4LogicalVolume(vacuum_manifold_gas_solid,
-								     G4NistManager::Instance()->FindOrBuildMaterial("G4_Galactic"),
-								     "VACUUM_MANIFOLD_GAS");
-
-    new G4PVPlacement(0, G4ThreeVector(0.,0.,0.), vacuum_manifold_gas_logic,
-		      "VACUUM_MANIFOLD_GAS", vacuum_manifold_logic, false, 0);
-
-
-
     // SETTING VISIBILITIES   //////////
     if (visibility_) {
       G4VisAttributes grey = DarkGrey();
@@ -299,15 +266,10 @@ namespace nexus {
       grey.SetForceSolid(true);
       vessel_gas_logic->SetVisAttributes(grey);
       //grey.SetForceSolid(true);
-      vacuum_manifold_logic->SetVisAttributes(grey);
-      //grey.SetForceSolid(true);
-      vacuum_manifold_gas_logic->SetVisAttributes(grey);
     }
     else {
       vessel_logic->SetVisAttributes(G4VisAttributes::Invisible);
       vessel_gas_logic->SetVisAttributes(G4VisAttributes::Invisible);
-      vacuum_manifold_logic->SetVisAttributes(G4VisAttributes::Invisible);
-      vacuum_manifold_gas_logic->SetVisAttributes(G4VisAttributes::Invisible);
     }
 
 
--- a/source/geometries/Next100Vessel.cc
+++ b/source/geometries/Next100Vessel.cc
@@ -57,10 +57,12 @@ namespace nexus {
     endcap_gate_distance_(48.62  * cm),
 
     // Ports (values set on Construct())
-    // They are defined global since are needed at the vertex generation
-    port_x_(0.),
-    port_y_(0.),
-    port_z_1a_ (0.),
+    port_base_height_(37. * mm),
+    port_tube_height_(port_base_height_),
+    // They are defined global because are needed at the vertex generation
+    port_x_ ((vessel_in_rad_ + port_base_height_ - port_tube_height_/2.)/sqrt(2.)), // inner port pos
+    port_y_ (port_x_),
+    port_z_1a_ (0.), // defined in the Construct()
     port_z_1b_ (0.),
     port_z_2a_ (0.),
     port_z_2b_ (0.),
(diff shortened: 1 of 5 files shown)


source/geometries/Next100Vessel.cc
```diff
@@ -120,112 +124,166 @@ namespace nexus {
...
-				    0, G4ThreeVector(0., bottom_nozzle_ypos_, 0.) );
-
-
-
-    // Body gas + Tracking endcap gas
+    // Body + Energy endcap + Tracking endcap + Energy flange + Tracking flange
+    vessel_solid = new G4UnionSolid("VESSEL", vessel_solid,
+				    vessel_tp_flange_solid, 0, tracking_flange_pos);
+
+    // Add ports (x,y at 45*deg)
```
halmamol: Add in the 'next100vessel.pdf' document the information about the ports position (45 deg angle)

source/geometries/Next100Vessel.cc
```diff
@@ -120,112 +124,166 @@ namespace nexus {
...
+    G4double port_x = (vessel_out_rad+port_base_height/2.)/sqrt(2.);
+    G4double port_y = port_x;
+    port_gas_x_ = (vessel_in_rad_+port_base_height/2.+vessel_thickness_/2.)/sqrt(2.);
+    port_gas_y_ = port_gas_x_;
+
+    G4double port_reference_z = -body_length_/2. + endcap_in_body_;
+    port_z_1a_ = port_reference_z +   702. * mm;
```
halmamol, line 236: I would add at some point in the 'Next100Vessel.cc' file which port is each one (or at least which side is a or b) with comments, not only in the PR document. It's not intuitive the way they are defined right now. Also, I wonder if we should numerate them with respect to their distant to the TP.

source/geometries/Next100Vessel.cc
```diff
@@ -253,86 +311,57 @@ namespace nexus {
...
 
-  }
+    perc_endcap_vol_    = 2.*endcap_vol / vessel_vol;
+    perc_ep_flange_vol_ = flange_ep_vol / vessel_vol;
+    perc_tp_flange_vol_ = flange_tp_vol / vessel_vol;
 
+    // G4double body_vol = vessel_body_solid->GetCubicVolume() - vessel_gas_body_solid->GetCubicVolume();
+    // std::cout << "Body vol (cm3)" << body_vol/1e6 << '\n';
+    // G4double endcap_vol = vessel_endcap_solid->GetCubicVolume() - vessel_gas_endcap_solid->GetCubicVolume();
+    // std::cout << "Endcap vol (cm3)" << endcap_vol/1e6 << '\n';
```
halmamol: Why are these lines commented?

gondiaz: I wrote those lines for debugging purpuses while developing the code, I dont find them usefull anymore thus I would remove them in a cosmetics commit.

source/geometries/Next100Vessel.cc
```diff
@@ -120,112 +124,166 @@ namespace nexus {
...
+    G4double flange_ep_length = 77.5 * mm + 41.5 * mm;
+    G4double flange_ep_z_pos  =   body_length_/2. - flange_ep_length/2. - endcap_in_body_;
+    G4double flange_tp_z_pos  = -(body_length_/2. - flange_tp_length/2. - endcap_in_body_);
+
+    G4Tubs* vessel_tp_flange_solid = new G4Tubs("VESSEL_TRACKING_FLANGE", vessel_in_rad_, flange_out_rad,
+                                                flange_tp_length/2., 0.*deg, 360.*deg);
+    G4Tubs* vessel_ep_flange_solid = new G4Tubs("VESSEL_ENERGY_FLANGE"  , vessel_in_rad_, flange_out_rad,
+                                                flange_ep_length/2., 0.*deg, 360.*deg);
+
+    // Calibration ports (approximate values)
```
halmamol: You can get the exact values from '000-90.pdf' (4 and 5, in LL)

source/geometries/Next100Vessel.cc
gondiaz: Done

source/geometries/Next100Vessel.cc
gondiaz: I have updated the values following this 2D drawing, except the heights that can not be infered from it, but can be easily measured in the 3D drawing.

gondiaz:
As we talked offline @halmamol, I have implemented the internal tube of the calibration port filled with air. The tube only spans up to the vessel internal wall, similarly as in NextNew. The vertex generation is done throughout the whole tube. 
This is simplified version of the source calibration generation and can be modified in the future once we implement the ICS holes and final confirmation on the source emission points.

source/geometries/Next100Vessel.cc
halmamol: Yes, could you remove them? I think they are not necessary.

gondiaz: Done

halmamol:
I have checked the PR and as it is mentioned, it gives an updated version of the NEXT100 Vessel using up to date design drawings. It includes the vessel body, removes parts that are not included in the final design (as the 'vacuum manifold') and implements the calibration ports and their vertex generators.

Prior any final approval, needs to be mentioned/discussed:
- As it was arised by @gondiaz, there are some overlaps with inner elements, since the ICS is still not updated. We need to take that into account and fix it in future PRs.
- With the exception of the calibration ports, there are no nozzles in this version. The ones implemented in the original design are left commented since they are outdated. We could discuss offline the current desing, and depending on that, see what is needed on this PR or in future ones. What do you think @paolafer @jmalbos @gondiaz?
- With respect to the calibration ports, two things. Firstly, they are named so far _1a/1b/2a/2b_, more intuitive names should be defined for final desings. Secondly, the vertex generation is along the whole port tube, and not from a specific vertex. We need to decide how do we want it, maybe we could discuss it together with the nozzles. 

With the exception of these points, what is included in this PR works fine and looks ready for approval.

source/geometries/Next100Vessel.cc
```diff
@@ -253,88 +322,79 @@ namespace nexus {
...
+      vessel_logic       ->SetVisAttributes(G4VisAttributes::Invisible);
+      vessel_gas_logic   ->SetVisAttributes(G4VisAttributes::Invisible);
+      port_tube_logic    ->SetVisAttributes(G4VisAttributes::Invisible);
+      port_tube_gas_logic->SetVisAttributes(G4VisAttributes::Invisible);
     }
 
-
     // VERTEX GENERATORS   //////////
-    body_gen_  = new CylinderPointSampler(vessel_in_rad_, vessel_length_, vessel_thickness_, 0.);
+    body_gen_  = new CylinderPointSampler(vessel_in_rad_, body_length_, vessel_thickness_, 0.);
```
paolafer: I think we should use the new version of `CylinderPointSampler` (`CylinderPointSampler2020`), since we're modifying the entire class.

source/geometries/Next100Vessel.cc
```diff
@@ -357,88 +417,78 @@ namespace nexus {
...
-    // Vertex in FLANGES
-    else if (region == "VESSEL_FLANGES") {
-      if (G4UniformRand() < 0.5)
-      	vertex = tracking_flange_gen_->GenerateVertex("BODY_VOL");
-      else
-      	vertex = energy_flange_gen_->GenerateVertex("BODY_VOL");
+    else if (region == "PORT_1a"){
+      vertex = port_gen_->GenerateVertex("WHOLE_VOL");
+
+      vertex = vertex.rotateX( 90. * deg);
```
paolafer, line 454: Is there a logical way to relate this rotation to that of the unions of the port volumes with the vessel, in order not to hard-code all these numbers?

paolafer (changes requested):

paolafer:
Please, review all the solid unions, to add some overlap between volumes, in case some of them don't have it.

paolafer:
> The vertex generation is done throughout the whole tube.
This is simplified version of the source calibration generation and can be modified in the future once we implement the ICS holes and final confirmation on the source emission points.

So, if I understand correctly, the `PORT_1a` etc. vertices are meant to simulate calibration sources, not background. I have two comments here: 1. do we need to generate also background from the port volumes (not the internal tube, but the ports of the vessel)? Or do we assume that the volume is negligible? I think that right now the `VESSEL` vertex generator does not include the ports (correct me if I'm wrong), 2. it's quite unlikely that a calibration source is distributed along the internal tube, right? It is usually a capsule, placed at some distance along the tube (usually at the bottom of it). Probably we will want to simulate calibration sources before having a true source to model, so I suggest to model the vertex as the sources we use in NEW.

source/geometries/Next100Vessel.cc
gondiaz: Done, `CylinderPointSampler` substituted by `CylinderPointSampler2020`

gondiaz:
Yes, `PORT_1a`, etc.. are the ports to simulate the calibration sources. They are composed by a nozzle and by an inner tube (as you might see in the last slide).

From your comment 2), I have modified the ports such that:
- the vertex is generated from the tip of the port, in a cylinder of size `source_height_`. This size is preliminary set to 5 mm, but we should get this value in the future, as well as the exact source position.
- I have added the variable `port_tube_height_` to set the inner tube height, where the source lives. Once the ICS is implemented with the holes at the right place, we can enlarge the inner tube by just changing this variable. This variable is set now to `port_base_height_` in order to avoid overlap with the currently implemented ICS.

From your comment 1), you are right that the nozzles are not included in the VESSEL vertex generator. The steel volume of the nozzles is negligible thus I think that implementing it would make the code unnecessary complex. We are also neglecting steel from the gas nozzles.

Also, I have added an offset to overlap the union of the steel port nozzle base and cap. That is the only union that was not overlaped.

source/geometries/Next100Vessel.cc
```diff
@@ -366,21 +366,23 @@ namespace nexus {
...
 
     energy_endcap_gen_ = new SpherePointSampler( endcap_in_rad_, vessel_thickness_, energy_endcap_pos, 0,
 						   0., twopi, 0., endcap_theta_);
 
     tracking_endcap_gen_ = new SpherePointSampler( endcap_in_rad_, vessel_thickness_, tracking_endcap_pos, xRot,
 						 0., twopi, 0., endcap_theta_);
 
-    tracking_flange_gen_  = new CylinderPointSampler(vessel_out_rad, flange_tp_length,
-						     flange_out_rad-vessel_out_rad, 0., tracking_flange_pos);
+    tracking_flange_gen_ = new CylinderPointSampler2020(vessel_in_rad_, flange_out_rad, flange_tp_length/2.,
```
paolafer, line 380: Is it possible to use the constructor that takes the physical volume? The idea of this new class of generators is to use the physical volume, whenever is possible, to avoid potential mistakes.

gondiaz: Probably, but:

1) the generator must go in Next100.cc since is where the physical volume is defined; and
2) since it is a cilindrical generator I understand that it would create a cilinder covering the whole vessel volume and just keep the vertices inside the vessel. I remember a conversation in which we should avoid that because it can be inneficient due to most of the randoms are discarded.

That's why I maintained the current approach, should I change it anyway?

paolafer: I see that the physical volume is the union of the several volumes that make up the vessel; in that case it's better to leave it as you implemented it, because that volume is not a cylinder. I don't understand your point 2: the generator reads the minimum and maximum radius, so if the cylinder is hollow, it takes that into account.

gondiaz: Right, the point is that the volume is not a cylinder: it is an union of a cylinder, the flanges, the endcaps and the port nozzles. Then, do we agree to leave it as it is?

paolafer: Yes.

source/geometries/Next100Vessel.cc
gondiaz: No, I didnt find any logical way since `G4PVPlacement` and `G4UnionSolid` rotations and `G4ThreeVector` rotations work differently. In the first two cases, when performing a multiple rotation (ie. one after the other) you do it with reference to the rotated axis. However, the `G4ThreeVector` references the original axis.

halmamol:
I have reviewed the last commits, and everything seems fine. Just for the record, I'm going to summarize the few points that we need to take into account in future PRs:
- Overlaps with inner elements should be corrected when the ICS is updated.
- There are no nozzles in this version, the ones implemented in the original design (outdated) are left commented. We should check if we want to include any from the final design.
- Calibration ports are preliminary named _1a/1b/2a/2b_ (check PR files for reference).
- The variable `source_height_` is created to define the vertex creation, with a preliminary value of 5mm, that need to be updated in the future.

halmamol (approved):
This PR uses up to date drawings, and gives an updated version of the NEXT100 Vessel. It removes parts that are not included in the final design (as the 'vacuum manifold', and nozzles) and includes the calibration ports and their vertex generators. I approve it!

PR #120, opened 2021-07-21
Next100 energy plane

gondiaz:
This PR reviews the current implementation of the NEXT-100 energy plane. The changes are the following:
- Set the dimensions according with the actual drawings.
- Fix `offset` displacement in the copper holes, in the subtraction solid definition.
- Remove the external pmt base vertex generator, since only the "internal" base exists.

Meassurements were taken from the following 2D drawings by Jordi, whenever possible. If not, the 3D drawings were used.
[02-07.pdf](https://github.com/next-exp/nexus/files/6856722/02-07.Copper.plate.pdf)
[02-08.pdf](https://github.com/next-exp/nexus/files/6856723/02-08.pdf)

Besides, Jordi provided the following measurements for the "HUTs" (this term with U is how the copper hats are named in the code) via private communication:
-18 “HATs” with L=120mm >>> central hexagon
-18 “HATs” with L=100mm >>> second periferic line
-24 “HATs” with L=60mm >>> external periferic line

### DIFF
--- a/source/geometries/Next100EnergyPlane.cc
+++ b/source/geometries/Next100EnergyPlane.cc
@@ -62,8 +62,8 @@ namespace nexus {
     optical_pad_thickn_ (1.0 * mm),
     tpb_thickn_ (1.*micrometer),
     pmt_stand_out_ (2. * mm), // length that PMTs stand out of copper, in the front
-    internal_pmt_base_diam_ (46.8 * mm),
-    internal_pmt_base_thickn_ (0.5 * mm),
+    pmt_base_diam_ (46.8 * mm),
+    pmt_base_thickn_ (0.5 * mm),
     visibility_(1),
     verbosity_(0)
   {
@@ -344,19 +344,18 @@ namespace nexus {
 
 
     /// Part of the PMT bases with pins and resistors ///
-    G4Tubs* internal_pmt_base_solid =
-      new G4Tubs("INTERNAL_PMT_BASE", 0., internal_pmt_base_diam_/2.,
-		 internal_pmt_base_thickn_/2., 0., twopi);
+    G4Tubs* pmt_base_solid =
+      new G4Tubs("PMT_BASE", 0., pmt_base_diam_/2., pmt_base_thickn_/2., 0., twopi);
 
-    G4LogicalVolume* internal_pmt_base_logic =
-      new G4LogicalVolume(internal_pmt_base_solid,
-			  G4NistManager::Instance()->FindOrBuildMaterial("G4_KAPTON"),
-			  "INTERNAL_PMT_BASE");
-    G4double int_pmt_base_posz =
+    G4LogicalVolume* pmt_base_logic =
+      new G4LogicalVolume(pmt_base_solid,
+			     G4NistManager::Instance()->FindOrBuildMaterial("G4_KAPTON"), "PMT_BASE");
+
+    G4double pmt_base_posz =
       vacuum_front_length/2. + hole_length_rear_ + hut_hole_length_/2.;
-    G4VPhysicalVolume* internal_pmt_base_phys =
-      new G4PVPlacement(0, G4ThreeVector(0., 0., int_pmt_base_posz),
-                        internal_pmt_base_logic, "INTERNAL_PMT_BASE",
+    G4VPhysicalVolume* pmt_base_phys =
+      new G4PVPlacement(0, G4ThreeVector(0., 0., pmt_base_posz),
+                        pmt_base_logic, "PMT_BASE",
                         vacuum_logic, false, 0, false);
 
     /// Placing the encapsulating volume with all internal components in place ///
@@ -404,7 +403,7 @@ namespace nexus {
       tpb_logic->SetVisAttributes(tpb_col);
       G4VisAttributes pmt_base_col = Yellow();
       pmt_base_col.SetForceSolid(true);
-      internal_pmt_base_logic->SetVisAttributes(pmt_base_col);
+      pmt_base_logic->SetVisAttributes(pmt_base_col);
       //G4VisAttributes vacuum_col = Red();
       //vacuum_col.SetForceSolid(true);
       // vacuum_logic->SetVisAttributes(vacuum_col);
@@ -413,7 +412,7 @@ namespace nexus {
       sapphire_window_logic->SetVisAttributes(G4VisAttributes::Invisible);
       optical_pad_logic->SetVisAttributes(G4VisAttributes::Invisible);
       tpb_logic->SetVisAttributes(G4VisAttributes::Invisible);
-      internal_pmt_base_logic->SetVisAttributes(G4VisAttributes::Invisible);
+      pmt_base_logic->SetVisAttributes(G4VisAttributes::Invisible);
     }
 
     //////////////////////////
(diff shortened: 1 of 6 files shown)


source/geometries/Next100EnergyPlane.cc
```diff
@@ -238,7 +237,7 @@ namespace nexus {
 			  "EP_COPPER_PLATE");
 
     G4double stand_out_length =
-      sapphire_window_thickn_ + optical_pad_thickn_ + pmt_stand_out_;
+      sapphire_window_thickn_ + tpb_thickn_ + optical_pad_thickn_ + pmt_stand_out_;
```
paolafer, line 204: I think that the TPB thickness should not be added, since the TPB is placed inside the sapphire window, therefore its thickness is included in that of the window.

source/geometries/Next100EnergyPlane.cc
```diff
@@ -207,11 +207,10 @@ namespace nexus {
 			 0, G4ThreeVector(0., 0., transl_z));
 
     G4Tubs* hole_hut_solid =
-      new G4Tubs("HOLE_HUT", 0., hut_int_diam_/2., (hut_hole_length_ + offset)/2.,
```
paolafer, line 210: I don't understand why you removed the offset here. We can discuss it online, if it's easier.

source/geometries/Next100EnergyPlane.cc
```diff
@@ -207,11 +207,10 @@ namespace nexus {
...
-		 0., twopi);
+      new G4Tubs("HOLE_HUT", 0., hut_int_diam_/2., hut_hole_length_/2.,
+		  0., twopi);
     transl_z =
-      (hole_length_front_ + offset)/2. + hole_length_rear_ + hut_hole_length_/2.
-      - offset/2.;
```
paolafer, line 214: Same as above. I believe you have to keep `-offset/2`.

source/geometries/Next100EnergyPlane.cc
gondiaz: I have added `tbp_thickn_` into the sapphire window volume, then this value must be added here also. In the previous implementation the `tpb_thickn_` is not added to the sapphire volume, I guess because it is negligible.

paolafer: Yes, it is irrelevant. Either implementation works.

paolafer: Screenshot 2021-07-26 at 11 30 17

However, the placement of TPB doesn't seem right: a geantino coming from the inside of the chamber should see the TPB first, not the sapphire window.

paolafer: I think that this line:

`G4double tpb_posz = - sapphire_window_thickn_/2. + tpb_thickn_/2.;`

should rather be

`G4double tpb_posz = - (sapphire_window_thickn_+ tpb_thickn_)/2. - tpb_thickn_/2.;`

, which allows one to see the steps logically.

paolafer: I think that 

`G4double tpb_posz = - sapphire_window_thickn_/2. + tpb_thickn_/2.;`

should be replaced with

`G4double tpb_posz = - sapphire_window_thickn_/2. ;`, 

although I think that writing

`G4double tpb_posz = - (sapphire_window_thickn_ + tpb_thickn_)/2.  - tpb_thickn_/2.`

is clearer.

source/geometries/Next100EnergyPlane.cc
gondiaz: Using an offset as in this class is usefull to make an G4Union between a a smaller volume into a larger one, since the `offset` part of the smaller volume is included into the larger. But this is not true here because the hole in the hut is larger (higher radious) than the hole in the copper plate. Such union makes the hole inside the copper to be biased, as you can see in the figures
Expected holes:
expected-holes

Holes with the offset between hut and copper:
holes with offset

As you see in the second figure, part of the hut hole is embedded into the copper plate, which is not correct.

source/geometries/Next100EnergyPlane.cc
gondiaz: Right, I forgot to add the `tpb_thickn_` to the translation in the sapphire volume.

source/geometries/Next100EnergyPlane.cc
```diff
@@ -46,24 +46,24 @@ namespace nexus {
     // hole_lat1_posy_ (-203.8 * mm),
     // hole_lat2_posx_ (-415 * mm),
     // hole_lat2_posy_ (-311.2 * mm),
-    hut_int_diam_ (76. * mm),
-    hut_thickn_ (5. * mm),
+    hut_int_diam_ (64. * mm),
+    hut_thickn_ (6. * mm),
```
MiryamMV: This variable is only used once in the whole class. Perhaps we should define hut_diam_(line 152) as global (which is used more) and whose value is explicitly indicated in the drawings.

gondiaz: Done

source/geometries/Next100EnergyPlane.cc
```diff
@@ -98,154 +98,114 @@ namespace nexus {
...
+                                                    copper_plate_thickn_/2., 0., twopi);
 
-    G4double offset = 1. * cm;
-
-    /// Holes for gas flow ///
+    /// Hole for gas flow ///
     G4Tubs* copper_plate_gas_hole_solid =
       new G4Tubs("COPPER_PLATE_CENTRAL_HOLE", 0., gas_hole_diam_/2.,
-		 (copper_plate_thickn_ + offset)/2., 0., twopi);
+		             (copper_plate_thickn_ + offset)/2., 0., twopi);
```
MiryamMV: The hole doesn't reach the back of the copper plate (it is shorter, about 103mm according to the 2D drawings).

source/geometries/Next100EnergyPlane.cc
```diff
@@ -260,119 +220,111 @@ namespace nexus {
...
+                       G4ThreeVector(0., 0., vacuum_front_length/2.+hole_length_rear_+hut_hole_length_/2.));
 
+    G4LogicalVolume* vacuum_logic = new G4LogicalVolume(vacuum_solid, vacuum, "EP_HOLE");
 
     /// Sapphire window ///
-    G4Tubs* sapphire_window_solid =
-      new G4Tubs("SAPPHIRE_WINDOW", 0., hole_diam_front_/2.,
-		 sapphire_window_thickn_/2., 0., twopi);
+    G4Tubs* sapphire_window_solid = new G4Tubs("SAPPHIRE_WINDOW", 0., hole_diam_front_/2.,
+                                               (sapphire_window_thickn_ + tpb_thickn_)/2., 0., twopi);
```
MiryamMV, line 247: I don't understand why `tpb_thickn_` is added to the `sapphire_window_thickn_` when creating the `sapphire_window_solid` and later adding another `G4Tubs` for the TPB outside this volume (according to line 262).

gondiaz: That's because the `tpb_logic` is placed inside the `sapphire_window_logic` instead of been placed inside `vacuum_logic`. I just maintained the convention used before rewiewing this class, but it would be equivalent to place the `tpb_logic` inside the `vacuum_logic` and deleting the `tpb_thickn_` in the definition of `sapphire_window_solid`.

source/geometries/Next100EnergyPlane.cc
gondiaz: Solved. I don't even know why this hole is implemented but it might be by historical reasons, its contribution seems irrelevant.

source/geometries/Next100EnergyPlane.cc
```diff
@@ -435,17 +387,10 @@ namespace nexus {
                                    0., twopi, nullptr,
```
MiryamMV, line 388: Wouldn't this way of defining the copper generator generate vertices in places where there is no copper? We could probably use ´copper_gen_=new CylinderPointSampler2020(copper_plate_solid)` instead.

gondiaz: In order to avoid generating in the holes, we use the `global_vertex` check below in the do-while statment. Using `CylinderPointSampler2020` would produce the same result.

MiryamMV (approved):

PR #121, opened 2021-07-28
Next100 tracking plane

gondiaz:
This PR adapts the Next100 tracking plane. This first version of the PR adds the following:
- Change generic photosensor SiPMs to Next100SiPMs
- Fix copper plate thicknesses.
- Fix SIPM_BOARD vertex which was wrongly named to DICE_BOARD in Next100 class
- Add DB-PLUGS. In the past they were included in the ICS class.
- Fix dimensions according to last design: [gate-tp-distance.pdf](https://github.com/next-exp/nexus/files/7337219/02.pdf)
Notice that the anode ring thickness (`gate_ring_thickn_`) is set to 9.9 mm in order to avoid its overlap with the square sipm boards masks at high radious. By simplicity, the gate ring thickness is set to the same value, therefore the parameter `gate_teflon_dist_` is now refered to the gate-grid instead to the gate ring as before.

The final design is still not decided thus the ultimate sipm holes implementation is left unchanged for the moment. It will likely be one of these:
- Mask holes covered by a coated membrane. No internal hole walls or SiPMs coated (DEMO Run 9).
- Mask internal hole walls coated with coated SIPMs and no membrane (DEMO Run 10).

TO-DO:
- Likely change rounded hole to square hole. But waiting for confirmation and dimensions.

### DIFF
--- a/source/geometries/Next100.cc
+++ b/source/geometries/Next100.cc
@@ -38,8 +38,9 @@ namespace nexus {
     lab_size_ (5. * m),
 
     // common used variables in geomety components
-    gate_tracking_plane_distance_(30. * mm), // to be confirmed
-    gate_sapphire_wdw_distance_  (1460.5 * mm),
+    // 0.1 mm gate thickness?
+    gate_tracking_plane_distance_((26.1 + 0.1)   * mm),
+    gate_sapphire_wdw_distance_  ((1458.2 - 0.1) * mm),
 
     specific_vertex_{},
     lab_walls_(false)
--- a/source/geometries/Next100.cc
+++ b/source/geometries/Next100.cc
@@ -221,7 +221,7 @@ namespace nexus {
              (region == "PMT") ||
              (region == "PMT_BASE") ||
              (region == "TP_COPPER_PLATE") ||
-             (region == "DICE_BOARD") ||
+             (region == "SIPM_BOARD") ||
              (region == "EL_TABLE") ||
              (region == "FIELD_RING") ||
              (region == "GATE_RING") ||
--- a/source/geometries/Next100.cc
+++ b/source/geometries/Next100.cc
@@ -38,7 +38,8 @@ namespace nexus {
     lab_size_ (5. * m),
 
     // common used variables in geomety components
-    // 0.1 mm gate thickness?
+    // 0.1 mm grid thickness
+    // note that if grid thickness change it must be also changed in Next100FieldCage.cc
     gate_tracking_plane_distance_((26.1 + 0.1)   * mm),
     gate_sapphire_wdw_distance_  ((1458.2 - 0.1) * mm),
 
--- a/source/geometries/Next100.cc
+++ b/source/geometries/Next100.cc
@@ -223,6 +223,7 @@ namespace nexus {
              (region == "PMT_BASE") ||
              (region == "TP_COPPER_PLATE") ||
              (region == "SIPM_BOARD") ||
+             (region == "DB_PLUG") ||
              (region == "EL_TABLE") ||
              (region == "FIELD_RING") ||
              (region == "GATE_RING") ||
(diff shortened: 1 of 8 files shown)


source/geometries/Next100.cc
```diff
@@ -38,8 +38,9 @@ namespace nexus {
     lab_size_ (5. * m),
 
     // common used variables in geomety components
-    gate_tracking_plane_distance_(30. * mm), // to be confirmed
-    gate_sapphire_wdw_distance_  (1460.5 * mm),
+    // 0.1 mm gate thickness?
```
MiryamMV: Do you mean grid thickness?

source/geometries/Next100.cc
```diff
@@ -38,8 +38,9 @@ namespace nexus {
...
+    gate_tracking_plane_distance_((26.1 + 0.1)   * mm),
```
MiryamMV, line 43: Does this take into account the reduction of the gate and anode rings?

source/geometries/Next100.cc
```diff
@@ -38,8 +38,9 @@ namespace nexus {
...
+    gate_sapphire_wdw_distance_  ((1458.2 - 0.1) * mm),
```
MiryamMV, line 44: Same here.

source/geometries/Next100.cc
gondiaz: Right. I have modified the comment to clarify.

source/geometries/Next100.cc
gondiaz: Yes, the measurement is performed from the grid position at the end of the gate ring, therefore the ring thickness is assumed to be the real one instead of the modified value that we use in the simulation in order to avoid overlap with the sipm boards.

source/geometries/Next100SiPMBoard.cc
```diff
@@ -32,18 +32,18 @@ using namespace nexus;
...
   time_binning_    (1. * microsecond),
   visibility_      (true),
   sipm_visibility_ (false),
   mpv_             (nullptr),
   vtxgen_          (nullptr),
-  sipm_            (new GenericPhotosensor("SiPM", 1.3 * mm))
+  sipm_            (new Next100SiPM())
 {
   msg_ = new G4GenericMessenger(this, "/Geometry/Next100/",
                                 "Control commands of the NEXT-100 geometry.");
```
MiryamMV, line 49: Change `mpv_->GetLogicalVolume()->GetMaterial()` in line 162 to `mother_gas` defined in line 81.

gondiaz: Solved

gondiaz:
I have changed the holes to:
- Hole walls coated with coated SIPMs and no membrane (DEMO Run 10).

source/geometries/Next100TrackingPlane.cc
```diff
@@ -61,69 +66,74 @@ void Next100TrackingPlane::Construct()
...
+  G4ThreeVector pos;
+  for (int i=0; i<int(board_pos_.size()); i++) {
+    pos = board_pos_[i];
+    pos.setY(pos.getY()-plug_y_displacement);
+    pos.setZ(plug_posz);
+    plug_pos_.push_back(pos);
+    new G4PVPlacement(0, pos, plug_logic, "DB_PLUG", mpv_->GetLogicalVolume(), false, i, false);
+  }
 
-  // Column on the far left has 5 boards.
```
paolafer, line 96: Is there any reason why you removed the comments? I think they are useful.

gondiaz: Probably I though that the function calls were clear enough like `PlaceSiPMBoardColumns(5, -3.5, zpos, board_index, sipm_board_logic)` is already saying that you place 5 boards and the position. However if you think is clearer with the comments I can recover them.

paolafer: You need to go to the function and understand what it does. I think the comments help.

gondiaz: Previous comments added

MiryamMV (approved):
This PR includes everything listed. I approve it. Great job!

PR #122, opened 2021-09-06
Next100 ICS

gondiaz:
This PR review the Next-100 ICS. The ICS was almost completely rewritted, since the current designs are different from what was implemented before. As usual, the dimensions in the code are obtained from 2D if possible, if not they are meassured in the 3D drawings. 2D drawings provided by Jordi can be found below.
[copper_bars.pdf](https://github.com/next-exp/nexus/files/7115666/01-00.DRAWINGS.OF.COPPER.BARS.1.TO.7.pdf)
[dimensions.png](https://user-images.githubusercontent.com/35993868/132211404-7317d6cb-e3d6-4e80-b297-c40e528ff56c.png)

The changes are substantial in the ICS implementation:
- Delete copper plate from the energy and tracking planes and corresponding nozzle holes. The copper plates are already included in its respective classes.
- Remove dice board plugs. These will be likely included in the tracking plane PR.
- Implement ICS holes in calibration ports and feedthroughts. Also extend the calibration ports inside the copper plate througout the hole (see vessel PR slides).
- Implement vertex generator for the new ICS.

Importantly since it overlaps with PR #118: `gate_tp_distance_` variable and field-cage length are needed to built the ICS. The first variable is set as global in `Next100.cc` and setted (`SetELtoTPdistance`) in `ics_` instance. The same should be made for `inner_elements_` to be consistent. For the field-cage-length a getter would be required from `inner_elements_`, its value is preeliminary hardcoded.

### DIFF
--- a/source/geometries/Next100.cc
+++ b/source/geometries/Next100.cc
@@ -146,6 +146,7 @@ namespace nexus {
 
     // INNER COPPER SHIELDING
     ics_->SetLogicalVolume(vessel_internal_logic);
+    ics_->SetELzCoord(gate_zpos_in_vessel_);
     ics_->Construct();
 
     G4ThreeVector gate_pos(0., 0., -gate_zpos_in_vessel_);
--- a/source/geometries/Next100.cc
+++ b/source/geometries/Next100.cc
@@ -39,12 +39,17 @@ namespace nexus {
     GeometryBase(),
     // Lab dimensions
     lab_size_ (5. * m),
+
+    // distance between EL gate and TP copper plate
+    gate_tp_distance_(25.6 * mm), // check
+
     // Nozzles external diam and y positions
     nozzle_ext_diam_ (9. * cm),
     up_nozzle_ypos_ (20. * cm),
     central_nozzle_ypos_ (0. * cm),
     down_nozzle_ypos_ (-20. * cm),
     bottom_nozzle_ypos_(-53. * cm),
+
     specific_vertex_{},
     lab_walls_(false)
   {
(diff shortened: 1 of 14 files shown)


source/geometries/Next100.cc
```diff
@@ -143,10 +148,17 @@ namespace nexus {
...
+    inner_elements_->SetELtoSapphireWDWdistance(gate_sapphire_wdw_distance_);
+    inner_elements_->SetELtoTPdistance         (gate_tracking_plane_distance_);
     inner_elements_->Construct();
 
     // INNER COPPER SHIELDING
     ics_->SetLogicalVolume(vessel_internal_logic);
+    ics_->SetELzCoord(gate_zpos_in_vessel_);
+    ics_->SetELtoSapphireWDWdistance(gate_sapphire_wdw_distance_);
+    ics_->SetELtoTPdistance         (gate_tracking_plane_distance_);
+    ics_->SetFieldCageLength(1432.*mm + 16.*mm); // gate-teflon distance + teflon length
```
MiryamMV: Should we consider the anode in this length as it is also part of the field cage class?

source/geometries/Next100Ics.cc
```diff
@@ -96,264 +57,141 @@ namespace nexus {
 
   void Next100Ics::Construct()
   {
+    G4double length = gate_tp_distance_ + gate_sapphire_wdw_dist_;
+
+    G4double offset = 1.* mm;
```
MiryamMV, line 61: Maybe we should add an explanation here of the use of this variable (for future developers)

MiryamMV (approved):

source/geometries/Next100Ics.cc
```diff
@@ -96,264 +57,143 @@ namespace nexus {
...
+                port_b_Rot, G4ThreeVector(-port_x, port_y, lat_hole_z-length/2.));
 
-    ics_solid = new G4UnionSolid("ICS", ics_solid, ics_energy_sph_solid,
-				 0, G4ThreeVector(0., 0., energy_sph_zpos_) );
 
-    G4double energy_cyl_zpos = -1. * (body_length_/2. + energy_cyl_length_/2.);
-    ics_solid = new G4UnionSolid("ICS", ics_solid, ics_energy_cyl_solid,
-				 0, G4ThreeVector(0., 0., energy_cyl_zpos) );
+    /// ICS step at the TP end.
+    // This avoids overlap with sipm boards at high radious
```
paolafer: radious --> radius

paolafer:
You can remove the following `include` from the `Next100` and `Next110Ics` classes:

```
#include <CLHEP/Units/SystemOfUnits.h>
#include <CLHEP/Units/PhysicalConstants.h>
#include <stdexcept>
```

The first two were needed in some old version of GEANT4, but not anymore. The third one is probably just a copy-paste from one class to another.

source/geometries/Next100Ics.cc
gondiaz: Solved in the commit that removes the includes

PR #123, opened 2021-09-07
Set gate z position in Next100 shielding

gondiaz:
This PR solves a buggy behaviour in the vertex generation of SHIELDING_LEAD and SHIELDING_STEEL due to the need of setting the z-position of the EL gate. Also corrected the INNER_AIR generator, and added some cosmetic changes.


paolafer (approved):
The code works correctly, the vertices are generated in the appropriate place, I approve it.

PR #124, opened 2021-09-14
Fix test code following latest changes in xenon properties

paolafer:
This PR fixes a problem with the code of a test, which had not been adapted to the latest changes in the material-related classes. The issue had gone unnoticed so far because that test was not included in the `scons` compilation. This PR also fixes the compilation.


MiryamMV:
Works for scans and make. Approved!

PR #127, opened 2021-11-11
Upgrade package version to solve building error with Nix.

paolafer:
This PR tries to fix an error occurring with Nix by upgrading two packages.


jmbenlloch (approved):
This update fixes the automated tests.

PR #128, opened 2021-12-17
Delete BatchSession class

gondiaz:
Motivation:
trying to compile NEXUS with Geant4 11, I found a first compilation error in BatchSession.cc, since G4String.cc class was modified in this new G4 version.

Since BatchSession is a difficult module, I propose to delete it and redefine NexusApp::ExecuteMacroFile as shown.


jmalbos:
We added that class to modify the default behavior of Geant4, stopping the initialization if an unknown parameter is found in a macro file. I'll have a look at the error you mention to see if there's an easy way to fix it rather than eliminating the class altogether.

gondiaz:
Ok, I think it can be modified instead, I close this PR.

PR #129, opened 2022-01-03
Add port tube tip

gondiaz:
This PR adds missing tip at the end of the port tubes, that was not implemented in the previous version. It may be relevant to account for the possible interactions at the tip for the calibration photons.

Port_Tube

Port_Tip

### DIFF
--- a/source/geometries/Next100Vessel.cc
+++ b/source/geometries/Next100Vessel.cc
@@ -58,9 +58,10 @@ namespace nexus {
 
     // Ports
     port_base_height_(37. * mm),
-    port_tube_height_(154.* mm),
+    port_tube_height_(159.* mm),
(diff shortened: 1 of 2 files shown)


halmamol:
I have reviewed the dimensions with @gondiaz offline. Just for the record, the dimensions quoted on the PR (178mm) are from the red line in the picture below, until the end of the tube. The whole tube is not simmulated to avoid overlaps with the nozzle. That makes the `port_tube_height` = 178mm - 15mm (from nozzle to vessel) - 4mm (from the tip) = 159mm 

imagen

halmamol (approved):
This PR adds a missing tip at the end of the calibration port of ~4mm. Tube dimensions have been reviewed, and are accordingly implemented. I approve it!

PR #130, opened 2022-01-04
Next100 field rings vertex

gondiaz:
This PR adds the vertex generator for the field-rings. As shown in the current [activities sheet](https://docs.google.com/spreadsheets/d/1hhyX4tDhYrJ7cFDp9pPROMCCYH5xY-EoAlkQplRNtcI/edit?usp=sharing) we should add this vertex. Also, the volume is renamed from RING to FIELD_RING, which must be taken into account in the G4 Volume name in the sheet.

### DIFF
--- a/source/geometries/Next100.cc
+++ b/source/geometries/Next100.cc
@@ -230,7 +230,8 @@ namespace nexus {
 	     (region == "TP_COPPER_PLATE") ||
 	     (region == "DICE_BOARD") ||
 	     (region == "AXIAL_PORT") ||
-	     (region == "EL_TABLE") ) {
+	     (region == "EL_TABLE") ||
+       (region == "FIELD_RING")) {
       vertex = inner_elements_->GenerateVertex(region);
     }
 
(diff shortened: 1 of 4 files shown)


source/geometries/Next100InnerElements.cc
```diff
@@ -97,11 +97,12 @@ namespace nexus {
 
     // Field Cage regions
     if ((region == "CENTER") ||
-	(region == "ACTIVE") ||
-	(region == "BUFFER") ||
-	(region == "XENON") ||
-  (region == "EL_GAP") ||
-	(region == "LIGHT_TUBE")) {
+	      (region == "ACTIVE") ||
```
paolafer: It seems that there is some misalignment here, can you fix it?

source/geometries/Next100.cc
```diff
@@ -230,7 +230,8 @@ namespace nexus {
 	     (region == "TP_COPPER_PLATE") ||
 	     (region == "DICE_BOARD") ||
 	     (region == "AXIAL_PORT") ||
-	     (region == "EL_TABLE") ) {
+	     (region == "EL_TABLE") ||
+       (region == "FIELD_RING")) {
```
paolafer: Fix the misalignment.

paolafer (approved):
This PR adds vertices to the volume of the field rings, which are necessary for the NEXT-100 background model. Approved.

PR #133, opened 2022-01-31
Use C++-17 in nexus and avoid warnings.

paolafer:
This PR modifies the SConstruct file to use the 17 standard of C++. This way, a lot of warnings coming from C++14 and C++17 extensions used in GEANT4 are removed.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -183,7 +183,7 @@ vars.AddVariables(
 
     ('CXXFLAGS',
      'c++ compiler options.',
-     ['-std=c++11']),
+     ['-std=c++17']),
 
     ('CPPPATH',
      'List of directories where the include headers are located.',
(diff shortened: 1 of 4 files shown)


gondiaz:
Can you also silence the following warnings? Both were missing at #75
```
Compiling source/generators/Decay0Interface.cc
source/generators/Decay0Interface.cc:202:16: warning: ‘contains’ is deprecated: Use G4StrUtil::contains instead
      [-Wdeprecated-declarations]
  while (!line.contains(“First event”)) getline(file_, line);
               ^
/Users/gonzalo/Software/Geant4_11/install/include/Geant4/G4String.hh:116:5: note: ‘contains’ has been explicitly marked
      deprecated here
  [[deprecated(“Use G4StrUtil::contains instead”)]]
    ^
1 warning generated.
```
```
Compiling source/sensdet/SensorSD.cc
In file included from source/sensdet/SensorSD.cc:10:
source/sensdet/SensorSD.h:77:26: warning: private field ‘boundary_’ is not used [-Wunused-private-field]
    G4OpBoundaryProcess* boundary_; ///< Pointer to the optical boundary process
                         ^
1 warning generated.
```

gondiaz (approved):
This PR does what it claims, approved!

On the other hand this is a patch to the SConstruct file, which I think deserves a full review since, for example:
- The defined PathVariables like GEANT4_BINDIR are set to a NULL_PATH instead of being read as external env variable (which I think should be the expected behaviour).
- There is not check for hdf5 libraries, and probably setting HDF5_LIB and HDF5_INC is not neccessary (a single HDF5_DIR path could be enough, which is in fact defined but set to NULL_PATH).

paolafer:
Ok, we can open a new PR with the changes you suggest.

PR #134, opened 2022-02-01
Adding fix to catch.hh to fix compilation error on M1 macbook

kvjmistry:
This PR edits the catch.hh file to enable compilation of nexus with mac M1. It may also be possible to update catch2 to a newer version to have the same effect.

### DIFF
--- a/source/tests/catch.hpp
+++ b/source/tests/catch.hpp
@@ -1,9 +1,9 @@
 /*
- *  Catch v2.7.1
- *  Generated: 2019-04-05 18:22:37.720122
+ *  Catch v2.13.8
+ *  Generated: 2022-01-03 21:20:09.589503
  *  ----------------------------------------------------------
  *  This file has been merged from multiple headers. Please don't edit it directly
- *  Copyright (c) 2019 Two Blue Cubes Ltd. All rights reserved.
+ *  Copyright (c) 2022 Two Blue Cubes Ltd. All rights reserved.
  *
  *  Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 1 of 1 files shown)


jmalbos:
Hi @kvjmistry,

> It may also be possible to update catch2 to a newer version to have the same effect.

Did you actually try this?

paolafer:
I've tried using the `v2.13.8` version (the latest stable one, it seems), and it seems to work alright. Shall we use this one, then?

kvjmistry:
I updated the file to v2.13.8 and compiled no problem too. The latest commit includes this new file.

gondiaz:
Works to me in Intel Mac

jmalbos:
OK, so just to be sure: we would just need this second commit, right? If that is the case, you could remove the first one from the pull request or squash both together. You should also amend the commit comment to comply with the following conventions: https://cbea.ms/git-commit/
It could be simply something like this: _"Update Catch2 header to v2.13.8"_

kvjmistry:
OK so I have removed the first commit and have everything as one commit which just replaces the header file to the new version.  The single commit has been renamed to "Update catch2 header to v2.13.8" (apologies for the poor name before). 

Please let me know if this is not the case!

paolafer:
That's great, Krishan, I'll approve and merge the PR as soon as the tests have passed. I'll add the link to the git conventions in the wiki, by the way!

jmalbos:
Great, @kvjmistry, and congrats for your first approved nexus PR!

paolafer (approved):
Catch2 header file has been updated to fix error compiling on M1 chip. Approved!

PR #135, opened 2022-02-03
Review of SConstruct

paolafer:
This PR makes a revision of the SConstruct file, used to compile nexus.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -155,6 +155,12 @@ vars.AddVariables(
                  'Path to gsl bin directory.',
                  NULL_PATH),
 
+    ## hdf5
+
+    PathVariable('HDF5_DIR',
+                 'Path to hdf5 installation directory.',
+                 NULL_PATH),
+
     ## installation directory
 
     PathVariable('PREFIX',
@@ -269,14 +275,21 @@ if not env['LIBPATH']:
 
     ## HDF5 configuration ----------------------------------
 
-    try:
-        env['HDF5_LIB']     = os.environ['HDF5_LIB']
-        env.Append( LIBPATH = [env['HDF5_LIB']] )
-        env.Append(LIBS     = ['hdf5'])
-        env['HDF5_INC']     = os.environ['HDF5_INC']
-        env.Append( CPPPATH = [env['HDF5_INC']] )
-    except KeyError:
-        Abort('HDF5 environment variables could not be found.')
+    if env['HDF5_DIR'] == NULL_PATH:
+        try:
+            env['HDF5_DIR'] = os.environ['HDF5_DIR']
+        except KeyError:
+            Abort('HDF5 installation directory could not be found.')
+
+    env.Append(LIBPATH = [env['HDF5_DIR']+'/lib'])
+    env.Append(LIBS    = ['hdf5'])
+    env.Append(CPPPATH = [env['HDF5_DIR']+'/include'])
+
+    if not conf.CheckCXXHeader('hdf5.h'):
+        Abort('HDF5 headers not found.')
+
+    if not conf.CheckLib(library='hdf5', language='CXX', autoadd=0):
+        Abort('HDF5 library not found.')
 
 ## ##################################################################
 
--- a/SConstruct
+++ b/SConstruct
@@ -233,6 +233,12 @@ if not env['LIBPATH']:
 
     env.ParseConfig('geant4-config --cflags --libs')
 
+    if not conf.CheckCXXHeader('G4Event.hh'):
+        Abort('Geant4 headers could not be found.')
+
+    if not conf.CheckLib(library='G4global', language='CXX', autoadd=0):
+        Abort('Geant4 libraries could not be found.')
+
 
     ## ROOT configuration ----------------------------------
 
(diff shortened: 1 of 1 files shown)


gondiaz (changes requested):
Remove
`from __future__ import print_function`

SConstruct
```diff
@@ -136,31 +136,33 @@ vars = Variables(BUILDVARS_FILE)
...
                  NULL_PATH),
 
-    ## gsl
+    ## hdf5
 
-    PathVariable('GSL_DIR',
-                 'Path to gsl installation.',
+    PathVariable('HDF5_DIR',
+                 'Path to hdf5 installation directory.',
                  NULL_PATH),
```
gondiaz, line 161: These variables are effectively not used. I would remove them

SConstruct
```diff
@@ -251,34 +257,40 @@ if not env['LIBPATH']:
...
-        env['HDF5_INC'] = os.environ['HDF5_INC']
-        env.Append( CPPPATH = [env['HDF5_INC']] )
-    except KeyError: pass
-
-
-    ## GSL configuration --------------------------   -------
+    ## GSL configuration ------------------------------------
 
-    if env['GSL_DIR'] != NULL_PATH:
+    if env['GSL_BINDIR'] != NULL_PATH:
```
gondiaz, line 259: Variables defined above are always equal to NULL_PATH since they are not passed as env variables. This kind of statements are never run. I would remove them.

SConstruct
```diff
@@ -209,7 +211,11 @@ vars.AddVariables(
 
 ## Create a construction environment adding the build vars and
 ## propagating the user's external environment
-env = Environment(variables=vars, ENV=os.environ)
+env = Environment(variables=vars)
+
+## Load the PATH variable of the user environment,
+## to look for third-party library paths
+env.PrependENVPath('PATH', os.environ['PATH'])
```
gondiaz, line 218: I would add something like:
```suggestion
try: 
    env.PrependENVPath("HDF5_DIR", os.environ["HDF5_DIR"])
except:
    Exception("HDF5_DIR must be defined")
```
Note that this defines the following entry in the `env` dictionary env["ENV"]["HDF5_DIR"], instead of env["HDF5_DIR"].

[Scons Documentation](https://www.scons.org/doc/production/HTML/scons-user.html#idm46358283013728), section 7.3

SConstruct
```diff
@@ -251,34 +257,40 @@ if not env['LIBPATH']:
...
 
-#    if not conf.CheckLib(library='GSL', language='CXX', autoadd=0):
-#        Abort('GSL library not found.')
+    ## HDF5 configuration ----------------------------------
+
+    if env['HDF5_DIR'] == NULL_PATH:
+        try:
+            env['HDF5_DIR'] = os.environ['HDF5_DIR']
+        except KeyError:
+            Abort('HDF5 installation directory could not be found.')
```
gondiaz, line 279: Following comment above these lines would be unnecessary

SConstruct
```diff
@@ -251,34 +257,40 @@ if not env['LIBPATH']:
...
+
+    env.Append(LIBPATH = [env['HDF5_DIR']+'/lib'])
+    env.Append(LIBS    = ['hdf5'])
+    env.Append(CPPPATH = [env['HDF5_DIR']+'/include'])
```
gondiaz, line 283: Following comment above, now `env["HDF5_DIR"]` would be `env["ENV"]["HDF5_DIR"]`

gondiaz (changes requested):

SConstruct
paolafer: They are used if the user passes those paths via the command line. I think it is a possibility that we should maintain.

SConstruct
paolafer: They are not, if the user passes those paths via command line. NULL_PATH is just the default value.

SConstruct
paolafer: I would leave it as it is. This way, it is possible to pass the HDF5 path via command line, like the other third-party libraries.

SConstruct
paolafer: See above.

SConstruct
paolafer: See above.

SConstruct
gondiaz: Ok, I see, I wasn't aware of this functionality. I thought the typical user case is to define them as environmental variables.

gondiaz:
For cosmetics, I would write the library checks of Geant4 just above the parser. Same for Root libs.

jmalbos (commented):
I have tried building this branch in my laptop (macOS 12.2, clang 13.0.0). It does work, but I got the warning below multiple times. 

```
G4String.hh:127:5: warning: use of the 'nodiscard' attribute is a C++17 extension [-Wc++17-extensions]
  [[nodiscard]] inline G4String strip(stripType strip_Type = trailing, char c = ' ');
    ^
1 warning generated.
```

paolafer:
> I have tried building this branch in my laptop (macOS 12.2, clang 13.0.0). It does work, but I got the warning below multiple times.

It's strange, the warning seems to complain about the C++17 standard, but `scons` should be compiling with C++17. Can you check the value of the `CXXFLAGS` variable stored in the `buildvars.scons` file?

paolafer:
I have added the removal of C++14, in case that was the problem, try it now.

paolafer:
> I have added the removal of C++14, in case that was the problem, try it now.

@jmalbos, have you had the chance to try it after this modification?

jmalbos:
> > I have added the removal of C++14, in case that was the problem, try it now.
> 
> @jmalbos, have you had the chance to try it after this modification?

I just did and, indeed, the warnings disappeared.

jmalbos (approved):

gondiaz (approved):

PR #138, opened 2022-02-10
Use default 293 K isotherm for Xe (gas)

gondiaz:
This PR modifies the pressure-density values of Xe (gas) to the default 293 K isotherm. The values are taken from the reference data file in the nexus repository.

In the near future the desired behaviour will be to have an unique temperature-pressure-density function instead on relying on hard-coded isotherms.


paolafer (approved):
This PR changes the operation temperature of the NEXT detectors, therefore the gas density. Approved!

PR #141, opened 2022-02-24
Solves G4 photon simulation issue

gondiaz:
In this PR:
- Remove ANODE-GAS volume, since the G4LogicalBorderSurface is the interface between the mother-gas (vessel-gas) and the tpb of the boards.
- Add an empty G4MaterialPropertiesTable to the teflon G4Material to avoid segmentation fault bug in G4, which is reported in [bugzilla-geant4-2471](https://bugzilla-geant4.kek.jp/show_bug.cgi?id=2471)

### DIFF
--- a/source/geometries/Next100FieldCage.cc
+++ b/source/geometries/Next100FieldCage.cc
@@ -256,6 +256,9 @@ void Next100FieldCage::DefineMaterials()
 
   /// Teflon for the light tube
   teflon_ = G4NistManager::Instance()->FindOrBuildMaterial("G4_TEFLON");
+  // teflon is the material used in the light-tube, and is covered by a G4LogicalSkinSurface
+  // In Geant4 11.0.0, a bug in treating the OpBoundaryProcess produced in the surface makes the code fail.
+  // This is avoided by setting an empty G4MaterialPropertiesTable of the G4Material.
   teflon_->SetMaterialPropertiesTable(new G4MaterialPropertiesTable());
 
   /// TPB coating
--- a/source/geometries/Next100FieldCage.cc
+++ b/source/geometries/Next100FieldCage.cc
@@ -255,8 +255,8 @@ void Next100FieldCage::DefineMaterials()
   copper_ = G4NistManager::Instance()->FindOrBuildMaterial("G4_Cu");
 
   /// Teflon for the light tube
-  teflon_ =
-  G4NistManager::Instance()->FindOrBuildMaterial("G4_TEFLON");
+  teflon_ = G4NistManager::Instance()->FindOrBuildMaterial("G4_TEFLON");
+  teflon_->SetMaterialPropertiesTable(new G4MaterialPropertiesTable());
 
   /// TPB coating
   tpb_ = materials::TPB();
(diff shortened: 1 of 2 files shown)


source/geometries/Next100SiPMBoard.cc
```diff
@@ -105,10 +105,11 @@ void Next100SiPMBoard::Construct()
   G4Box* mask_solid_vol =
     new G4Box(mask_name, size_/2., size_/2., mask_thickness_/2.);
 
+  G4Material* teflon = G4NistManager::Instance()->FindOrBuildMaterial("G4_TEFLON");
+  teflon->SetMaterialPropertiesTable(new G4MaterialPropertiesTable());
```
jmalbos, line 112: Same as below.

source/geometries/Next100FieldCage.cc
```diff
@@ -255,8 +255,8 @@ void Next100FieldCage::DefineMaterials()
   copper_ = G4NistManager::Instance()->FindOrBuildMaterial("G4_Cu");
 
   /// Teflon for the light tube
-  teflon_ =
-  G4NistManager::Instance()->FindOrBuildMaterial("G4_TEFLON");
+  teflon_ = G4NistManager::Instance()->FindOrBuildMaterial("G4_TEFLON");
+  teflon_->SetMaterialPropertiesTable(new G4MaterialPropertiesTable());
```
jmalbos, line 262: Perhaps you should add here a comment explaining why we are adding an empty `G4MaterialPropertiesTable`. It may be helpful for future reference.

jmalbos (changes requested):

source/geometries/Next100FieldCage.cc
```diff
@@ -255,8 +255,11 @@ void Next100FieldCage::DefineMaterials()
   copper_ = G4NistManager::Instance()->FindOrBuildMaterial("G4_Cu");
 
   /// Teflon for the light tube
-  teflon_ =
-  G4NistManager::Instance()->FindOrBuildMaterial("G4_TEFLON");
+  teflon_ = G4NistManager::Instance()->FindOrBuildMaterial("G4_TEFLON");
+  // teflon is the material used in the light-tube, and is covered by a G4LogicalSkinSurface
+  // In Geant4 11.0.0 there is a bug in treating the OpBoundaryProcess produced in the surface,
+  // which is avoided by setting the G4MaterialPropertiesTable of the G4Material
```
paolafer: It's probably a bit pedantic, but I would be even more explicit: "In Geant4 11.0.0, a bug in treating the OpBoundaryProcess produced in the surface makes the code fail. This is avoided by setting an empty G4MaterialPropertiesTable of the G4Material.". What is avoided is the crash of the code, not the bug.

source/geometries/Next100FieldCage.cc
gondiaz: Done

source/geometries/Next100FieldCage.cc
gondiaz: Done

paolafer (approved):
This PR overcomes a bug of Geant4 and eliminates an unnecessary volume, which caused a misbehaviour for the optical photons. Approved!

jmalbos (approved):

PR #143, opened 2022-03-01
Destruct explicitly the pointer to the generator

paolafer:
This PR adds an explicit call to the destructor of the generator. This way, the destructor of the generator class is called when the run manager is deleted.

### DIFF
--- a/source/base/DetectorConstruction.h
+++ b/source/base/DetectorConstruction.h
@@ -9,14 +9,13 @@
 #ifndef DETECTOR_CONSTRUCTION_H
 #define DETECTOR_CONSTRUCTION_H
 
-#include "GeometryBase.h"
-
 #include <G4VUserDetectorConstruction.hh>
 
 class G4GenericMessenger;
 
-
 namespace nexus {
+
+  class GeometryBase;
 
   class DetectorConstruction: public G4VUserDetectorConstruction
   {
--- a/source/base/DetectorConstruction.h
+++ b/source/base/DetectorConstruction.h
@@ -9,14 +9,14 @@
 #ifndef DETECTOR_CONSTRUCTION_H
 #define DETECTOR_CONSTRUCTION_H
 
+#include "GeometryBase.h"
+
 #include <G4VUserDetectorConstruction.hh>
 
 class G4GenericMessenger;
 
 
 namespace nexus {
-
-  class GeometryBase;
 
   class DetectorConstruction: public G4VUserDetectorConstruction
   {
@@ -31,22 +31,22 @@ namespace nexus {
     virtual G4VPhysicalVolume* Construct();
 
     /// Set a detector geometry
-    void SetGeometry(GeometryBase*);
+    void SetGeometry(std::unique_ptr<GeometryBase>);
     /// Get the detector geometry
     const GeometryBase* GetGeometry() const;
 
   private:
-    GeometryBase* geometry_;
+    std::unique_ptr<GeometryBase> geometry_;
   };
 
 
   // INLINE DEFINITIONS /////////////////////////////////////////////
 
-  inline void DetectorConstruction::SetGeometry(GeometryBase* g)
-  { geometry_ = g; }
+  inline void DetectorConstruction::SetGeometry(std::unique_ptr<GeometryBase> g)
+  { geometry_ = std::move(g); }
 
   inline const GeometryBase* DetectorConstruction::GetGeometry() const
-  { return geometry_; }
+  { return geometry_.get(); }
 
 } // end namespace nexus
 
--- a/source/base/NexusApp.cc
+++ b/source/base/NexusApp.cc
@@ -154,9 +154,7 @@ NexusApp::NexusApp(G4String init_macro): G4RunManager(), gen_name_(""),
 NexusApp::~NexusApp()
 {
   // Close output file before finishing
-  PersistencyManagerBase* current = dynamic_cast<PersistencyManagerBase*>
-    (G4VPersistencyManager::GetPersistencyManager());
-  current->CloseFile();
+  pm_->CloseFile();
 }
 
 
@@ -208,7 +206,7 @@ void NexusApp::ExecuteMacroFile(const char* filename)
   std::unique_ptr<G4UIsession> batchSession =
     std::make_unique<BatchSession>(filename, UI->GetSession());
   UI->SetSession(batchSession.get());
-  G4UIsession* previousSession = UI->GetSession()->SessionStart();
+  G4UIsession* previousSession = batchSession->SessionStart();
   UI->SetSession(previousSession);
 }
 
--- a/source/base/NexusApp.cc
+++ b/source/base/NexusApp.cc
@@ -99,7 +99,7 @@ NexusApp::NexusApp(G4String init_macro): G4RunManager(), gen_name_(""),
   if (gen_name_ == "") {
     G4Exception("[NexusApp]", "NexusApp()", FatalException, "A generator must be specified.");
   }
-  pg->SetGenerator(ObjFactory<G4VPrimaryGenerator>::Instance().CreateObject(gen_name_));
+  pg->SetGenerator(std::unique_ptr<G4VPrimaryGenerator>(ObjFactory<G4VPrimaryGenerator>::Instance().CreateObject(gen_name_)));
   this->SetUserAction(pg);
 
   if (pm_name_ == "") {
--- a/source/base/NexusApp.cc
+++ b/source/base/NexusApp.cc
@@ -10,9 +10,6 @@
 
 #include "NexusApp.h"
 
-#include "DetectorConstruction.h"
-#include "PrimaryGeneration.h"
-#include "BatchSession.h"
 #include "FactoryBase.h"
 
 #include <G4GenericPhysicsList.hh>
@@ -36,7 +33,7 @@ NexusApp::NexusApp(G4String init_macro): G4RunManager(), gen_name_(""),
                                          stkact_name_("")
 {
   // Create and configure a generic messenger for the app
-  msg_ = new G4GenericMessenger(this, "/nexus/", "Nexus control commands.");
+  msg_ = std::make_unique<G4GenericMessenger>(this, "/nexus/", "Nexus control commands.");
 
   // Define the command to register a configuration macro.
   // The user may invoke the command as many times as needed.
@@ -77,29 +74,29 @@ NexusApp::NexusApp(G4String init_macro): G4RunManager(), gen_name_(""),
   // by the time we process the initialization macro.
 
   // The physics lists are handled with Geant4's own 'factory'
-  physicsList = new G4GenericPhysicsList();
+  auto pl = std::make_unique<G4GenericPhysicsList>();
 
-  BatchSession* batch = new BatchSession(init_macro.c_str());
-  batch->SessionStart();
+  batch_ = std::make_unique<BatchSession>(init_macro.c_str());
+  batch_->SessionStart();
 
   // Set the physics list in the run manager
-  this->SetUserInitialization(physicsList);
+  this->SetUserInitialization(pl.release());
 
   // Set the detector construction instance in the run manager
-  DetectorConstruction* dc = new DetectorConstruction();
+  dc_ = std::make_unique<DetectorConstruction>();
   if (geo_name_ == "") {
     G4Exception("[NexusApp]", "NexusApp()", FatalException, "A geometry must be specified.");
   }
-  dc->SetGeometry(ObjFactory<GeometryBase>::Instance().CreateObject(geo_name_));
-  this->SetUserInitialization(dc);
+  dc_->SetGeometry(ObjFactory<GeometryBase>::Instance().CreateObject(geo_name_));
+  this->SetUserInitialization(dc_.release());
 
   // Set the primary generation instance in the run manager
-  PrimaryGeneration* pg = new PrimaryGeneration();
+  pg_ = std::make_unique<PrimaryGeneration>();
   if (gen_name_ == "") {
     G4Exception("[NexusApp]", "NexusApp()", FatalException, "A generator must be specified.");
   }
-  pg->SetGenerator(ObjFactory<G4VPrimaryGenerator>::Instance().CreateObject(gen_name_));
-  this->SetUserAction(pg);
+  pg_->SetGenerator(ObjFactory<G4VPrimaryGenerator>::Instance().CreateObject(gen_name_));
+  this->SetUserAction(pg_.release());
 
   if (pm_name_ == "") {
     G4Exception("[NexusApp]", "NexusApp()", FatalException, "A persistency manager must be specified.");
@@ -152,8 +149,6 @@ NexusApp::~NexusApp()
   PersistencyManagerBase* current = dynamic_cast<PersistencyManagerBase*>
     (G4VPersistencyManager::GetPersistencyManager());
   current->CloseFile();
-
-  delete msg_;
 }
 
 
@@ -202,10 +197,10 @@ void NexusApp::Initialize()
 void NexusApp::ExecuteMacroFile(const char* filename)
 {
   G4UImanager* UI = G4UImanager::GetUIpointer();
-  G4UIsession* batchSession = new BatchSession(filename, UI->GetSession());
-  UI->SetSession(batchSession);
+  std::unique_ptr<G4UIsession> batchSession =
+    std::make_unique<BatchSession>(filename, UI->GetSession());
+  UI->SetSession(batchSession.get());
   G4UIsession* previousSession = UI->GetSession()->SessionStart();
-  delete UI->GetSession();
   UI->SetSession(previousSession);
 }
 
--- a/source/base/NexusApp.cc
+++ b/source/base/NexusApp.cc
@@ -12,7 +12,6 @@
 
 #include "DetectorConstruction.h"
 #include "PrimaryGeneration.h"
-#include "PersistencyManagerBase.h"
 #include "BatchSession.h"
 #include "FactoryBase.h"
 
@@ -105,8 +104,8 @@ NexusApp::NexusApp(G4String init_macro): G4RunManager(), gen_name_(""),
   if (pm_name_ == "") {
     G4Exception("[NexusApp]", "NexusApp()", FatalException, "A persistency manager must be specified.");
   }
-  std::unique_ptr<PersistencyManagerBase> pm = ObjFactory<PersistencyManagerBase>::Instance().CreateObject(pm_name_);
-  pm->SetMacros(init_macro, macros_, delayed_);
+  pm_ = ObjFactory<PersistencyManagerBase>::Instance().CreateObject(pm_name_);
+  pm_->SetMacros(init_macro, macros_, delayed_);
 
  // PersistencyManager::Initialize(init_macro, macros_, delayed_);
 
--- a/source/base/NexusApp.cc
+++ b/source/base/NexusApp.cc
@@ -10,6 +10,9 @@
 
 #include "NexusApp.h"
 
+#include "BatchSession.h"
+#include "DetectorConstruction.h"
+#include "PrimaryGeneration.h"
 #include "FactoryBase.h"
 
 #include <G4GenericPhysicsList.hh>
@@ -76,27 +79,27 @@ NexusApp::NexusApp(G4String init_macro): G4RunManager(), gen_name_(""),
   // The physics lists are handled with Geant4's own 'factory'
   auto pl = std::make_unique<G4GenericPhysicsList>();
 
-  batch_ = std::make_unique<BatchSession>(init_macro.c_str());
-  batch_->SessionStart();
+  auto batch = std::make_unique<BatchSession>(init_macro.c_str());
+  batch->SessionStart();
 
   // Set the physics list in the run manager
   this->SetUserInitialization(pl.release());
 
   // Set the detector construction instance in the run manager
-  dc_ = std::make_unique<DetectorConstruction>();
+  auto dc = std::make_unique<DetectorConstruction>();
   if (geo_name_ == "") {
     G4Exception("[NexusApp]", "NexusApp()", FatalException, "A geometry must be specified.");
   }
-  dc_->SetGeometry(ObjFactory<GeometryBase>::Instance().CreateObject(geo_name_));
-  this->SetUserInitialization(dc_.release());
+  dc->SetGeometry(ObjFactory<GeometryBase>::Instance().CreateObject(geo_name_));
+  this->SetUserInitialization(dc.release());
 
   // Set the primary generation instance in the run manager
-  pg_ = std::make_unique<PrimaryGeneration>();
+  auto pg = std::make_unique<PrimaryGeneration>();
   if (gen_name_ == "") {
     G4Exception("[NexusApp]", "NexusApp()", FatalException, "A generator must be specified.");
   }
-  pg_->SetGenerator(ObjFactory<G4VPrimaryGenerator>::Instance().CreateObject(gen_name_));
-  this->SetUserAction(pg_.release());
+  pg->SetGenerator(ObjFactory<G4VPrimaryGenerator>::Instance().CreateObject(gen_name_));
+  this->SetUserAction(pg.release());
 
   if (pm_name_ == "") {
     G4Exception("[NexusApp]", "NexusApp()", FatalException, "A persistency manager must be specified.");
@@ -108,28 +111,33 @@ NexusApp::NexusApp(G4String init_macro): G4RunManager(), gen_name_(""),
 
   // Set the user action instances, if any, in the run manager
   if (runact_name_ != "") {
-    runact_ = ObjFactory<G4UserRunAction>::Instance().CreateObject(runact_name_);
-    this->SetUserAction(runact_.release());
+    std::unique_ptr<G4UserRunAction> runact =
+      ObjFactory<G4UserRunAction>::Instance().CreateObject(runact_name_);
+    this->SetUserAction(runact.release());
   }
 
   if (evtact_name_ != "") {
-    evtact_ = ObjFactory<G4UserEventAction>::Instance().CreateObject(evtact_name_);
-    this->SetUserAction(evtact_.release());
+    std::unique_ptr<G4UserEventAction> evtact =
+      ObjFactory<G4UserEventAction>::Instance().CreateObject(evtact_name_);
+    this->SetUserAction(evtact.release());
   }
 
   if (stkact_name_ != "") {
-    stkact_ = ObjFactory<G4UserStackingAction>::Instance().CreateObject(stkact_name_);
-    this->SetUserAction(stkact_.release());
+    std::unique_ptr<G4UserStackingAction> stkact =
+      ObjFactory<G4UserStackingAction>::Instance().CreateObject(stkact_name_);
+    this->SetUserAction(stkact.release());
   }
 
   if (trkact_name_ != "") {
-    trkact_ = ObjFactory<G4UserTrackingAction>::Instance().CreateObject(trkact_name_);
-    this->SetUserAction(trkact_.release());
+    std::unique_ptr<G4UserTrackingAction> trkact =
+      ObjFactory<G4UserTrackingAction>::Instance().CreateObject(trkact_name_);
+    this->SetUserAction(trkact.release());
   }
 
   if (stepact_name_ != "") {
-    stepact_ = ObjFactory<G4UserSteppingAction>::Instance().CreateObject(stepact_name_);
-    this->SetUserAction(stepact_.release());
+    std::unique_ptr<G4UserSteppingAction> stepact =
+      ObjFactory<G4UserSteppingAction>::Instance().CreateObject(stepact_name_);
+    this->SetUserAction(stepact.release());
   }
 
 
--- a/source/base/NexusApp.cc
+++ b/source/base/NexusApp.cc
@@ -11,6 +11,7 @@
 #include "NexusApp.h"
 
 #include "BatchSession.h"
+#include "GeometryBase.h"
 #include "DetectorConstruction.h"
 #include "PrimaryGeneration.h"
 #include "FactoryBase.h"
@@ -18,6 +19,7 @@
 #include <G4GenericPhysicsList.hh>
 #include <G4UImanager.hh>
 #include <G4StateManager.hh>
+#include <G4VPrimaryGenerator.hh>
 #include <G4VPersistencyManager.hh>
 #include <G4UserRunAction.hh>
 #include <G4UserEventAction.hh>
--- a/source/base/NexusApp.cc
+++ b/source/base/NexusApp.cc
@@ -99,32 +99,43 @@ NexusApp::NexusApp(G4String init_macro): G4RunManager(), gen_name_(""),
   if (gen_name_ == "") {
     G4Exception("[NexusApp]", "NexusApp()", FatalException, "A generator must be specified.");
   }
-  pg->SetGenerator(std::unique_ptr<G4VPrimaryGenerator>(ObjFactory<G4VPrimaryGenerator>::Instance().CreateObject(gen_name_)));
+  pg->SetGenerator(ObjFactory<G4VPrimaryGenerator>::Instance().CreateObject(gen_name_));
   this->SetUserAction(pg);
 
   if (pm_name_ == "") {
     G4Exception("[NexusApp]", "NexusApp()", FatalException, "A persistency manager must be specified.");
   }
-  PersistencyManagerBase* pm = ObjFactory<PersistencyManagerBase>::Instance().CreateObject(pm_name_);
+  std::unique_ptr<PersistencyManagerBase> pm = ObjFactory<PersistencyManagerBase>::Instance().CreateObject(pm_name_);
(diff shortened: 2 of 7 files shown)


jacg:
Hold the generator by `std::unique_ptr<T>` rather than a raw pointer (`T*`), and all these things take care of themselves automatically.

Generally speaking, since C++11, *anything* that you own, should be held by `std::unique_ptr<T>`. I appreciate that Geant4 *sometimes* makes this difficult.

paolafer:
Thanks, Jacek, I'll try the `std::unique_ptr`: I'm getting some errors, but I'll try to figure out how to use them.

paolafer:
I'm afraid that in this particular case Geant4 won't allow us to use `std::unique_ptr`. The problem is that we use the same factory to create geometries, generators, persistency managers and actions and in the case of the actions the object the factory returns must be a raw pointer, because it has to be passed to the `SetUserAction()` function of the run manager, which doesn't accept `std::unique_ptr`. For simplicity, manteinance and homogeneity of the code, I think that it is better to keep the same factory for all.

jacg:
It's fine for the factory to return a bare pointer, but whatever *owns* (is responsible for the lifetime of) the thing at the end of that pointer, should hold on to it via `std::unique_ptr`. In other words, you were planning to `delete` that pointer in the destructor of some class: *that* class should hold the thing via `unique_ptr` rather than a bare pointer, and then the destruction will happen automatically.

Specifically, (not tested, off the top of my head) rather than

```c++
class MyClass {
public:
  MyClass(T* t) : t{t} {}
  ~MyClass() { delete t; }
private:
  T* t;
};
```
you should have
```c++
class MyClass {
public:
  MyClass(T* t) : t{t} {}
private:
  std::unique_ptr<T> t;
};
```
or rather
```c++
class MyClass {
public:
  MyClass(std::unique_ptr<T> t) : t{std::move(t)} {}
private:
  std::unique_ptr<T> t;
};
```
to make it clear to the caller of `MyClass`' constructor, that it is going to take ownership.

In the first two cases, passing the `T*` to `MyClass::MyClass` looks no different from passing it to `SetUserAction` so there is no way to tell, by just reading the client code, who owns the `T`.

In the third case, the caller is *obliged* to pass in a `std::unique_ptr<T>` (trivially created from the bare pointer, right there) which sends a clear signal that `MyClass` takes ownership.

This is a huge improvement to the language that arrived in C++11. Take advantage of it!

source/base/NexusApp.cc
```diff
@@ -99,7 +99,7 @@ NexusApp::NexusApp(G4String init_macro): G4RunManager(), gen_name_(""),
   if (gen_name_ == "") {
     G4Exception("[NexusApp]", "NexusApp()", FatalException, "A generator must be specified.");
   }
-  pg->SetGenerator(ObjFactory<G4VPrimaryGenerator>::Instance().CreateObject(gen_name_));
+  pg->SetGenerator(std::unique_ptr<G4VPrimaryGenerator>(ObjFactory<G4VPrimaryGenerator>::Instance().CreateObject(gen_name_)));
```
jmalbos: I'd pass here the raw pointer and handle the conversion inside the `PrimaryGeneration` class.

source/base/PrimaryGeneration.h
```diff
@@ -29,21 +28,21 @@ namespace nexus {
     void GeneratePrimaries(G4Event*);
 
     /// Sets the primary generator
-    void SetGenerator(G4VPrimaryGenerator*);
+    void SetGenerator(std::unique_ptr<G4VPrimaryGenerator> pg);
```
jmalbos, line 32: This function should take a raw pointer as argument, as per my other comments.

source/base/PrimaryGeneration.h
```diff
@@ -29,21 +28,21 @@ namespace nexus {
...
   private:
-    G4VPrimaryGenerator* generator_; ///< Pointer to the primary generator
+    std::unique_ptr<G4VPrimaryGenerator> generator_; ///< Pointer to the primary generator
   };
 
   // INLINE DEFINITIONS //////////////////////////////////////////////
 
-  inline void PrimaryGeneration::SetGenerator(G4VPrimaryGenerator* pg)
-  { generator_ = pg; }
+  inline void PrimaryGeneration::SetGenerator(std::unique_ptr<G4VPrimaryGenerator> pg)
```
jmalbos, line 42: I'd do here the initialisation of the `std::unique_ptr` that you are doing right now in the class `NexusApp`.

jmalbos (changes requested):

source/base/NexusApp.cc
jacg: No, please don't do that!

Dramatic TL;DR: you are suggesting that `SetGenerator` should **lie** about (not) taking ownership of a resource. Please don't do that.

C++ smart pointers serve two major purposes:

+ automatic resource management;
+ giving clear messages about resource management, *in the source code itself*.

Specifically, any function that accepts a reference or a raw pointer is stating that it is only *borrowing* the resource and will *not* take part in the lifetime management of that resource; any function that accepts a smart pointer (i.e. has a parameter with some smart pointer type) is stating that it *will take ownership* of the resource it is being given (shared ownership in the case of `shared_ptr` and sole ownership in the case of `unique_ptr`).

On the other side of the coin, a caller passing a reference or bare pointer into a function should be confident that the function will only borrow the resource and not participate in the management of its lifetime, so passing the resource in will have no effect on its lifetime management whatsoever; a caller passing a *smart* pointer can clearly see, without having to refer to any documentation or ***dig through the source of the function being called***, that the lifetime of this resource **will** be managed by the callee.

Your suggestion to pass in a raw pointer and convert it into a smart one inside the implementation, completely violates this principle: the callee **will** take ownership of the resource, but in order to understand this, the caller has to dig through the source code of the callee. This is exactly the sort of pain and confusion that smart pointers are supposed to avoid.

IIRC, the initial objection to using a smart pointer here was that the resource is allocated by a factory, and that factory's product is passed to Geant4, which does not expect smart pointers. I haven't read the rest of the code in detail, but I suspect that the factory *should* return a `unique_ptr`, and its `.get()` method should be used to lend the resource to Geant4. This would be impossible if G4 calls the factory itself directly, and we then extract the pointer from the bowels of G4 later, before passing it to `SetGenerator`. If this is not the case, then I strongly recommend that you have the factory return a `unique_ptr`. In which case, you wouldn't be creating a `unique_ptr` just before passing it in to `SetGenerator` but `std::move`ing one into `SetGenerator`. This sends a clear message: the factory used to own the resource, but we've `move`d it out of there and given ownership to `SetGenerator`. That's the way `uniqe_ptr` was intended to be used, and unless G4 makes it *completely impossible*, do it that way.

[One key feature that Rust introduces is that the compiler will not allow you to lie about the ownership of resources in the way you suggested! The downside of this is that Rust forces you to understand what you are doing, and that takes time and effort ... and who has time to understand the code we're writing, right?]

jmalbos: > IIRC, the initial objection to using a smart pointer here was that the resource is allocated by a factory, and that factory's product is passed to Geant4, which does not expect smart pointers. I haven't read the rest of the code in detail, but I suspect that the factory should return a unique_ptr, and its .get() method should be used to lend the resource to Geant4. This would be impossible if G4 calls the factory itself directly, and we then extract the pointer from the bowels of G4 later, before passing it to SetGenerator. If this is not the case, then I strongly recommend that you have the factory return a unique_ptr. In which case, you wouldn't be creating a unique_ptr just before passing it in to SetGenerator but std::moveing one into SetGenerator. This sends a clear message: the factory used to own the resource, but we've moved it out of there and given ownership to SetGenerator. That's the way uniqe_ptr was intended to be used, and unless G4 makes it completely impossible, do it that way.

This would be the best solution, I agree.

source/base/NexusApp.cc
```diff
@@ -99,32 +99,32 @@ NexusApp::NexusApp(G4String init_macro): G4RunManager(), gen_name_(""),
...
-  PersistencyManagerBase* pm = ObjFactory<PersistencyManagerBase>::Instance().CreateObject(pm_name_);
+  std::unique_ptr<PersistencyManagerBase> pm = ObjFactory<PersistencyManagerBase>::Instance().CreateObject(pm_name_);
   pm->SetMacros(init_macro, macros_, delayed_);
 
  // PersistencyManager::Initialize(init_macro, macros_, delayed_);
 
   // Set the user action instances, if any, in the run manager
   if (runact_name_ != "")
-    this->SetUserAction(ObjFactory<G4UserRunAction>::Instance().CreateObject(runact_name_));
+    this->SetUserAction(ObjFactory<G4UserRunAction>::Instance().CreateObject(runact_name_).get());
```
jacg: This looks wrong.

The 'unique' in `unique_ptr` means that the object (the smart pointer, that is) can only exist in one place at a time: it can be *moved* from place to place, but it cannot be *copied*.

If you pass it to a function that takes a `unique_ptr<T>` you must move it from the caller into the function. If you return one from a function, it must be moved out of the function to the caller. This is what is happening here: 

1. The `unique_ptr` is moved (implicitly, because it's an rvalue) out of `CreateObject`. `CreateObject` no longer owns the  resource that the `unique_ptr` points to.
2. You use `.get()` to borrow (in other words, get a non-owning, observing handle: a raw pointer to) the resource, and lend it to `SetUserAction`.
3. The `unique_ptr` is dropped on the floor, so it gets destructed, so it destructs the resource it points to ... the same resource you just lent to `SetUserAction`.

Consequently, `SetUserAction` has been given a pointer which is guaranteed to be invalidated immediately.

By not moving the `unique_ptr` anywhere, not giving it to anyone to keep, but dropping it on the floor, you are saying 'nobody wants this any more', at which point the resource it owns gets deleted.

Who is responsible for the lifetime of the thing the factory just made?

+ If it's Geant4 via `SetUserAction`, (and Geant4 doesn't understand/use smart pointers (even though they were a crucial addition to the standard over a decade ago)) then use the `.release()` method of `unique_ptr` instead of `.get()` to say "I'm relinquishing ownership, it's your problem now".

+ If your code owns this thing (is responsible for deleting it), then make sure that you move the `unique_ptr` you got from the factory, into whatever it is that is supposed to own it.

jacg: > Consequently, `SetUserAction` has been given a pointer which is guaranteed to be invalidated immediately.

Which is another example of nonsense that is simply not allowed by the Rust compiler.

paolafer: I've tried to implement what you said, let me know if that's the correct way.

jacg: Yes, this looks much more plausible. Now:

+ `this->XXXact_` is the owner and is responsible for its lifetime,
+ `SetUserAction` gets to look at it and use it, but is not responsible for its lifetime.

But it begs the question: who was deleting these before? Or is this another instance of what this PR was trying to fix: these weren't being deleted at all?

paolafer: I suspect they weren't being deleted (unless Geant4 did it itself).

jacg: > unless Geant4 did it itself.

Eeeeek!

It's important to know whether Geant4 releases it or not! (And because G4 doesn't use smart pointers, you can't tell by just looking at the signature ... which is half of the point of these smart pointers.) 

+ If G4 *does* delete them, then what you have done is very wrong, and will result in double-deletion (G4 owns it, you also think you own it, and both of you delete it). In which case you should `.release()` into `SetUserAction`, and drop the `unique_ptr` on the floor. 
+ If G4 does *not* delete them, then your code looks correct.

paolafer: Interesting point. I found this:

https://geant4.kek.jp/lxr/source/run/src/G4RunManager.cc#L216

It seems that in the case of the actions Geant4 does delete them, right? I'm not sure I understand the comment: is it saying that we should not take over the ownership and set to zero the pointers in the destructor?

jacg: Let's just take a moment to appreciate how much of our time is being wasted by this comment, and how much time we would save if G4 used smart pointers as the Standard Committee intended, rather than whatever ad-hoc scheme this is.

Now, what does it mean?

> [...] for all RunManager types if derived class [...]

[Here](https://gitlab.cern.ch/geant4/geant4/-/blob/master/source/run/include/G4RunManager.hh#L154) we see that `G4RunManager` has `virtual` methods (destructor, in this case) so it is meant to be subclassed. So 'all RunManager types' probably means 'all subclasses of `G4RunManager`'.

> [...] does the correct thing in the derived destructor that is: set to zero pointers of user initialization objects for which have no ownership

+ `delete`-ing a `nullptr` is guaranteed to be safe, essentially a null-op.
+ [`DeleteUserInitializations`](https://gitlab.cern.ch/geant4/geant4/-/blob/master/source/run/src/G4RunManager.cc#L244) `delete`s a bunch of member pointers.

So it seems to be saying that the onus is on the implementor of subclasses of `G4RunManager` to nuillify pointers it does not own, in the destructor, because this base class' destructor will `delete` anything that hasn' been nullified. Effectively, if it ain't `NULL` at destruction time, I'll assume that I own it and I will `delete` it.

Which tells us absolutely nothing about how the subclass of `G4RunManager` and its clients are supposed to decide who owns what. Maybe this decision process is documented somewhere? In general, it would seem that you have to go to the specific subclass of `G4RunManager` that you happen to be using and find out which of the things that you gave it, are nullified in its destructor (which might be done by a superclass' destructor). Have fun and good luck!

And now let's remind ourselves how this would look if G4 used smart pointers as intended:

+ If G4 takes ownership, it accepts the pointer as a `unique_ptr` and the compiler makes it clear to the client that ownership is being taken, because the client will either have to create a `unique_ptr` or `std::move` an existing `unique_ptr`, when calling the function, before the compiler is happy.

+ If G4 does not take ownership, it accepts the a bare pointer (or a reference) and the client knows that it keeps ownership.

The first case is always obvious because the compiler can enforce it. The second case, *crucially* requires G4 to be playing by the rules of smart pointers, which it isn't. In such cases the C++ compiler (unlike the Rust compiler) has absolutely no way of helping you. And that's the situation in which we find ourselves here. Yippeee!

paolafer: After discussing with @jmalbos we've come to the conclusion that, given that Geant4 does not use smart pointers, it's cumbersome and potentially error-prone to keep track of each pointer given to its classes. So, we think that it's better to use smart pointers only when nexus classes are the owners for all their lifetime. In this PR that means going back to commit 26de80f5.

jacg: > given that Geant4 does not use smart pointers, [...] we think that it's better to use smart pointers only when nexus classes are the owners for all their lifetime.

Whether or not a nexus class should hold a smart pointer, is *completely independent* of whether or not Geant4 is smart pointer aware!

I get the impression that we haven't understood the rules of smart pointer usage. Given how useful and important (and in comparison to most other C++ features, **simple**) they are, it's crucial that we get this right. Let me have another go at explaining them, this time with a bit more detail, and explicitly covering legacy libraries. (I talk only about `unique_ptr`, but the rules are very similar for `shared_ptr`.) It may seem confusing and complicated at first, especially if there are legacy libraries to deal with, but, trust me, it really is one of the simpler and most useful things in C++.

1. Producers:

   + Legacy producers will return bare pointers. There is nothing you can do about this. But, if you don't *immediately* pass it to a legacy consumer, you should wrap it in a `unique_ptr` asap.
   + Modern producer: *Every* factory that you write in C++11 and beyond should return a `std::unique_ptr<T>`, without exception. (Yes, I did come up with a possible exception upthread. Let's ignore such obscure possibilities until we have thoroughly understood the fundamental rules of the game.)

2. Observers (things that use but do not own the `T` made by a producer):

   + It doesn't matter whether the observer be legacy or modern, in both cases it will take `T*` or `T&`. The owner of the `T` will use `std::unique_ptr::get()` to give the bare pointer or reference to the observer.

3. Consumers (this includes owners):

   + Legacy consumers will take (and in the case of owners, also keep) `T*` or `T&`. Whoever passes ownership to a legacy consumer, must use `std::unique_ptr::release()` to relinquish ownership to the consumer. The consumer will have to take care to `delete` the `T` at some stage, or pass ownership on to somebody else (but crucially, not both). Distinguishing legacy consumers from legacy observers, and ensuring that they `delete` every resource exactly once between them, is one of the major challenges of C++ programming.
   + Modern consumers will take (and in the case of owners, also keep) `std::unique_ptr<T>`. Whoever passes ownership to a modern consumer must use `std::move` to relinquish ownership to the consumer (though the move is implicit in the case of rvalues). The consumer won't have to `delete` the `T`, because it happens automatically when the `unique_ptr` is destructed, it will happen automatically regardless of whether ownership has been passed on or not; and it will happen exactly *once*, no more, no less. This is a big deal. It's a HUGE deal. Take the deal. Embrace the deal. It's one of the biggest and best deals there is in C++. Thank your lucky stars (and the standard committee) that the deal is now available.

Applying this to Nexus and Geant4:

+ Every factory you write in Nexus, should return a `unique_ptr<T>`
+ If the factory's product is owned by Geant4 and observed (used) by nexus:

  ```c++
  auto thing = thing_factory();
  nexus_observer(thing.get());
  geant4_owner(thing.release());
  ```
  NB: after `thing.release()`, `thing.get() == nullptr`

  ```c++
  class nexus_class_that_does_NOT_own_T {
    T*;
  };
  ```


+ If the factory's product is observed by Geant4 and owned by nexus:

  ```c++
  auto thing = thing_factory();
  geant4_observer(thing.get());
  nexus_owner(std::move(thing));
  ```
  NB: (generally speaking, not just in the case of `unique_ptr`) after `std::move(thing)`, `thing` is in an unusable state (unless you re-initialize it). `std::move()` effectively means: 'Dear compiler, I promise that `thing` will not be used again'. (In C++ it's up to the programmer to make sure this is the case; in Rust, the compiler makes sure that objects are not used after having been moved.)

  ```c++
  class nexus_class_that_DOES_own_T {
    std::unique_ptr<T>;
  };
  ```

> we think that it's better to use smart pointers only when nexus classes are the owners for all their lifetime.

While it is not uncommon for ownership of a resource to be transferred during its lifetime (well, it's guaranteed at least once: from factory to first owner), I suspect that this won't be happening much in nexus. Which gives me the impression that you have some, erm ... unorthodox ideas about what ownership is. (However, if nexus really is the temporary owner of some resource, and you think that it's better not to use smart pointers in such cases, well, I'm afraid you are simply worng. In fact, that would be an example of a situation where it's especially important to use smart pointers!)

I *suspect* that in most of these cases the fundamental question is: "This thing I just passed to Geant4, will Geant4 delete it?"

+ If the answer is yes:

  - Store it as `T*` in a nexus class.
  - `.get()` when giving it to the nexus class.
  - `.release()` when passing it to Geant4.
  
+ If the answer is no:

  - Store it as `std::unique_ptr<T>` in a nexus class.
  - `.get()` when passing it to Geant4.
  - `std::move()` when giving it to the nexus class.
  
NB: `.get()` must precede `.release()` or `std::move()`.

One more thing: *always* fully qualify `std::move`. That is, *never* `using std::move; [...] move()`.

> [...] it's cumbersome and potentially error-prone to keep track of each pointer given to its classes. [...]

Not sure what exactly is meant by this, but I can assure you that doing anything other than following the rules set out above is a lot more cumbersome and error-prone following the rules.

At the core of all of this is a very simple idea: 

+ If I have a `unique_ptr<T>` it means that I am the *sole owner* of this `T`. <sup><sub>If I wish, I can give my `T` away to someone else with `std::move()` (or `.release()`, if the new owner is legacy code).</sub></sup>
+ If I have a `T*` or `T&` it means that I am looking at someone else's `T`. (NB: in *legacy* code, it means absolutely nothing.)

paolafer: I think this is clear.

> [...] it's cumbersome and potentially error-prone to keep track of each pointer given to its classes. [...]
Not sure what exactly is meant by this, but I can assure you that doing anything other than following the rules set out above is a lot more cumbersome and error-prone following the rules.

What I meant is that it is not always straightforward to know the answer to

>  "This thing I just passed to Geant4, will Geant4 delete it?"

That's why we thought it was less error prone not to use smart pointers for pointers that are passed to Geant4.
Maybe we can do as you suggest and one way to be sure that we don't delete pointers twice (once in nexus and a second time in Geant4) is to delete them always in nexus and then set them to `nullptr` afterwards, as the G4RunManager class seems to suggest?

jacg: >> What I meant is that it is not always straightforward to know the answer to

> "This thing I just passed to Geant4, will Geant4 delete it?"

Of course! This difficulty is part and parcel of using legacy libraries!

Regardless of how difficult it is to answer this question, you *still* have to find the correct answer, not just in this case, but in all cases where you pass pointers. Using smart pointers can only make it easier to reason about these questions and write correct programs, it cannot make it more difficult. You cannot diminish the difficulty by not using smart pointers.

That is why you should migrate code that you own, away from this legacy style. Use smart pointers consistently in your code, and your clients will be able to answer this question trivially when reasoning about your code. In other words, don't inflict on your clients the same kind of pain that Geant4 is inflicting on you by not using smart pointers.

> That's why we thought it was less error prone not to use smart pointers for pointers that are passed to Geant4.

I trust that we agree that any pointer you pass to Geant4 (today) will *always* be a bare pointer. If you're suggesting that it will be less error prone to avoid using smart pointers altogether for anything that will eventually end up being passed to Geant4, then I *strongly* disagree.

> Maybe we can do as you suggest and one way to be sure that we don't delete pointers twice (once in nexus and a second time in Geant4) is to delete them always in nexus and then set them to nullptr afterwards, as the G4RunManager class seems to suggest?

In other words, you are suggesting to avoid the cost of answering the question "Will G4 delete this?" by hacking G4 in an ad-hoc way, tracking down every pointer you ever passed it, and ensuring that G4 *never* deletes any of them, so that you know that you have to delete them in your code? Please don't torture yourselves (and your users) like this! Madness lies this way.

The C++ standard committee provided you (pretty decent, as C++ goes) tools for ensuring that (dynamically allocated) resources get deleted exactly once: the C++11 standard smart pointers (`uniqe_ptr` and `shared_ptr`, not to be confused with the C++98 `auto_ptr`, which was *fundamentally broken*). Please don't avoid using them in favour of your own schemes that you make up on the spot! This cannot end well. 

You **have** to answer the crucial ownership question. There's no way around it. (Well, there is: write buggy code.) 

If you believe [this](https://gitlab.cern.ch/geant4/geant4/-/blob/master/examples/basic/B1/exampleB1.cc#L105-L107), then you should conclude that

+ user actions
+ physics lists
+ detector description

are owned by the run manager. If you need to store these in your own classes or pass them around your own functions, do so via bare pointers.

Way at the top of this PR, you added a deletion of a generator in a nexus class' destructor. It seems that G4 does not take ownership of these, and your nexus class does. Store these in your class via `unique_ptr<T>`, but pass them around via `T*`. Except when transferring them from the factory into the class: that should be done with `unique_ptr`.

For anything else, you'll have to find the correct answer to the ownership question, and act accordingly.

Even if you don't store it in your code at all and pass it to G4 as soon as it is made, you should probably still use smart pointers! It might be edifying to put yourselves in the shoes of a (smart pointer aware - and if you're not smart pointer aware in 2022, you shouldn't be programming in C++) programmer reading your code > 6 months in the future, and compare and contrast these alternatives 

1. ```c++
   auto t = my_factory();
   some_g4_instance.set_something(t);
   ```

2. ```c++
   auto t = my_factory();
   some_g4_instance.set_something(t.get());
   ```
   
3. ```c++
   auto t = my_factory();
   some_g4_instance.set_something(t.release());
   ```

+ What do these different samples tell you?
+ One of them should be ringing alarm bells, and have you asking to see the next few lines of the code! (Actually, *two* of them should be ringing alarm bells: one because of a very specific and obvious problem which might be solved on the next line; another because of the general paranoia any decent C++ programmer should have.)


Key takeaway:

+ In C++, you *must* know who will delete dynamically allocated resources (however difficult it may be to find out).
+ Smart pointers can only make this easier, not more difficult.

source/base/NexusApp.cc
```diff
@@ -86,12 +85,12 @@ NexusApp::NexusApp(G4String init_macro): G4RunManager(), gen_name_(""),
   this->SetUserInitialization(physicsList);
 
   // Set the detector construction instance in the run manager
-  DetectorConstruction* dc = new DetectorConstruction();
+  dc_ = std::make_unique<DetectorConstruction>();
```
jacg: Given that you 

1. create the DC
2. set its geometry
3. then release it immediately to G4

I don't think it makes much sense to keep it as a member. It's not incorrect, but it's not the clearest way of expressing what is going on. It gives the impression to the reader elsewhere, that this class owns the detector construction for its whole lifetime, whereas it owns it very briefly and keeps an empty shell whose contents have been released.

Just make `dc` a local variable here. Ideally, limit its visibility with an anonymous scope.

```c++
{
  auto dc = make_unique<DC>();  // I've made a DC. It's mine, ALL MINE, I telly you!
  dc -> SetGeometry(...);       // I take good care of my DC, give it what it needs.
  whatever(dc.release())        // You can have my DC, look after it well. I'll miss you my old friend. <*snif*>
} // The variable `dc` goes out of scope here, so nobody gets the opportunity to abuse it or get confused by it further down.
```

It's a very clear (if somewhat melancholy) story, when told like this.

source/base/NexusApp.cc
```diff
@@ -76,27 +79,27 @@ NexusApp::NexusApp(G4String init_macro): G4RunManager(), gen_name_(""),
   // The physics lists are handled with Geant4's own 'factory'
   physicsList = new G4GenericPhysicsList();
 
-  batch_ = std::make_unique<BatchSession>(init_macro.c_str());
-  batch_->SessionStart();
+  auto batch = std::make_unique<BatchSession>(init_macro.c_str());
+  batch->SessionStart();
```
jacg: + `batch` is a `unique_ptr<BatchSession>` therefore it owns the `BatchSession`.
+ It is also a *local* variable, therefore it will go out of scope as soon as we reach the end of this function.
+ Ownership is not transferred anywhere before we reach the end of this function.
+ Ergo, the `BatchSession` gets deleted as soon as we reach the end of this function (`NexusApp::NexusApp`).

Now, maybe that's what is supposed to happen (I don't have the time try to follow the logic of `BatchSession`) but it does seem a bit odd and forces the reader to hunt for uses of `batch` in this veeery long function.

Which is why I *thoroughly* recommend that you use local scopes to limit lifetimes: put these two lines between a pair of `{ ... }` and it becomes immediately obvious that `batch` will not be used again after these two lines. It *really* does simplify reading the code.

Once you've done that, you might wonder, Why don't I just write it as `BatchSession(init_macro.c_str()).SessionStart()`? which is semantically equivalent and much less verbose.

Valid answers might be

+ The average Nexus contributor might not spot the full implications of this line: A temporary, stack-allocated `BatchSession` is automatically destructed as soon as `SessionStart()` returns.

+ The smart pointer version suggests to the reader of the code that the author of the code has actually thought about lifetimes here, rather than being completely oblivious to the need to think about them.

As much as I believe that you should write as few comments as possible, because you should write code that is self-explanatory, in this case I would recommend a comment. You are programming in a context where you cannot assume that the people who will read and modify this code have a sufficiently good understanding of the lifetime implications. Consequently, code which is self-explanatory to a C++ expert, will remain mysterious to  many Nexus contributors. Write a comment which points out that/where the `BatchSession` is destructed.

And *do* limit the scope of variables such as these, using anonymous scopes. That way the aforementioned comment will be here, rather than 70 lines further down.

source/base/NexusApp.cc
```diff
@@ -206,7 +206,7 @@ void NexusApp::ExecuteMacroFile(const char* filename)
   std::unique_ptr<G4UIsession> batchSession =
     std::make_unique<BatchSession>(filename, UI->GetSession());
   UI->SetSession(batchSession.get());
-  G4UIsession* previousSession = UI->GetSession()->SessionStart();
+  G4UIsession* previousSession = batchSession->SessionStart();
   UI->SetSession(previousSession);
```
jacg: 1. I recommend `using std::unique_ptr;` and `using std::make_unique;` at the top of this file, so that you can avoid having to qualify them with `std` all over the place. NB: do **NOT** do `using std::move`. Always fully qualify `std::move`.

2. Is there any good reason for writing
   ```c++
   std::unique_ptr<G4UIsession> batchSession = std::make_unique<BatchSession>(filename, UI->GetSession());
   ```
   instead of 
   ```c++
   auto batchSession = make_unique<BatchSession>(filename, UI->GetSession());
   ```
   ?

source/base/NexusApp.cc
```diff
@@ -77,13 +77,13 @@ NexusApp::NexusApp(G4String init_macro): G4RunManager(), gen_name_(""),
...
   // The physics lists are handled with Geant4's own 'factory'
-  physicsList = new G4GenericPhysicsList();
+  auto pl = std::make_unique<G4GenericPhysicsList>();
 
   auto batch = std::make_unique<BatchSession>(init_macro.c_str());
   batch->SessionStart();
 
   // Set the physics list in the run manager
-  this->SetUserInitialization(physicsList);
+  this->SetUserInitialization(pl.release());
```
jacg: The initialization of `pl` should not be separated from its use, by the completely unrelated `batch` code.

Lines 80 and 86 could, and probably should, be condensed down to nothing more than
```c++
this -> SetUserInitialization(new G4GenericPhysicsList());
```
which states very clearly that the lifetime of the physics list is `SetUserInitialization`'s problem, not ours.

In fact, this is a classic example of the only situation where `new` should appear in modern C++ code[*]: I am creating something and immediately giving ownership of it to a legacy consumer. 

[*]  unless you are implementing a low-level utility such as `std::unique_ptr` or `std::vector` itself, which you will not be doing in Nexus.

paolafer: In this case the order cannot be changed, since we need to instantiate a `G4GenericPhysicsList` first, then read the `init` configuration file, which contains the physics lists that needs to be registered and finally hand over to G4 our pointer through `SetUserInitialization`.

source/base/NexusApp.cc
paolafer: > * Is there any good reason for writing
>   ```c++
>   std::unique_ptr<G4UIsession> batchSession = std::make_unique<BatchSession>(filename, UI->GetSession());
>   ```

>   instead of
>   ```c++
>   auto batchSession = make_unique<BatchSession>(filename, UI->GetSession());
>   ```     
>   ?

`batchSession` is passed as an argument of the `SetSession()` function of the `G4UIsession` class and this function needs a `G4UIsession` object, so I think that the reason was to make clear that this object is of type `G4UIsession`. Maybe it's not necessary.

source/base/NexusApp.cc
jacg: So `batch->SessionStart()` results in some code being executed that, magically, behind the scenes, depends on the existence of a `G4GenericPhysicsList` instance whose existence has not been communicated anywhere (at least not in any way that is visible in this code)?

Genuis! (Well, I shouldn't be surprised: It's Geant4.)

From the perspective of someone reading the C++ code that is written here, these lines commute! If there is a reason why these lines do not commute, then it should be thoroughly and loudly documented right here, with unmissable comments in the code.

(Looks like another G4 trap that I should try to address in `nain4`.)

source/base/NexusApp.cc
jacg: The type of `batchSession` is definitely `BatchSession` (give or take a pointer). `BatchSession` implements (is a subtype of) `G4UIsession` which is why it can be passed to functions which formally accept the latter. All this requires *very* basic understanding of C++ inheritance and absolutely anyone getting anywhere near C++ should be thoroughly aware of it, so I don't think we're doing anyone any favours by adding the extra noise.

If it were up to me, I'd get rid of it.

jacg:
It looks like you've got the hang of it now. Good work.

source/base/DetectorConstruction.h
```diff
@@ -9,15 +9,15 @@
...
+#include "GeometryBase.h"
+
 #include <G4VUserDetectorConstruction.hh>
 
 class G4GenericMessenger;
 
 
 namespace nexus {
 
-  class GeometryBase;
```
jmalbos, line 19: This forward declaration can be kept.

source/base/DetectorConstruction.h
```diff
@@ -9,15 +9,15 @@
 #ifndef DETECTOR_CONSTRUCTION_H
 #define DETECTOR_CONSTRUCTION_H
 
+#include "GeometryBase.h"
```
jmalbos: This can be moved to the `.cc` file.

jmalbos: This can be moved to the `.cc` file.

source/base/NexusApp.cc
```diff
@@ -10,10 +10,9 @@
 
 #include "NexusApp.h"
 
+#include "BatchSession.h"
```
jmalbos, line 13: You'd need to include here the `GeometryBase` header if `DetectorConstruction.h` only has a forward declaration, as suggested above.

jmalbos, line 13: You'd need to include here the `GeometryBase` header if `DetectorConstruction.h` only has a forward declaration, as suggested above.

jmalbos (changes requested):
The code looks good to me, but perhaps you could tidy the commits a bit (there are 18 right now) before merging into master.

paolafer:
I've cleaned up the history a little bit, let me know if there's anything else.

jmalbos (approved):

PR #69, opened 2021-01-22
Lab generator

andLaing:
Improves the way muons are generated using the angular correlations expected in a lab.

Introduces `ProjectToRegion` methods in the geometries (NEW, NEXT100 and NEXT100Shielding and LSCHallA for now) and `GetIntersect` to the `BoxPointSampler` and `CylinderPointSampler2020`.

Tests added to nexus-test

Replaces #54 which was closed in the transition to a public repository.

PR #146, opened 2022-03-08
Remove NEXUS ROOT Dependence

kvjmistry:
This PR removes the ROOT dependence in NEXUS, primarily in the Muon Generators. Now, events are generated using a csv file rather than a root file. 

An additional simulated muon file is included which is generated from the MCEq and Proposal simulation. This can be chosen from the config macro file when running the generator. 

Tested compilation on Harvard Cluster (Cent Os) and Macbook M1. 

Please see https://next.ific.uv.es/cgi-bin/DocDB/private/ShowDocument?docid=1325 for slides explaining this PR.


source/actions/MuonsEventAction.cc
```diff
@@ -21,10 +21,8 @@
 #include <G4SDManager.hh>
 #include <G4HCtable.hh>
```
gondiaz: `G4HCtable.hh` and `G4HCofThisEvent.hh` are not needed

source/actions/ValidationTrackingAction.cc
```diff
@@ -23,9 +23,8 @@
 #include <G4ParticleDefinition.hh>
```
gondiaz: `G4ParticleDefinition.hh` not needed

kvjmistry: Thanks, I have removed the headers.

source/actions/MuonsEventAction.cc
```diff
@@ -18,13 +18,9 @@
 #include <G4Trajectory.hh>
 #include <G4GenericMessenger.hh>
 #include <G4HCofThisEvent.hh>
```
gondiaz: Remove `G4HCofThisEvent.hh` also

source/generators/MuonAngleGenerator.h
```diff
@@ -71,12 +78,22 @@ namespace nexus {
...
+    std::vector<G4double> azimuth_bins; ///< List of Azimuth bin edges
+    std::vector<G4double> zenith_bins;  ///< List of Zenith bin edges
+    std::vector<G4double> az_BW; ///< List of Azimuth bin widths
+    std::vector<G4double> zen_BW; ///< List of Zenith bin widths
+
+    std::mt19937 RN_engine;  ///< Random Number Generator Engine Index
+    std::mt19937 RN_engine_az; ///< Random Number Generator Engine Azimuth
+    std::mt19937 RN_engine_zen; ///< Random Number Generator Engine Zenith
+
+    std::discrete_distribution<G4int> discr_dist; ///< Discrete distribution to sample flux
```
gondiaz: By convention, data members in nexus are followed by a trailing underscore, see the [Coding conventions and style guide](https://github.com/next-exp/nexus/wiki/Coding-conventions-and-style-guide)

kvjmistry: Thanks for bringing this to my attention, I was not aware of this practice and will update the variable names to match this style.

source/generators/MuonAngleGenerator.cc
```diff
@@ -110,6 +164,18 @@ void MuonAngleGenerator::SetupAngles()
 void MuonAngleGenerator::GeneratePrimaryVertex(G4Event* event)
 {
 
+  // Initalise RNG seeds and load file on the first event
+  if (event->GetEventID() == 0){
```
gondiaz: I think we must avoid this clause. Notice that the first event id is not 0 in general, since this is set via `/nexus/persistency/start_id`. What would the seed be in such a case? Could we just take the seed from `/nexus/random_seed`?

kvjmistry: Right, I couldn't figure out a way to initialize it only once, this protection was added to stop initializing it on every new event. We could possibly use the same protection as the rotation is initialised ie. if ( rPhi_ == NULL) instead if you prefer.  

The CLHEP::HepRandom::getTheSeed() should be already grabbing the /nexus/random_seed configured.

kvjmistry: or we can add a data member bool to flag if initialised, that is a little clearer actually.

jmalbos: Do we really need those `RN_engine`? Can't you just use the functions (e.g. `G4UniformRandom()`) provided by Geant4?

https://geant4.kek.jp/lxr/source/global/HEPRandom/include/Randomize.hh

jmalbos: > I think we must avoid this clause. Notice that the first event id is not 0 in general, since this is set via `/nexus/persistency/start_id`. 

Just for the sake of correctness: The first event ID is always 0. We only modify it at the time we write the events in the output file.

kvjmistry: The random engine is used as an input to the std::discrete_distribution which is what I use to sample the events from. I don't think I can use the G4UniformRandom() for that. 

I can probably replace the other RN generators used for the Gaussian smearing, but I just chose to use the std functions for consistency. Let me know if you would like to use the Geant4 function in this case and I can swap it.

kvjmistry:
Updated simulated muon file so that the flux is normalised by the bin-width. This is because the histogram doesn't have even binning. The small-angle bins have a larger bins, so this contributed to the large number of events at muon dir z = 0. See the picture below of the new simulated angular dist in x-z direction. Note this did not affect the data file since it had even binning. 

New Simulated Muon Dist:
image

Old Simulated Muon Dist without normalising by bin width:
image

kvjmistry:
Managed to find the problem with using the geant4 ntuples. Needed to open the root file in the BeginEventAction -_-. Now write TTree to the same file as the histogram without the need for an additional CSV file.

source/actions/MuonsEventAction.cc
```diff
@@ -17,14 +17,9 @@
 #include <G4VVisManager.hh>
 #include <G4Trajectory.hh>
 #include <G4GenericMessenger.hh>
-#include <G4HCofThisEvent.hh>
-#include <G4SDManager.hh>
-#include <G4HCtable.hh>
 #include <globals.hh>
+#include "g4root_defs.hh"
```
gondiaz: Is this header necessary?

source/actions/ValidationTrackingAction.cc
```diff
@@ -20,12 +20,10 @@
 #include <G4Track.hh>
 #include <G4TrackingManager.hh>
 #include <G4Trajectory.hh>
-#include <G4ParticleDefinition.hh>
 #include <G4OpticalPhoton.hh>
 #include <G4Gamma.hh>
+#include "g4root_defs.hh"
```
gondiaz: Same as above

kvjmistry: These headers pull in the geant4 histogram and ntuples that we are replacing the ROOT histograms with. To keep with convention, I should actually use <> rather than "" in these include lines.

gondiaz: Histograms and NTuples are created through G4AnalysisManager methods [Analysis Manager Classes](https://geant4-userdoc.web.cern.ch/UsersGuides/ForApplicationDeveloper/html/Analysis/managers.html#files-handling). Therefore whatever in this include is not been used.
The manager chooses the default output file from the filename extension (.csv, .hdf5, .root or .xml). I have tried .csv, .root and .xml and seems to work. The .hdf5 should work if Geant4 is compiled against hdf5.

gondiaz (commented):
Review includes in MuonsEventAction.cc, some of them are not needed

source/generators/MuonAngleGenerator.cc
```diff
@@ -175,20 +245,98 @@ G4String MuonAngleGenerator::MuonCharge() const
...
+    // Catch negative value and resample if so
+    if (zen_samp < 0.0)
+        invalid_evt = true;
+    else
+        invalid_evt = false;
+
+    // Calculate the vector components of the muon
+    dir.setX(sin(zen_samp) * sin(az_samp));
+    dir.setY(-cos(zen_samp));
+    dir.setZ(-sin(zen_samp) * cos(az_samp));
```
gondiaz: This assumes that azimuth is the angle between muon direction and -Z axis. Is it the intented behaviour? I would naively consider the azimuth to be between muon direction and +X. 
Also, which is the value of `azimuth_rotation` for NEXT100? I see that `NEXT100_muons_lsc` and `NEXT100_muons_hallA` use the value of 150 degrees. The rotation here is performed anticlockwise throughout Y axis (meaning -150). Is it the correct orientation?

andLaing: All these slightly non-intuitive definitions are due to the difference in axis orientation between NEXT, set so that positive z points towards the cathode, and the 'natural' orientation where positive z points upwards. Additionally, the measurements of the flux in the lab are oriented so that the azimuthal angle 0 is at north so we need to have the azimuth angle around Y (which points up in next) and rotate the axes in the input info to the orientation in NEXT. Since NEXT100 will be in the same shielding as NEW (unless things have changed since I was in NEXT) we always assumed the 150 degrees rotation of the axes would be the same but it's a configuration variable in case of changes there.

There's PR, that's probably very out of date, which updates a lot of important things for muon generation (#69 , think) to make it closer to reality and to aid normalisation, maybe worthwhile having a look.

Hope that helps.

source/generators/MuonAngleGenerator.cc
```diff
@@ -110,6 +164,21 @@ void MuonAngleGenerator::SetupAngles()
 void MuonAngleGenerator::GeneratePrimaryVertex(G4Event* event)
```
gondiaz, line 178: `MuonAngleGenerator` breaks when used with `MuonsEventAction`. Add the angles to the userInfo (check how it is done in`MuonGenerator`)

source/generators/MuonAngleGenerator.cc
```diff
@@ -175,20 +245,98 @@ G4String MuonAngleGenerator::MuonCharge() const
...
+        else {
+          if (zeniths_[RN_indx] >= zenith_bins_[i] 
+                && zeniths_[RN_indx] < zenith_bins_[i+1]){
+              
+              zen_BW_smear = zen_BW_[i];
+
+          }
+        }
+    
+    }
```
gondiaz: Assuming all bins are in the input file, I would suggest to simplify the code to (same for azimuth smearing)

```suggestion
    for (G4int i = 0; i < zenith_bins_.size()-1; i++){
          if (zeniths_[RN_indx] >= zenith_bins_[i] 
              && zeniths_[RN_indx] < zenith_bins_[i+1]){
              zen_BW_smear = zen_BW_[i];
              break;
          }
```

Also, I would set the default `zen_BW_smear` and `az_BW_smear` to a meaninless negative value, say -1.

source/generators/MuonAngleGenerator.cc
```diff
@@ -175,20 +245,98 @@ G4String MuonAngleGenerator::MuonCharge() const
...
+
+          }
+        }
+
+    }
+
+    // Check if the smear values are set properly
+    if (az_BW_smear == 1.0e6 || zen_BW_smear == 1.0e6 )
+      G4Exception("[MuonAngleGenerator]", "GetDirection()",
+                FatalException, " bin smear value was not set correctly ");
```
gondiaz: Following previous comment, if smear values are -1, it would mean that the sampled value (`azimuths_[RN_indx], zeniths_[RN_indx]`) is not inside any of the bins in the input file. Therefore I think a message such that "Sampled angle not inside binning, review input file" will be helpfull.

source/generators/MuonAngleGenerator.cc
```diff
@@ -175,20 +245,98 @@ G4String MuonAngleGenerator::MuonCharge() const
...
+
+    // Gaussian dist to smear by bin widths in azimuth and zenith
+    std::normal_distribution<G4double> Gauss_az(0, az_BW_smear);
+    std::normal_distribution<G4double> Gauss_zen(0, zen_BW_smear);
+    
+    // Get the Gaussian smear values 
+    G4double az_smear  = Gauss_az(RN_engine_az_);
```
gondiaz: As commented above, I would use G4 builtins, in this case `G4RandGauss::shoot`

gondiaz (changes requested):

source/generators/MuonAngleGenerator.cc
gondiaz: Thanks for the clarification!

source/generators/MuonAngleGenerator.cc
gondiaz: We might also avoid doing this check for every sampling, and add a check above such that
`azimuth_bins_[0] <= azimuths_.min() && azimuths_.max() <=  azimuth_bins_[-1]`. This also avoids to define meaninless `az_BW_smear`. Same for zenith

source/generators/MuonAngleGenerator.cc
kvjmistry: Tidied up code by including a new function. Also set default values from 1.0e6 to std::numeric_limits<G4double>::lowest().

Kept two if statements, still need to keep the very last bin with the additional <= sign so it is included.

source/generators/MuonAngleGenerator.cc
kvjmistry: Updated error message to be more descriptive. 

I do not quite understand what you mean by the second check here, happy to follow up with you on slack about this.

source/generators/MuonAngleGenerator.cc
kvjmistry: Updated generator to use the CLHEP RandGauss tools rather than the std library.

source/generators/MuonAngleGenerator.cc
kvjmistry: Added angles to the user info.

source/actions/ValidationTrackingAction.cc
kvjmistry: Yep your right, I removed the superfluous header.

source/generators/MuonAngleGenerator.cc
gondiaz: I would avoid defining two extra random engines but use `G4RandGauss::shoot` instead, what do you think @jmalbos @paolafer?

source/generators/MuonAngleGenerator.cc
gondiaz: Defining a meaningless `smearVal` is only used to check is the sampled `sampVal` is inside any of the bins provided at the configuration file, explicitly:

```
    // Check if the smear values are set properly
    if (az_BW_smear == std::numeric_limits<G4double>::lowest() ||
        zen_BW_smear == std::numeric_limits<G4double>::lowest() ){
                G4Exception("[MuonAngleGenerator]", "GetDirection()",
                FatalException, " Sampled angle not inside binning, review input file ");
    }
```

My point is that instead doing this check at each call to `GetDirection`, doing it globally in `LoadMuonDistribution` with something like (pseudocode):
`azimuth_bins_[0] <= min(azimuths_) && max(azimuths_) <= azimuth_bins_[-1]`

Also, we must be sure that `azimuth_bins_` are sorted, then I would suggest to sort them once they are pulled from the input file. 
Sure, contact me if I am not explaining myself!

gondiaz (commented):
Are the smear values in the `MuonAnaAllRuns.csv` distribution file correct?
The azimuth bin-widths are ~0.15, computed from the bins, but the azimuth_smear is always 0.05. For zenith the values are ~0.039 and 0.0125.

kvjmistry:
> 

I will double-check that. I also launched some jobs yesterday to validate all these changes and will let you know if they look good or not. Thanks!

kvjmistry:
Generated three distributions: data file with angle option, sim file with angle option, sim file with angle and energy option. Here are the distributions in X-Z direction:

Data:
image

Sim (angle only):
image

Sim (Angle and Energy):
image
image

data/README.md
```diff
@@ -0,0 +1,14 @@
+# Folder Contents
+
+## 0nuBB files
+Bi214.genbb: File for generating Bi214 events  
+Tl208.genbb: File for generating Tl208 events  
+Xe136_bb0nu.genbb: File for generating 0nu double beta events  
+Xe136_bb2nu.genbb: File for generating 2nu double beta events
```
gondiaz: I don't think this is correct, they seem generated instead of generation files...

source/generators/MuonGenerator.cc
```diff
@@ -21,12 +21,8 @@
 #include <G4PrimaryVertex.hh>
 #include <G4Event.hh>
 #include <G4RandomDirection.hh>
-#include <Randomize.hh>
 #include <G4OpticalPhoton.hh>
 
-#include <TF1.h>
-#include <TMath.h>
-
 #include "CLHEP/Units/SystemOfUnits.h"
```
gondiaz: not needed, same for `using namespace CLHEP;` below

source/generators/MuonAngleGenerator.cc
```diff
@@ -39,7 +35,7 @@ REGISTER_CLASS(MuonAngleGenerator, G4VPrimaryGenerator)
 MuonAngleGenerator::MuonAngleGenerator():
   G4VPrimaryGenerator(), msg_(0), particle_definition_(0),
   angular_generation_(true), rPhi_(NULL), energy_min_(0.),
-  energy_max_(0.), distribution_(0), geom_(0), geom_solid_(0)
+  energy_max_(0.), geom_(0), geom_solid_(0), bInitialize_(false)
```
gondiaz: Add default value of `dist_name_` to "za". If it is not set the code breaks without any hint about the reason.

source/generators/MuonAngleGenerator.cc
```diff
@@ -25,10 +25,6 @@
 #include <Randomize.hh>
 #include <G4OpticalPhoton.hh>
 
-#include <TMath.h>
-#include <TFile.h>
-#include <TH2F.h>
-
 #include "CLHEP/Units/SystemOfUnits.h"
```
gondiaz, line 27: not needed, same for `using namespace CLHEP;` below

gondiaz (commented):
For consistency, remove execution file permissions in SimulatedMuonsProposalMCEqEnergy.csv.

gondiaz (approved):
This PR replaces the ROOT dependency on nexus, which was present in many parts of the code. The most relevant feature is the substitution of ROOT's `TFH` histograms by the Geant4's `G4RandGeneral` sampling function. This requires the change of the previously used root format to csv format. The PR also adds the muon flux files from simulations.
This PR overcomes #55, providing a long ago desired fix. Great work!

PR #147, opened 2022-03-09
Add missing variable for Scons compilation

paolafer:
In some Linux systems not loading the full environment of the user makes the building fail, with errors related to the inability of locating some libraries. This PR fixes this issue by loading  the `LD_LIBRARY_PATH` explicitly.


gondiaz (approved):
Just for the record: this modification won't be needed after this change [Add missing links in geant4-config script](https://github.com/Geant4/geant4/pull/37) is included in the next Geant4 11 patch.

PR #148, opened 2022-03-11
Add empty properties table

gondiaz:
Same as #141. It adds an empty material properties table to kapton and steel in order to avoid Geant4 bug in optical boundary process.

### DIFF
--- a/source/geometries/Next100EnergyPlane.cc
+++ b/source/geometries/Next100EnergyPlane.cc
@@ -196,9 +196,13 @@ namespace nexus {
 
     }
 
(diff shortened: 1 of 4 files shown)


paolafer:
We could probably define an empty material property table for all the materials exposed to optical photons in the NEXT-100 geometry, to prevent any possible problems.

gondiaz:
I think copper was the only missing material that could be hitted by optical photons

paolafer (approved):
This PR fixes a Geant4 issue with optical photons hitting materials with no material properties table defined. Approved!

PR #151, opened 2022-03-28
Fix angles in MuonGenerator

gondiaz:
The user can set the muon's momentum direction in the MuonGenerator to a fixed value.
However if the momentum is set by the user, the histograms provided by MuonEventAction (theta and phi angles) result in the default angle distributions of MuonGenerator class, instead of the fixed values provided by the user.
This comes from the fact that the momentum angles are not computed when it is set by the user.

### DIFF
--- a/source/actions/MuonsEventAction.cc
+++ b/source/actions/MuonsEventAction.cc
@@ -49,7 +49,7 @@ REGISTER_CLASS(MuonsEventAction, G4UserEventAction)
 
     // Muons Control plots
     hist1_ = new TH1D ("Edepo","Energy_deposited",100,-1.0,3.4);
-    hist2_ = new TH1D ("Theta","Theta generated",100,-pi,pi);
+    hist2_ = new TH1D ("Theta","Theta generated",100,0.,pi);
     hist3_ = new TH1D ("Phi","Phi generated",100,0.,twopi);
     tree_  = new TTree("Tree nexus","Flat tree with some nexus info");
     tree_->Branch("tree_theta", &tree_theta_, "tree_theta/D");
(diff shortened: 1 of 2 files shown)


kvjmistry:
This fix is correct. If we set the momentum direction manually, then we should also set the values of theta and phi to be consistent with this momentum direction, not a random sampled value. Theta also does not take a negative value, so the change from -pi -> 0 is fine.

paolafer:
@kvjmistry, if you are OK with the PR, please, approve it formally (you can go to the `File changes` tab, `Review changes` green button, `Approve` with a brief comment). 
Thanks!

source/actions/MuonsEventAction.cc
```diff
@@ -49,7 +49,7 @@ REGISTER_CLASS(MuonsEventAction, G4UserEventAction)
 
     // Muons Control plots
     hist1_ = new TH1D ("Edepo","Energy_deposited",100,-1.0,3.4);
-    hist2_ = new TH1D ("Theta","Theta generated",100,-pi,pi);
+    hist2_ = new TH1D ("Theta","Theta generated",100,0.,pi);
```
kvjmistry, line 52: there should be a 2*pi here I think.

gondiaz: why? phi here is azimuth which goes between 0 and 2*pi, and theta is zenith going from 0 and pi, no?

kvjmistry: Ah damn, your right. I confused my theta and phi around and with the comment added below!
Its all good

kvjmistry (approved):

Issue #137, opened 2022-02-10
Set default volume visibility in NEXT-100 geometry


gondiaz:
Solving this issue will provide you the hability to identify the structure of the nexus code that builds the geometry of a NEXT detector.

NEXT-100 geometry has a default visibility in nexus as shown below:

<img width="367" alt="next100-default-nexus-vis" src="https://user-images.githubusercontent.com/35993868/153388541-d1ee3127-f5d9-4490-8e2d-5ce71d558842.png">

The rendering time is very slow because the TPB coating inside the teflon mask holes in the tracking plane is being draw. This issue can be easily solved by modifying the default visibility of this volume to Invisible in the SiPM board geometry code.

With this issue I propose to change the default visibility above to another one that could help to visualice the inner elements of the detector, for example to something like or variant that might be useful for the user

<img width="367" alt="next100-ics-vis" src="https://user-images.githubusercontent.com/35993868/153389895-db13a22b-3c9d-4ef8-b54c-68a7ecee84e3.png">

**Hint/How-to-start:**

To run NEXT-100 default visibility go to nexus directory and run `./bin/nexus -i macros/NEXT100.init.mac` (Caution: this macro uses Next100OpticalGeometry instead of Next100, that does not include the external shielding geometry)

You can start modifying the energy plane visibilities, for example, in 

https://github.com/next-exp/nexus/blob/e3e292567cde778ed4c0dc89a2c9ead175d4fbeb/source/geometries/Next100TrackingPlane.cc#L145-L154

kvjmistry:
Tackling the first issue highlighted here.
Digging into the code, I already see that the Teflon mask holes are set to invisible in `Next100SiPMBoard.cc`:

`mask_wls_logic_vol     ->SetVisAttributes(G4VisAttributes::GetInvisible());`

I still however get the following error in my visualisation:

```
ERROR: G4VSceneHandler::RequestPrimitives
  Polyhedron not available for SIPM_BOARD_MASK_WALL_WLS
  Touchable path:  WORLD 0 LAB 0 LEAD_BOX 0 STEEL_BOX 0 INNER_AIR 0 VESSEL 0 VESSEL_GAS 0 SIPM_BOARD 1 SIPM_BOARD_MASK 0 SIPM_BOARD_MASK_HOLE 0 SIPM_BOARD_MASK_WALL_WLS 0
  This means it cannot be visualized in the usual way on most systems.
  1) The solid may not have implemented the CreatePolyhedron method.
[...]

PR #152, opened 2022-03-31
Change default NEXT-100 visibilities

kvjmistry:
This PR addresses issue #137 where the TPB coating on the teflon mask walls were being drawn and slowing down the NEXT-100 visualisation rendering. In addition, the default settings of the NEXT-100 visibilities have been changed so that the inner parts of the detector are more clear. I have added colours in the Visibility.hh file with an alpha setting to select colours with a level of transparency. 

Image below shows the NEXT-100 geometry with the following settings switched on:
/Geometry/Next100/tracking_plane_vis true
/Geometry/Next100/energy_plane_vis true
/Geometry/Next100/field_cage_vis true
/Geometry/Next100/sipm_board_vis true
/Geometry/Next100/sipm_vis true
/Geometry/PmtR11410/visibility true
/Geometry/Next100/vessel_vis true
/Geometry/Next100/ics_vis true
image

The inner field cage looks like this:
image

### DIFF
--- a/source/geometries/Next100EnergyPlane.cc
+++ b/source/geometries/Next100EnergyPlane.cc
@@ -351,6 +351,7 @@ namespace nexus {
     //////////////////////
 
     vacuum_logic->SetVisAttributes(G4VisAttributes::GetInvisible());
+    pmt_base_logic->SetVisAttributes(G4VisAttributes::GetInvisible());
     if (visibility_) {
       G4VisAttributes copper_col = CopperBrownAlpha();
       copper_col.SetForceSolid(true);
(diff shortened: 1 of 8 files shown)


gondiaz (commented):
This is the default visibility, if none of the commands are used:
default

Notice that tracking and energy planes are being draw, and some of their elements. I think the default visibility should be something simpler, showing for example the dice boards, the PMTs and the anode, cathode and gate in case you wish to visualize an event at the active volume.
In the above figure I would remove the db-plugs and the gas holes in the energy plane (vacuum_logic, drawn as a wireframe). Also kapton board and mask are being draw (with blue and light-blue) and they are difficult to distinguish, I would only draw the mask since the kapton board is much thinner.
The visibility using the commands looks good, but I would remove the db-plugs and the vacuum_logic...

kvjmistry:
More updates to default visibility NEXT-100 geometry:
image

gondiaz (approved):
I removed the PMT transparency in the last commit, since there is no volumes inside the PMTs I find it clearer...

This PR updates the default visibility of Next100 detector, avoiding a slow rendering of the image by setting the wall's tpb volume to invisible. It also modiifies other volumes. Good work!

PR #153, opened 2022-04-01
Add number of interacting events to output file

paolafer:
I've realized that in `SaveAllEventAction` the information about the event having interacted or not is not saved. However, in the `configuration` table the field `interacting_events` always appears, and in this case it has the value `0`.
Since this can be misleading, this PR adds the information about events interacting in the detector to the `SaveAllEventAction`, too.


paolafer:
This PR has become obsolete, since the action has been removed.

PR #154, opened 2022-04-11
Remove subtraction holes in NEXT100 EP

gondiaz:
This PR removes unnecessary subtraction holes in the NEXT100 EP-copper-plate.
This change reduces by half the simulation time of backgrounds from the EP vertices (copper-plate, pmt and pmt-bases)

### DIFF
--- a/source/geometries/Next100EnergyPlane.cc
+++ b/source/geometries/Next100EnergyPlane.cc
@@ -18,7 +18,6 @@
 #include <G4Material.hh>
 #include <G4LogicalVolume.hh>
 #include <G4Tubs.hh>
-#include <G4SubtractionSolid.hh>
 #include <G4UnionSolid.hh>
 #include <G4OpticalSurface.hh>
 #include <G4LogicalSkinSurface.hh>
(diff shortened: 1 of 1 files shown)


source/geometries/Next100EnergyPlane.cc
```diff
@@ -105,20 +104,8 @@ namespace nexus {
     GeneratePositions();
 
     /// Copper Plate ///
-    G4Tubs* copper_plate_origin_solid =  new G4Tubs("COPPER_PLATE_ORIGIN", 0., copper_plate_diam_/2.,
-                                                    copper_plate_thickn_/2., 0., twopi);
-
-    /// Hole for gas flow ///
-    G4double gas_hole_diam   = 12.  * mm;
```
paolafer, line 112: Removing this hole changes the volume. Have you assessed the impact on the background simulation? Also, there may be gammas penetrating through the hole that now would be blocked. I'm note sure we should remove it.

gondiaz:
As talked offline, the reason for the holes to be subtracted from the copper-plate is needed to place the pmts, since they stand out 2 mm above the plate edge. The change tried in this PR requires a study on the effect of this placement with simulation of optical photons, since detection probability might change. Also, there is a bug in this PR since volumes overlap, as vacuum volume placement must be changed from mother to copper-plate volume.

Since simulation times could be greatly reduced, I propose to move this PR's aims to an issue for future geometry improvements.

PR #157, opened 2022-04-20
Fix EJ286 bug

jurenag:
I noticed a bug within opticalprops::EJ286() regarding its WLS absorption length.


jmalbos (changes requested):
Thanks for submitting this bug fix, @jurenag!
Before merging it, could you please elaborate a bit more on the commit message, for future reference?

jurenag:
The bug I noticed affected the optical wavelength shifting (WLS) absorption length of opticalprops::EJ286(). Within some of the energy array entries, the wavelength value is divided by its unit (nm) instead of being multiplied by it.

jmalbos (approved):

PR #158, opened 2022-04-20
Fix Trajectory visualization bug

### LINKED ISSUE #9: Fix our TrackingActions to allow the visualization of photons and geantinos from the very beginning

jurenag:
As pointed out by issue #9, the trajectories of photons and geantinos cannot be visualized from the very beginning.
With this commit, the creation of the first trajectory point within Trajectory class constructor is added, thus resolving the incompleteness of the visualized trajectories, which actually affected every trajectory regardless its track.

Resolves: #9

Find below the effects of this change on the configuration given by macros/nexus_example1.init.mac , after executing run/beamOn 100

Before this commit:
image

After this commit:
image

### DIFF
--- a/source/base/Trajectory.cc
+++ b/source/base/Trajectory.cc
@@ -43,6 +43,10 @@ Trajectory::Trajectory(const G4Track* track):
   initial_volume_ = track->GetVolume()->GetName();
 
   trjpoints_ = new TrajectoryPointContainer();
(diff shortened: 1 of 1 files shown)


jmalbos (approved):
This pull request fixes a long-standing bug in the visualisation of trajectories in NEXUS. Well done, @jurenag!

PR #159, opened 2022-04-21
Interactive mode signal interrupt

gondiaz:
A nexus process in interactive mode is killed by an interrupt (Ctrl-C). Nevertheless memory is not explicitly deallocated, leaving a corrupt output file. This PR solves this issue by using a `signal`.

### DIFF
--- a/source/nexus.cc
+++ b/source/nexus.cc
@@ -13,9 +13,11 @@
 #include <G4VisExecutive.hh>
 
 #include <getopt.h>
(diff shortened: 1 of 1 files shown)


jmalbos:
Use the `exit` command instead of ctrl-C. That way the application closes correctly, generating a valid output file.

gondiaz:
Great, I wasn't aware of this command..

PR #160, opened 2022-04-21
Clean up actions

paolafer:
This PR cleans up the code of the event actions, eliminating obsolete code and  the `SaveAllEventAction` class. To save all the events in a run one should simply use the Geant4 default event action, which is achieved by removing any event action command from the configuration file.
In addition, the persistency manager has been modified so that the number of interacting events is saved only if `DefaultEventAction` is used, because it's the only case where the variable used to store the number of interacting events is meaningful. In any other case, that field will not appear in the output file.


gondiaz (approved):
This PR slighty modifies the DefaultEventAction: removes uneeded call for drawing trajectories since it can be done by a Geant4 command, renames config parameter for minimum energy and adds the number of interacting events to the output file (in addition the saved and total events). It also removes the SaveAllEventAction since it is equivalent to the default event action of Geant4. Well done!

PR #162, opened 2022-04-21
Protect against failed cast

gondiaz:
In interactive mode, you might want to use **DefaultEventAction** without **DefaultTrackingAction**. Since you are in interactive mode, Geant4 will provide a not empty `G4TrajectoryContainer`, which is a vector of `G4VTrajectory`. The dowcast fails since `Trajectory` is a nexus class.

### DIFF
--- a/source/actions/DefaultEventAction.cc
+++ b/source/actions/DefaultEventAction.cc
@@ -86,10 +86,21 @@ REGISTER_CLASS(DefaultEventAction, G4UserEventAction)
 
       G4TrajectoryContainer* tc = event->GetTrajectoryContainer();
       if (tc) {
+        // in interactive mode, a G4TrajectoryContainer would exist
+        // but the trajectories will not cast to Trajectory
+        Trajectory* trj = dynamic_cast<Trajectory*>((*tc)[0]);
+        if (trj == nullptr){
+          G4Exception("[DefaultEventAction]", "EndOfEventAction()", FatalException,
+                      "DefaultTrackingAction is required when using DefaultEventAction");
+        }
         for (unsigned int i=0; i<tc->size(); ++i) {
           Trajectory* trj = dynamic_cast<Trajectory*>((*tc)[i]);
           edep += trj->GetEnergyDeposit();
         }
+      }
+      else {
+        G4Exception("[DefaultEventAction]", "EndOfEventAction()", FatalException,
+                    "DefaultTrackingAction is required when using DefaultEventAction");
       }
 
       PersistencyManager* pm = dynamic_cast<PersistencyManager*>


paolafer:
I would say that in the interactive mode there is no need to use `DefaultEventAction` either, since one is not concerned with how many events are saved in the output file. One could simply use the default actions by not specifying any custom actions.
In fact, if one wanted to save only events with energy deposition above a threshold, and at the same time is not using `DefaultTrackingAction`, it wouldn't work, because the calculated energy deposited would be always zero.  I will document the use of each action better, once the PR on the clean-up of the actions is approved.

paolafer:
What we could do is, instead of adding a protection, adding a meaningful error message and let the program crash.

gondiaz:
Agree. What do you think about placing such an error before the `SetUserAction` in `NexusApp.cc`? This would assure that you always choose a compatible set of user actions

paolafer:
> Agree. What do you think about placing such an error before the `SetUserAction` in `NexusApp.cc`?

I don't like it vey much, I would prefer to keep the code in the  `NexusApp` class as simple as possible. I would rather put the protection in `DefaultEventAction`, which is the class that accesses the `Trajectory object`.

paolafer:
@gondiaz, what about this PR? Are you going to add the comment to the crash or shall we close it?

gondiaz:
At least not in the next two weeks

paolafer:
> At least not in the next two weeks

This doesn't answer my question :-).

gondiaz:
The answer is yes then, but I don't know when. 
If it is annoying to have this pending, it can be closed. If anyone wants to contribute and finish it, also works for me.

paolafer:
It's not annoying, I just want to know if you forgot about it or you are going to finish it at some point.

source/actions/DefaultEventAction.cc
```diff
@@ -86,11 +86,22 @@ REGISTER_CLASS(DefaultEventAction, G4UserEventAction)
 
       G4TrajectoryContainer* tc = event->GetTrajectoryContainer();
       if (tc) {
+        // in interactive mode, a G4TrajectoryContainer would exist
+        // but the trajectories will not cast to Trajectory
+        Trajectory* trj = dynamic_cast<Trajectory*>((*tc)[0]);
+        if (trj == nullptr){
+          G4Exception("[DefaultEventAction]", "EndOfEventAction", FatalException,
```
paolafer: In the nexus convention, the second argument of `G4Exception` should be the function, so `()` should be added.

source/actions/DefaultEventAction.cc
```diff
@@ -86,11 +86,22 @@ REGISTER_CLASS(DefaultEventAction, G4UserEventAction)
...
+                      "Adequate TrackingAction not registered for DefaultEventAction");
```
paolafer: I would say explicitly which one is the correct action, something like "DefaultTrackingAction is required when using DefaultEventAction".

paolafer (changes requested):

paolafer (approved):
This PR adds a useful error message when `DefaultTrackingAction` is not used together with `DefaultEventAction`. Approved!

PR #164, opened 2022-04-27
Move function definitions where the smart pointer is complete

paolafer:
I've noticed that a compilation error is prompted when nexus is used as a third party library by an external software. The reason of the error is that we use a forward declaration of the smart pointer in `DetectorConstruction.h` and we use the move assignment operator in an inline function in the same header file. However, when the move assignment operator is used, the smart pointer needs to be complete, which is not the case in DetectorConstruction.h, because of forward declaration.
Therefore, this PR moves the definitions of the functions to the `.cc` file.

### DIFF
--- a/source/base/DetectorConstruction.cc
+++ b/source/base/DetectorConstruction.cc
@@ -73,3 +73,14 @@ G4VPhysicalVolume* DetectorConstruction::Construct()
 
   return world_physi;
 }
+
+
+void DetectorConstruction::SetGeometry(std::unique_ptr<GeometryBase> g)
+{
+  geometry_ = std::move(g);
+}
+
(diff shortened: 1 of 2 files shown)


paolafer:
This PR is becoming urgent, can the designated reviewers (@jmalbos and @gondiaz) deal with it before the end of the week? Otherwise I would ask @gonzaponte to step in if he's available.

gondiaz:
I wont be able to make it soon...

jmalbos:
> This PR is becoming urgent, can the designated reviewers (@jmalbos and @gondiaz) deal with it before the end of the week? Otherwise I would ask @gonzaponte to step in if he's available.

I'll do it by Friday.

jmalbos (approved):

PR #165, opened 2022-04-27
Simulate muons from above

gondiaz:
Muons are simulated by default with downward direction, therefore they should come from above the detector.

### DIFF
--- a/source/generators/MuonAngleGenerator.cc
+++ b/source/generators/MuonAngleGenerator.cc
@@ -33,7 +33,7 @@ REGISTER_CLASS(MuonAngleGenerator, G4VPrimaryGenerator)
 MuonAngleGenerator::MuonAngleGenerator():
   G4VPrimaryGenerator(), msg_(0), particle_definition_(0),
   angular_generation_(true), rPhi_(NULL), energy_min_(0.),
-  energy_max_(0.), geom_(0), geom_solid_(0), bInitialize_(false), dist_name_("za")
+  energy_max_(0.), dist_name_("za"), bInitialize_(false), geom_(0), geom_solid_(0)
 {
   msg_ = new G4GenericMessenger(this, "/Generator/MuonAngleGenerator/",
 				"Control commands of muongenerator.");
--- a/source/generators/MuonAngleGenerator.cc
+++ b/source/generators/MuonAngleGenerator.cc
@@ -81,10 +81,10 @@ MuonAngleGenerator::~MuonAngleGenerator()
 
 void MuonAngleGenerator::LoadMuonDistribution()
 {
-  
+
   // File pointer
   std::ifstream fin(ang_file_);
-  
+
   // Check if file has opened properly
   if (!fin.is_open() && angular_generation_){
     G4Exception("[MuonAngleGenerator]", "LoadMuonDistribution()",
@@ -94,7 +94,7 @@ void MuonAngleGenerator::LoadMuonDistribution()
   // Check the input filename for the keyword Energy
   size_t found = ang_file_.find("Energy");
   if (found!=std::string::npos){
-    
+
     // Only angle option, but energy specified
     if (dist_name_ == "za")
       G4Exception("[MuonAngleGenerator]", "LoadMuonDistribution()",
@@ -167,11 +167,6 @@ void MuonAngleGenerator::SetupAngles()
   // Rotates anticlockwise about Y.
   rPhi_ = new G4RotationMatrix();
   rPhi_->rotateY(-axis_rotation_);
-
-  // Get the solid to check overlap
-  geom_solid_ =
-    geom_->GetLogicalVolume()->GetDaughter(0)->GetLogicalVolume()->GetSolid();
-
 }
 
 
@@ -179,8 +174,8 @@ void MuonAngleGenerator::GeneratePrimaryVertex(G4Event* event)
 {
 
   // Initalise RNG seeds and load file on the first event
-  if (!bInitialize_){ 
-    
+  if (!bInitialize_){
+
     // Load in the Muon angular distribution from file
     if (angular_generation_)
       LoadMuonDistribution();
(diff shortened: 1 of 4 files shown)


source/generators/MuonAngleGenerator.cc
```diff
@@ -216,7 +216,7 @@ void MuonAngleGenerator::GeneratePrimaryVertex(G4Event* event)
   // Overwrite default p_dir, zenith and azimuth from angular distribution file
   if (angular_generation_){
     GetDirection(p_dir, zenith, azimuth, energy, kinetic_energy, mass);
-    while ( !CheckOverlap(position, p_dir) )
+    while ( !CheckOverlap(position, p_dir) || position.getY()<0)
```
kvjmistry: My summary of PR: 
Muons by default are generated to be downwards going. Muons with a vertex y position below the detector are pointless to generate, so this PR is addressing this issue.

I am assuming y = 0 is at the centre of the detector, if this is the case, then wouldn't events generated in the active volume with a vertex in the lower half of the detector be removed. Is this the desired behaviour? 

If events are generated in the world volume, this would be ok though.

gondiaz:
I have modified the previous implementation: now the lab-wall geometry only consists of a semi cilinder, and the generation is done in this part. The detector is properly placed inside this cilinder, see figure (assuming the previous values of y and z translations are correct).
Lab walls

On the other hand, MuonAngleGeneration uses a CheckOverlap function that was biasing the vertex generation positions. Since it doesnt seems relevant and creates a significant issue in the simulation, I removed it.

paolafer:
The overlap check was done to avoid generating particles that don't cross the geometry, therefore wouldn't create a useful event. Which bias do you see?

source/geometries/LSCHallA.h
```diff
@@ -37,8 +37,6 @@ namespace nexus {
     G4double GetLSCHallALength() const;
     G4double GetLSCHallARadius() const;
     G4double GetLSCHallAWallThickn() const;
-    G4double GetLSCHallACastleZ() const;
```
paolafer: Why are you removing this?

source/geometries/Next100.cc
```diff
@@ -154,8 +154,10 @@ namespace nexus {
 
     G4ThreeVector gate_pos(0., 0., -gate_zpos_in_vessel_);
     if (lab_walls_){
-      G4ThreeVector castle_pos(0., hallA_walls_->GetLSCHallACastleY(),
```
paolafer: Why are you replacing the castle position, extracted from the geometry, with some hard-coded value?

gondiaz:
About the overlap: you can see it in the `master`, using the angular distribution file in nexus (`SimulatedMuonsProposalMCEq.csv`) and the `HALLA_OUTER` vertex. In the two figures above I show the vertex generated using the overlap check, and removing it, respectively. The generation is expected to be uniformly distributed in the full cilinder that represents the HallA walls, but it is not the case when the check-overlap.

with_check

without_check

source/geometries/Next100.cc
gondiaz: Not a real reason, just a remanent of the checks I was doing during the development...

paolafer: I would go back to the original version, then.

paolafer:
> About the overlap: you can see it in the `master`, using the angular distribution file in nexus (`SimulatedMuonsProposalMCEq.csv`) and the `HALLA_OUTER` vertex. In the two figures above I show the vertex generated using the overlap check, and removing it, respectively. The generation is expected to be uniformly distributed in the full cilinder that represents the HallA walls, but it is not the case when the check-overlap.
> 
> with_check
> 
> without_check

The check overlap should avoid generating muons with a direction that does not intersect the geometry, therefore I expect that it will bias the vertices. From your plot I'm not sure if the missing vertices are the ones that would not intersect the geometry. The question is if we want to save time doing that or we need to generate all the muons for accounting, for instance, or any other reason.

paolafer:
Change the order of the initialization of variables, to avoid compilation warnings. Specifically, `geom_solid_`, `bInitialize_` and  `dist_name`.

source/geometries/LSCHallA.cc
```diff
@@ -91,16 +89,15 @@ namespace nexus {
...
     hallA_vertex_gen_ =
       new CylinderPointSampler2020(lab_radius_ - 0.1 * mm,
 				   lab_radius_, (lab_length_ - 0.1 *mm) /2.,
-				   0., twopi, nullptr, hall_centre);
+				   0., pi, nullptr, G4ThreeVector(0., 0., 0.));
     hallA_outer_gen_ =
       new CylinderPointSampler2020(lab_radius_ + lab_wall_thickn_ + 0.1 * mm,
 				   lab_radius_ + lab_wall_thickn_ + 0.2 * mm,
 				   lab_length_/2. + lab_wall_thickn_ + 0.1 * mm,
-				   0, twopi, nullptr, hall_centre);
```
paolafer: The HallA geometry is not placed in `(0, 0, 0)`, therefore, if you centre the vertex generator in `(0, 0, 0)`, the vertices will not be generated correctly in the inside or outside of the rock, right?

gondiaz: that's right, good catch

source/geometries/Next100.cc
```diff
@@ -154,13 +154,15 @@ namespace nexus {
...
+      G4ThreeVector castle_pos(0., hallA_walls_->GetLSCHallACastleY() + shielding_->GetHeight(),
```
paolafer: What's the reason of adding the height of the castle?

gondiaz: To see this you can assume that `hallA_walls_->GetLSCHallACastleY()` is zero, then you want the floor of the lead-castle to be aligned with the floor of the lab. In fact, instead of being the height it should be height/2.
castle_floor

kvjmistry:
On removing the checking overlap, I think this is fine since we don't want all muons to intersect the detector. I am sure that we want to generate some that interact near the detector to model the daughter neutrons which may make it inside the active. We probably want to set a minimum distance because if the muons are too far away to create any backgrounds reaching the active, then these events are simply inefficient to make. However, this would require some optimisation work. 

The suggested generation from a semi cylinder looks good to me.

paolafer (approved):
This PR improves the efficiency of the muon simulation. Approved!

PR #166, opened 2022-04-27
Simplify shielding volume names

gondiaz:
This PR simplifies the name of the shielding-struct and pedestal physical volumes. The current names might cause confusion (for example in slide8 [here](https://next.ific.uv.es/DocDB/0013/001327/001/Cosmo_NEXT100_March22.pdf)) since these volumes are splited in several physical volumes with different names. I assign the same name to them as the name of its associated vertex generator.


paolafer (approved):
The change is fine with me. Approved.

PR #168, opened 2022-05-18
Fix segmentation fault due to G4 bug by setting G4MaterialPropertiesTable

DiegoLopezGutierrez:
Similar to #141 and #148, this PR adds an empty G4MaterialPropertiesTable to materials that lack it (e.g. copper, teflon, steel, etc). The implementation is for all the geometries not covered by the previous two PRs (i.e. all except NEXT100).

### DIFF
--- a/source/geometries/Next1EL.cc
+++ b/source/geometries/Next1EL.cc
@@ -255,19 +255,25 @@ void Next1EL::DefineMaterials()
 {
   // AIR
   air_ = G4NistManager::Instance()->FindOrBuildMaterial("G4_AIR");
+  air_->SetMaterialPropertiesTable(new G4MaterialPropertiesTable());
   // GASEOUS XENON
   gxe_ = materials::GXe(pressure_, 303);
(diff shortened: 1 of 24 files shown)


source/geometries/Next1EL.cc
```diff
@@ -955,8 +963,11 @@ void Next1EL::BuildFieldCage()
     new G4UnionSolid("SUPPORT_BAR", bar_base, addon, 0,
 		     G4ThreeVector(bar_thickn_, 0., (bar_length - bar_addon_length_)/2.));
 
+  G4Material* peek_mat_ = materials::PEEK();
```
paolafer: In [nexus conventions](https://github.com/next-exp/nexus/wiki/Coding-conventions-and-style-guide), we use the trailing underscore for members of the classes. In this case, since you are defining a local variable, you should just use `peek_mat`. Please, review the rest of the code following this convention.

DiegoLopezGutierrez: I have now added a commit with the changes.

paolafer (approved):
This PR fixes the Geant4 issue with optical photons hitting materials with no material properties table defined, for the rest of NEXT geometries other than NEXT-100. Approved!

PR #170, opened 2022-05-25
Fix wrong distances in NEXT100 sipm boards

gondiaz:
NEXT100 sipm-board size is 1 mm larger and distance between boards is 1 mm instead 2 mm (see attached drawings)

[masks.pdf](https://github.com/next-exp/nexus/files/8768911/03-03.pdf)
[tracking_plane.pdf](https://github.com/next-exp/nexus/files/8768919/00-03-00.Tracking.Plane.Shielding.pdf)


paolafer (approved):
The dimensions have been changed according to the the latest drawings, approved.

PR #171, opened 2022-06-15
Dist sampler

kvjmistry:
This PR creates two classes in utilities: FileReaderCSV and DistributionSampler. They contain generic functions used in the muon angle generator so any 1D, 2D and 3D input distribution (as a csv file) can be sampled from to produce a variable that follows the input distribution. 

The muon angle generator class has been updated to use these functions as a use case.

Also add the option to not smear the sampled events controlled with a bool in case we want to generate events without smearing.


source/generators/MuonAngleGenerator.cc
```diff
@@ -107,80 +99,16 @@ void MuonAngleGenerator::LoadMuonDistribution()
...
-  // Check if the specified energy range has been set to a suitable value
-  if (dist_name_ == "zae" && (energy_min_ < file_emin*GeV || energy_max_ > file_emax*GeV )){
-    std::cout << "The minimum energy allowed is: " << file_emin*GeV/1000 << " GeV"<< std::endl;
-    std::cout << "The maximum energy allowed is: " << file_emax*GeV/1000 << " GeV" << std::endl;
-    G4Exception("[MuonAngleGenerator]", "LoadMuonDistribution()",
-              FatalException, " Specified energy range for sampling is outside permitted range or min_energy/max_energy has not been set");
-    
   }
 
   // Convert flux vector to arr
```
paolafer: The compiler issues a warning about using a variable length array in line 115. I think that the correct way to transform a vector in an array is: `auto arr_flux = flux_.data();` (see, for instance, https://en.cppreference.com/w/cpp/container/vector/data).
Have you tried it?

paolafer (changes requested):
In line 201, you can avoid writing the `RandomEnergy()` function, and use `UniformRandomInRange()` from `source/utils/RandomUtils.cc`.

Also, I think that an explanation of the data file format could be useful, maybe in a README file in the data folder, or in the file itself, if that doesn't complicate the reading too much.

source/utils/DistributionSampler.h
```diff
@@ -0,0 +1,50 @@
...
+#ifndef DISTRIBUTION_SAMPLER
+#define DISTRIBUTION_SAMPLER
+
+#include <Randomize.hh>
+#include <G4Types.hh>
+
+namespace nexus {
+
+  /// Generator of random points in a sphere
+  class DistributionSampler
```
paolafer: Do we need to have a class for this? Maybe we could just use the functions with a specific namespace (such as `sampler`, or `distrsampler`). Same for `FileReaderCVS`.

paolafer (changes requested):
If the `angles_on` parameter is false, muons are simulated coming in the (0, 1, 0) direction: is that the expected behaviour? Is there any real case where we want that behaviour?

source/utils/DistributionSampler.h
paolafer: In fact, I'm thinking that `GetRandBinIndex()` and `Sample()` may well belong to the `RandomUtils` collection of functions, what do you think? The `CheckOutOfBound` functions are very general and they may be used also in a different context than distribution sampling, so they may belong to a list of general functions.

paolafer:
I've realized that there are two sets of example macros using the `MuonAngleGenerator` class: one with the `hallA` label and the other one with the `lsc` label. The only difference I see (apart from the kind of simulation file used - although in both the other one is present, commented) is the region of generation and the `angles_on` parameter. In the former case, it is the hall A walls, and in the latter a surface just outside the lead castle.  I understand that in this second case the file with the angle distribution is not necessary in the example, right?
I would either make two different examples, with a different description to explain their use, or eliminate one of the two examples and add a commented line with the second option for the region, provided that we use it currently in some cases. @pnovella is the surface outside the castle (`EXTERNAL` region) without angle sampling used currently in the cosmogenic simulation?

kvjmistry:
> If the `angles_on` parameter is false, muons are simulated coming in the (0, 1, 0) direction: is that the expected behaviour? Is there any real case where we want that behaviour?

Right, this just simulated muons coming from top down. I can't think of anyone using this case for something useful. 
I am thinking when the angles_on is false, we should actually be using the MuonGenerator class functionality, but its in a different class. I wonder if it makes sense to merge these two classes where we use the MuonGenerator.cc code for angles_on = false and the existing MuonAngleGenerator.cc code for when angles_on = true. What do you think?

source/generators/MuonAngleGenerator.cc
kvjmistry: Yes this also works, I have added a commit that updates this.

paolafer:
> > If the `angles_on` parameter is false, muons are simulated coming in the (0, 1, 0) direction: is that the expected behaviour? Is there any real case where we want that behaviour?
> 
> Right, this just simulated muons coming from top down. I can't think of anyone using this case for something useful. I am thinking when the angles_on is false, we should actually be using the MuonGenerator class functionality, but its in a different class. I wonder if it makes sense to merge these two classes where we use the MuonGenerator.cc code for angles_on = false and the existing MuonAngleGenerator.cc code for when angles_on = true. What do you think?

I think it's a good idea. We could keep just one class (`MuonGenerator` seems the clearest name to me) and select if we want the sea level, no shielding or the LSC angular distribution via a parameter. I would remove the possibility of the (0, -1, 0) direction, if it's not used. Maybe the parameter could be called `lsc`, instead of `angles_on`, to make it clearer that it is the distribution at the LSC. In any case, a few lines of comments can also be placed in the code.

kvjmistry:
Latest changes include:
- Merging Muon Generator and Muon Angle Generator Classes
- Move Dist Sampler class functions to Random Utils, remove dist sampler class
- Rename FileSampler Class to IOUtilities
- Update config files so they are documented for running the latest muon generator class
- Make sure angles are named consistently zenith and azimuth in the code (some places had theta, phi)
- Change angles_on config var to use_lsc_dist to be clearer that this mode uses an angular distribution from a file for the distribution measured/simulated at lsc

source/generators/MuonGenerator.cc
```diff
@@ -50,13 +53,77 @@ MuonGenerator::MuonGenerator():
...
+
+  msg_->DeclareProperty("angle_file", ang_file_,
+			"Name of the file containing angular distribution.");
+  msg_->DeclareProperty("angle_dist", dist_name_,
+			"Name of the angular distribution histogram.");
 
+  msg_->DeclareProperty("user_dir",  user_dir_, "Set fixed muon direction.");
+
+  G4GenericMessenger::Command& rotation =
+    msg_->DeclareProperty("azimuth_rotation", axis_rotation_,
```
paolafer, line 68: The `axis_rotation_` variable should be initialized, either to 0 or to 150 deg, since it's the value we'll be using mostly. What do you think it's best?

source/generators/MuonGenerator.cc
```diff
@@ -13,24 +12,28 @@
...
 using namespace nexus;
 
 REGISTER_CLASS(MuonGenerator, G4VPrimaryGenerator)
 
-
 MuonGenerator::MuonGenerator():
   G4VPrimaryGenerator(), msg_(0), particle_definition_(0),
-  energy_min_(0.), energy_max_(0.), geom_(0), momentum_{}
+  use_lsc_dist_(true), rPhi_(NULL), energy_min_(0.),
+  energy_max_(0.), dist_name_("za"), user_dir_{}, bInitialize_(false), geom_(0), geom_solid_(0)
```
paolafer: The compiler issues a warning because `dist_name_` is declared after `user_dir`, but is initialized before. In order to avoid this kind of messages it's good practice to keep the same order in the header and source file.

source/generators/MuonGenerator.cc
```diff
@@ -66,105 +133,200 @@ MuonGenerator::MuonGenerator():
...
-  particle_definition_ = G4ParticleTable::GetParticleTable()->FindParticle(MuonCharge());
+
+  // Initalise RNG seeds and load file on the first event
+  if (!bInitialize_){
+
+    // Load in the Muon angular distribution from file
+    if (use_lsc_dist_){
+      std::cout << "[MuonGenerator]: Generating muons using lsc distribution loaded from file" << std::endl;
+      LoadMuonDistribution();
+    }
```
paolafer, line 169: As a protection against a possible user's mistake of setting both a specific direction and `use_lsc_dist` to `true`, we could raise an error after line 161, `if user_dir_ != G4ThreeVector{}`.

source/utils/IOUtilities.cc
```diff
@@ -0,0 +1,230 @@
...
+
+#include "IOUtilities.h"
+
+#include <G4ExceptionSeverity.hh>
+#include <G4Exception.hh>
+
+
+namespace nexus {
+
+  IOUtilities::IOUtilities(){}
```
paolafer: Do we really need a class here or could we just use functions, as in `RandomUtils`?

macros/NEXT100_muons.config.mac
```diff
@@ -23,23 +22,37 @@
 /Geometry/Next100/pressure 15 bar
 /Geometry/Next100/gas enrichedXe
 /Geometry/Next100/elfield false
-/Geometry/Next100/lab_walls true
```
paolafer: I would keep this line, although commented, since it is necessary when using `HALLA_OUTER` as a vertex region.

macros/ton_muons_vertical.config.mac
```diff
@@ -29,13 +28,22 @@
...
+
+# User specified muon direction. Must set use_lsc_dist to false for this to work
+/Generator/MuonGenerator/user_dir 0 -1 0
+
+# For coordinate system transformation, do not edit
+/Generator/MuonGenerator/azimuth_rotation 150 deg
 
 ### ACTIONS
 /Actions/DefaultEventAction/min_energy 0.01 MeV
+#/Actions/MuonsEventAction/stringHist MuonsDistribution.root
```
paolafer: I would use `csv` as a format in the example, to make it as general as possible (so that one doesn't need to install ROOT to read it). The same applies to `NEXT100_muons.config` and `DEMO_muons.config`.

paolafer (changes requested):

kvjmistry:
- Fixed header and source of variables in the constructor to fix warning message
- Use csv as default file format in config files
- Move IO Utility class functions to Random Utils. Delete the IO class
- Add exception clause for user dir
- Add back in deleted line in config file
- Set default axis rotation in constructor

paolafer (approved):
This PR introduces generic functions to read distributions and generate random numbers sampling from them. The muon generator has been modified accordingly and it is a good example of how to use them. Finally, the two existing muon generators have been unified into one, to avoid repetition of code and to make it clearer to the user. 
It's been a big amount of work, which has improved significantly the code: excellent job, @kvjmistry! I approve it!

paolafer:
@jmalbos, you were the original reviewer of this PR: do you have any further comments?

PR #174, opened 2022-07-11
Add source info for sapphire optical properties.

MiryamMV:
This PR adds the link where the refractive index info of sapphire was extracted from.


paolafer (approved):
Simple PR that improves documentation. Approved.

PR #175, opened 2022-07-12
Eliminate hadronic physics verbosity, to clean up automatic tests

paolafer:
This PR sets the verbosity of the hadronic physics to zero in the example macros, to avoid having a lot of output in the automatic tests, which makes difficult to understand the error when they fail. The electromagnetic verbosity was already set to zero, but in previous versions of Geant4 the hadronic one didn't work properly (that's why the command was commented in the examples). Starting from geant4.11 the command works fine and it has been introduced in all the macros that use hadronic physics lists.

### DIFF
--- a/macros/NEXT100.Neutron.config.mac
+++ b/macros/NEXT100.Neutron.config.mac
@@ -13,7 +13,7 @@
 /tracking/verbose 0
 
 /process/em/verbose 0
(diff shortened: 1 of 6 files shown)


FabianKellerer (approved):
All tests passed successfully on a local machine. Approved.

PR #176, opened 2022-07-14
Automatic build and test using docker

jmbenlloch:
This PR adds a new mechanism for continuous integration using docker. The total size of the image including `Geant4`, `Root` and the `conda` environment is ~3.8GB.

It is built in several steps, the `docker` folder contains the definitions for several images:

- Geant4 11.0.2
- Root 6.26.04
- Miniconda with the corresponding environment
- A `nexus-base` image with all the compiled dependencies. 

The new github action included builds `nexus` and run the tests. An example can be seen [here](https://github.com/jmbenlloch/nexus/runs/7327955513?check_suite_focus=true).

When an update is needed, the corresponding `Dockerfile` will need to be updated, as well as the image in the registry. Now we are using https://registry.cern.ch, which offers no limitations in terms of pulling images.

The docker image could be used directly in several HPC clusters. In those not supporting docker, singularity is usually an option and singularity can build an image from a docker container.

The next step would be to modify the github action, so on every merge to master, a new docker image gets compiled and pushed to the registry automatically. In that way, we would have a pre-compiled version for every status of the master branch. That could be pulled from any production machine.

The size of the new images will be small since only the last layer would be updated (this repo and the compiled version of nexus).

### DIFF
--- a/.github/workflows/install-via-docker-and-test.yml
+++ b/.github/workflows/install-via-docker-and-test.yml
@@ -1,0 +1,17 @@
+name: Install via docker and Test
+on:
+  pull_request:
+  push:
+jobs:
+  build-and-test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-22.04]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Compile software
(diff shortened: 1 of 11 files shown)


paolafer (approved):
This PR adds the possibility of doing CI with docker.  Approved.

PR #177, opened 2022-07-14
Remove nix workflow

jmbenlloch:
This PR removes the old nix workflow that was not working anymore and it is already replaced by a new docker-based one (#176).


paolafer (approved):
Last bit to remove nix.

PR #178, opened 2022-07-15
Optical fibre2

FabianKellerer:
Added the OpticalFibre geometry to perform studies with optical fibres. Added the GenericWLSFiber geometry to use in the NextFlex geometry as individual fibres.


.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,16 @@
+{
```
paolafer: These three `.json` files don't belong to this PR. You can remove this commit, if you don't know how to do it I can show you.

.gitignore
```diff
@@ -25,6 +25,7 @@ nexus
 nexus-test
```
paolafer: Why have you added this? Are you creating `.sh` files in nexus?

macros/NextFlex_fullKr.init.mac
```diff
@@ -22,7 +22,7 @@
 ### ACTIONS
 /nexus/RegisterRunAction      DefaultRunAction
 /nexus/RegisterEventAction    DefaultEventAction
-/nexus/RegisterTrackingAction DefaultTrackingAction
+/nexus/RegisterTrackingAction OpticalTrackingAction
```
paolafer: Why are you changing this example to save all the optical photons? Take into account that this macro is run every time we run the automatic tests, and saving the optical photons increases the execution time.

paolafer:
As a generic comment: try to include in one commit only changes logically related among them and write a comment that describes these changes. For instance, in commit [174ccc3](https://github.com/next-exp/nexus/pull/178/commits/174ccc319d36c369cac76e59f8bead8bfdbc32bf) you create a new class, change the value of the reflectivity of a material and change the value of a parameter in a configuration file, but the the comment of the commit only refers to a bug in the event action (I guess it is related to the latter of the changes).

.vscode/c_cpp_properties.json
FabianKellerer: Yes, please let me know how to fix this.

.gitignore
FabianKellerer: I have a local sh script that compiles nexus for me. I added this so it wouldn't get uploaded to git.

macros/NextFlex_fullKr.init.mac
FabianKellerer: This was to check if the optical fiber barrel works as intended (For this of course the optical photons are of interest). Considering the automatic tests, this can be changed back.

.gitignore
paolafer: We don't want to add to `.gitignore` any files that the users may create in their repositories. If you don't want to add this script to git, you just don't add it to the tracking and that's it.

paolafer:
I also think that commits 60690bf4ed6b4ac05d41e3856c746bca04518069 and 0bb89f381c2cb631f22cd9d0b9dbe3a17af87337 can be removed altogether, since the latter removes lines introduced in the former.

.gitignore
gonzaponte: If I may add... you can add stuff to your `.gitignore` locally so git stops tracking specific files, just don't commit the changes in the `.gitignore` file.

paolafer:
Another general comment: if the comment of your commit exceeds the maximum number of characters, choose a shorter sentence in the first line and add more information in the following lines. I can show you how to do it, too.

macros/NextFlex_fullKr.init.mac
paolafer: Fabian, we think that the tests are failing because of the large RAM consumption due to this change. Please, change it back to `DefaultTrackingAction` and push the commit, to see if that's the case.

paolafer:
@FabianKellerer I guess that this PR can be deleted, right?

FabianKellerer:
> @FabianKellerer I guess that this PR can be deleted, right?

Yes! This was an earlier version of the GenericWLSfiber PR which is now merged, so we don't need this one anymore.

PR #181, opened 2022-10-21
Fix hit ids

paolafer:
When a particle crossed two different `ionization` sensitive detectors and deposited energy in both of them, the hit IDs of the hits were repeated, starting from 0 in each sensitive detector.
This PR fixes this bug, and now the hits of a given particle have increasing  hit ID, no matter which sensitive detector has registered them.


gondiaz (approved):
This PR makes `hit_id` to be unique in the hit list even after a change in the volume. Approved!

PR #183, opened 2022-10-24
Generic wls fiber

FabianKellerer:
Added the GenericWLSFiber geometry. Updated the MaterialsList to include the EJ-286 fiber core material.


source/geometries/GenericWLSFiber.cc
```diff
@@ -0,0 +1,299 @@
+// -----------------------------------------------------------------------------
+//  nexus | GenericWLSFiber.cc
+//
+//  Geometry of a configurable wave-length shifting optical fiber.
```
jmalbos: wave-length → wavelength

source/geometries/GenericWLSFiber.h
```diff
@@ -0,0 +1,87 @@
+// -----------------------------------------------------------------------------
+//  nexus | GenericWLSFiber.h
+//
+//  Geometry of a configurable wave-length shifting optical fiber.
```
jmalbos: wave-length → wavelength

source/geometries/GenericWLSFiber.h
```diff
@@ -0,0 +1,87 @@
...
+class G4GenericMessenger;
+class G4MaterialPropertiesTable;
+
+namespace nexus {
+
+  class GenericWLSFiber: public GeometryBase
+  {
+
+  public:
+    // Constructor for a generic wave-length shifting optical fiber
```
jmalbos: wave-length → wavelength

source/geometries/GenericWLSFiber.h
```diff
@@ -0,0 +1,87 @@
...
+    GenericWLSFiber(G4String    name,
+                    G4String    shape,            // "round"  or "square"
```
jmalbos: This should be an _enum_ or a _bool_ (there are only two options) instead of a _string_.

source/materials/MaterialsList.cc
```diff
@@ -818,6 +818,28 @@ namespace materials {
     return mat;
   }
 
+  // WLS EJ-286
```
jmalbos: Is there any difference between this definition and that of polyvinyl toluene (PVT)? What the manufacturer calls EJ286 is just PVT doped with a certain WLS. That matters for its optical properties, but not for the material definition. In the geometry I'd use PVT. This probably applies as well to EJ280, Y11, etc.

source/geometries/GenericWLSFiber.h
```diff
@@ -0,0 +1,87 @@
...
+                    G4double    thickness,        // diameter or side
+                    G4double    length,
+                    G4int       num_claddings,
```
jmalbos: Considering that, in practice, there are only two options (single or double cladding), I'd make this either an _enum_ or a _boolean_.

jmalbos (changes requested):
The geometry assumes now that the external coating will always be TPB, and that won't always be the case. Also, I don't see any optical material properties attached to the inner volumes of the fiber. Those should be incorporated into this class, giving the user the option to pass them via a function, as we do in the generic photosensor.

soleti:
Hello, I was wondering if there were any updates about this. If the code is fine it would be helpful for us working on fibers simulations to merge it.

paolafer:
@jmalbos is the reviewer, how is it going, Justo?

jmalbos:
> @jmalbos is the reviewer, how is it going, Justo?

I'll try to finish the review asap, but, most likely, I won't be able till next week.

paolafer:
Hi, everybody! How are we going with this PR? As far as I know, several people are using this geometry, without the code being approved, and I think it's not optimal. Could we move this forward? @jmalbos, @soleti?

source/geometries/GenericWLSFiber.cc
```diff
@@ -0,0 +1,298 @@
...
+                                 G4double    thickness,
+                                 G4double    length,
+                                 G4bool      doubleclad,
+                                 G4bool      with_coating,
+                                 G4Material* coating_mat,
+                                 G4Material* core_mat,
+                                 G4bool      visibility):
+  GeometryBase    (),
+  name_           (name),
+  isround_        (isround),         // "round"  or "square"
```
soleti: comment is misleading here, I would say `true` if round `false` if square

source/geometries/GenericWLSFiber.cc
```diff
@@ -0,0 +1,298 @@
...
+    else
+      new G4PVPlacement(nullptr, G4ThreeVector(0., 0., 0.), oclad_logic,
+                        oclad_name, innermost_logic, false, 0, false);
+
+    innermost_logic = oclad_logic;
+  }
+
+  // Inner Cladding (always built)
+  G4String iclad_name = name_ + "_ICLAD";
+  G4cout << "**** Building ICLAD " << iclad_name << G4endl;
```
soleti: Do we need maybe a verbosity check here (and in the following G4couts?)

source/geometries/NextFlexFieldCage.cc
```diff
@@ -330,11 +330,11 @@ void NextFlexFieldCage::DefineMaterials()
 
   // Fiber core material
   if (fiber_mat_name_ == "EJ280") {
-    fiber_mat_ = materials::EJ280();
+    fiber_mat_ = materials::PVT();
     fiber_mat_->SetMaterialPropertiesTable(opticalprops::EJ280());
   }
   else if (fiber_mat_name_ == "EJ286") {
-    fiber_mat_ = materials::EJ280();   // Same base material than EJ280
+    fiber_mat_ = materials::PVT();   // Same base material than EJ280
```
soleti, line 337: By looking at the code PVT is not exactly the same as the EJ280 material you are deleting. Can you explain why?

soleti (commented):
Apart few small things, I don't see any obvious problem in the code (which I have been using for a while after all...). However, before merging, I think it would be nice to verify that we reproduce some fundamental properties of the WLS fibers. At least two plots might be added for completeness to this PR (maybe they already exist but I couldn't find them on DocDB):
- attenuation length: light collected as a function of the distance from a generic photosensor
- trapping efficiency: amount of light collected by generating photons at the center of the fiber.

paolafer:
@FabianKellerer, how is this PR going?

source/geometries/GenericWLSFiber.cc
FabianKellerer: Fixed in the latest commit!

source/geometries/GenericWLSFiber.cc
FabianKellerer: Fixed in the latest commit!

FabianKellerer:
> Apart few small things, I don't see any obvious problem in the code (which I have been using for a while after all...). However, before merging, I think it would be nice to verify that we reproduce some fundamental properties of the WLS fibers. At least two plots might be added for completeness to this PR (maybe they already exist but I couldn't find them on DocDB):
> 
>     * attenuation length: light collected as a function of the distance from a generic photosensor
> 
>     * trapping efficiency: amount of light collected by generating photons at the center of the fiber.

This information can be found in the attached file. The trapping efficiency given by the manufacturer is reproduced very well, and the attenuation is aprrox. 60% per meter.
[Fibre Trapping Efficiencies.pdf](https://github.com/next-exp/nexus/files/12685966/Fibre.Trapping.Efficiencies.pdf)

soleti:
Thank you @FabianKellerer for the slides, very informative. The only remaining doubt I have is about the attenuation length. For e.g. Kuraray Y11 fibers that have an attenuation length of >3.5 m we should lose only ~25% of the light and not 60% after 1 m. Or am I misinterpreting your numbers? Thanks!

paolafer:
@FabianKellerer, could you have a look at this last question from Roberto, so that we can approve the PR? Thanks!

source/geometries/NextFlexFieldCage.cc
paolafer: In a previous comment, @jmalbos said that EJ280 is the same as PVT, except for the doping. However, this is not what our code says (the composition in terms of H and C is different). We need to be sure about what was wrong in the code, before eliminating EJ280 instead of PVT, for instance.

source/materials/MaterialsList.cc
paolafer: Also, Y11 and B2 are exactly the same material, so one of those should be eliminated. 
As for the others, there are differences in the density and in the composition (number of H and C atoms), therefore we should check which materials we are really using.

FabianKellerer:
> Thank you @FabianKellerer for the slides, very informative. The only remaining doubt I have is about the attenuation length. For e.g. Kuraray Y11 fibers that have an attenuation length of >3.5 m we should lose only ~25% of the light and not 60% after 1 m. Or am I misinterpreting your numbers? Thanks!

I made a plot comparing light loss in the simulation to the light loss measured by Kuraray [(Link)](http://kuraraypsf.jp/psf/ws.html) The agreement seems quite good.
AttenuationLength

soleti:
This looks good to me! I approve the PR.

paolafer:
The code is not compiling!

paolafer:
My compiler is also issuing the following compilation warning:
`source/geometries/GenericWLSFiber.cc:48:3: warning: field 'coating_optProp_' will be initialized after field 'core_mat_' [-Wreorder-ctor]
  coating_optProp_(nullptr)`. 

To avoid this kind of warnings, it is a good rule to declare variables in the header file and to initialize them in the `.cc` file in the same order. Can you fix that, too?

soleti (approved):

PR #184, opened 2022-10-26
Add B2 WLS fiber to the optical materials

soleti:
This PR adds the Kuraray B2 WLS fiber (https://www.kuraray.com/products/psf, from UV to blue) to the optical materials.
Kuraray doesn't provide the WLS absorption length as a function of the wavelength, it provides only an absorption spectrum. To convert it into an absorption length, we assume that the absorption length at the peak is the same as for the Y11 fiber and we scale accordingly. This procedure is an approximation but it was verified to work well with Y11, for which Kuraray did provide the WLS absorption length.

### DIFF
--- a/source/materials/MaterialsList.cc
+++ b/source/materials/MaterialsList.cc
@@ -842,6 +842,29 @@ namespace materials {
     return mat;
   }
 
+  // Kuraray B-2
+  G4Material* B2()
+  {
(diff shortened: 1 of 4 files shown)


paolafer (approved):
This PR adds a new WLS material with its properties, approved!

PR #185, opened 2022-10-26
First implementation of p-terphenyl wavelength shifter

soleti:
P-terphenyl is an organic compound which can be used to shift Xe light into the near UV. Coupled to UV-to-blue fibers (such as B2 from Kuraray), it may provide a higher light collection efficiency than the usual blue-to-green fibers + TPB.
This PR includes a first implementation of this material. The literature doesn't provide the WLS absorption length as a function of the wavelength, so we make an estimate based on the absorption spectrum provided in https://iopscience.iop.org/article/10.1088/1748-0221/5/04/P04007.

### DIFF
--- a/source/materials/MaterialsList.h
+++ b/source/materials/MaterialsList.h
@@ -81,9 +81,6 @@ namespace materials {
   // PS (Polystyrene)
   G4Material* PS();
 
-  // PTP (tetraphenyl butadiene)
-  G4Material* PTP();
-
   // TPB (tetraphenyl butadiene)
   G4Material* TPB();
 
--- a/source/materials/MaterialsList.h
+++ b/source/materials/MaterialsList.h
@@ -81,6 +81,9 @@ namespace materials {
   // PS (Polystyrene)
   G4Material* PS();
 
+  // PTP (tetraphenyl butadiene)
+  G4Material* PTP();
+
   // TPB (tetraphenyl butadiene)
   G4Material* TPB();
 
--- a/source/materials/OpticalMaterialProperties.cc
+++ b/source/materials/OpticalMaterialProperties.cc
@@ -736,6 +736,105 @@ namespace opticalprops {
   }
 
 
+  /// PTP (p-terphenyl) ///
+  G4MaterialPropertiesTable* PTP()
+  {
+    // Data from https://doi.org/10.1016/j.nima.2011.12.036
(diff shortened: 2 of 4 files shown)


source/materials/MaterialsList.h
```diff
@@ -81,6 +81,9 @@ namespace materials {
   // PS (Polystyrene)
   G4Material* PS();
 
+  // PTP (tetraphenyl butadiene)
```
paolafer: Typo: tetraphenyl butadiene --> p-terphenyl?

source/materials/OpticalMaterialProperties.cc
```diff
@@ -736,6 +736,105 @@ namespace opticalprops {
...
+
+
+
+    // WLS ABSORPTION LENGTH
+    // There are no tabulated values in the literature for the PTP absorption
+    // length as a function of the wavelength. 
+    // However in https://iopscience.iop.org/article/10.1088/1748-0221/5/04/P04007/
+    // it says that "the thickness >=150 nm of a polycrystalline p-terphenyl layer 
+    // is enough for absorption of >=99.9% of the xenon light". Thus, using the
+    // formula P = e^(-x/lambda) where lambda is the absorption length, we can
```
paolafer: To be clearer, I would say: `1-P = e^(-x/lambda), where P is the absorption probability`.

paolafer (changes requested):

soleti:
Now that I look, there was already a p-terphenyl material stub, but it was called TPH. Let me fix this...

paolafer:
Ok, I understand then that TPH is a specific type of p-terphenyl?

soleti:
Not that I am aware of? The materials list states `G4String name = "TPH"; // p - terphenyl`

paolafer:
> Not that I am aware how? The materials list states `G4String name = "TPH"; // p - terphenyl`

Ok, it's just an abbreviation of `terphenyl`, then :-).

jmalbos:
Some publications refer to p-terphenyl as TPH. The function @soleti mentions comes from the early days of NEXT, where we explored the use of several WLS. You can remove it or call the new function from within, if we want to keep both names.

paolafer:
I think it's better to keep just one name, as it is in the current version proposed by @soleti, it's simpler and avoids confusion.

paolafer (approved):
This PR updates the optical properties of p-terphenyl, approved!

PR #186, opened 2022-10-26
Muon energy fix

kvjmistry:
This PR fixes a binning issue with the energy of the simulated muon distributions. Previously the specified binning in the file did not match up with the actual bins leaving an unphysical gap at ~800 GeV. Additional bins in energy have been added.

The zenith range has also been expanded to include values from 0 to pi/2 (before a cut was applied for cos(zenith) > 0.4). 

The plots below show the angular and energy spectrum of the generated muons (~800k events). I only generated up to 3000 GeV (forgot to change the range!), however, the spectrum does go up to 5500 GeV in the updated file. 
image

Also added an extra catch in the code when generating from the angle+energy mode to restrict to only energies in the file (200 - 5500 GeV) otherwise there will be no available bin to sample from. The user must set an energy range if using this zenith-azimuth-energy ("zae") option.


source/generators/MuonAngleGenerator.cc
```diff
@@ -149,8 +150,26 @@ void MuonAngleGenerator::LoadMuonDistribution()
...
+      
+      if (s_header == "energy"){
+        EnergyRange.push_back(stod(s_energy));
+      }
+    
+    }
   }
 
+  // Check if the specified energy range has been set to a suitable value
+  if (dist_name_ == "zae" && (energy_min_ < EnergyRange.front()*GeV || energy_max_ > EnergyRange.back()*GeV )){
```
gondiaz: Instead of using `.back` and `.front`, I would compute the minimum and maximum values, just in case at some point the values in the file were not arranged

gondiaz (commented):
Is this clause needed?
https://github.com/kvjmistry/nexus/blob/f2b3c8f04f80671528e7831209976a0a2db31891/source/generators/MuonAngleGenerator.cc#L303

source/generators/MuonAngleGenerator.cc
kvjmistry: Fixed latest commit:
- Get max/min energies from file.
- remove the redundant clause.

source/generators/MuonAngleGenerator.cc
```diff
@@ -149,8 +154,34 @@ void MuonAngleGenerator::LoadMuonDistribution()
...
+
+      }
+    
+    }
   }
 
+  // Check if the specified energy range has been set to a suitable value
+  if (dist_name_ == "zae" && (energy_min_ < file_emin*GeV || energy_max_ > file_emax*GeV )){
+    std::cout << "The minimum energy allowed is: " << EnergyRange.front()*GeV/1000 << " GeV"<< std::endl;
+    std::cout << "The maximum energy allowed is: " << EnergyRange.back()*GeV/1000 << " GeV" << std::endl;
```
gondiaz: `.front` and `.back` still there

gondiaz (commented):
I would also add a `continue` here
https://github.com/kvjmistry/nexus/blob/bc621ffaae9b06b57964fda204a6562bd0e7c0f5/source/generators/MuonAngleGenerator.cc#L318

source/generators/MuonAngleGenerator.cc
kvjmistry: done!
also added the continue

gondiaz (approved):
This PR updates the muon flux files and avoids setting an energy range above the provided file limits. Good work!

PR #187, opened 2022-10-31
Add SS hexagonal mesh for NEXT100

kvjmistry:
This PR adds the option of stainless steel hexagonal meshes to the NEXT100 geometry. The SS hexagonal mesh is set by default, however, we can switch to the fake grid by setting the `useDielectricGrid_` bool.

Below is a picture of the cathode mesh (for the drawing, I blow up the hex size by a factor of 10 so we can draw them without causing the drawer to hang. By default, the visibility of drawing the meshes is turned off. 

image

This PR needs to be benchmarked for memory and speed in NEXT100 before merging so I mark it as a draft until this is complete.

### DIFF
--- a/source/geometries/Next100EnergyPlane.cc
+++ b/source/geometries/Next100EnergyPlane.cc
@@ -213,6 +213,16 @@ namespace nexus {
     G4LogicalVolume* copper_plate_logic =
       new G4LogicalVolume(copper_plate_solid, copper, "EP_COPPER_PLATE");
 
+    // Add optical surface
+    G4OpticalSurface* gas_copperplate_opsur = new G4OpticalSurface("GAS_COPPER_PLATE_OPSURF");
+    gas_copperplate_opsur->SetType(dielectric_metal);
+    gas_copperplate_opsur->SetModel(unified);
+    gas_copperplate_opsur->SetFinish(ground);
+    // gas_mesh_opsur->SetSigmaAlpha(0.0);
+    gas_copperplate_opsur->SetMaterialPropertiesTable(opticalprops::Copper());
+    new G4LogicalSkinSurface("GAS_COPPER_PLATE_OPSURF",
+                            copper_plate_logic, gas_copperplate_opsur);
+
     G4double stand_out_length =
       sapphire_window_thickn_ + tpb_thickn_ + optical_pad_thickn_ + pmt_stand_out_;
 
--- a/source/geometries/Next100EnergyPlane.cc
+++ b/source/geometries/Next100EnergyPlane.cc
@@ -213,7 +213,6 @@ namespace nexus {
     G4LogicalVolume* copper_plate_logic =
       new G4LogicalVolume(copper_plate_solid, copper, "EP_COPPER_PLATE");
 
-
     G4double stand_out_length =
       sapphire_window_thickn_ + tpb_thickn_ + optical_pad_thickn_ + pmt_stand_out_;
 
--- a/source/geometries/Next100EnergyPlane.cc
+++ b/source/geometries/Next100EnergyPlane.cc
@@ -213,15 +213,6 @@ namespace nexus {
     G4LogicalVolume* copper_plate_logic =
       new G4LogicalVolume(copper_plate_solid, copper, "EP_COPPER_PLATE");
 
-    // Add optical surface
-    G4OpticalSurface* gas_copperplate_opsur = new G4OpticalSurface("GAS_COPPER_PLATE_OPSURF");
-    gas_copperplate_opsur->SetType(dielectric_metal);
-    gas_copperplate_opsur->SetModel(unified);
-    gas_copperplate_opsur->SetFinish(ground);
-    // gas_mesh_opsur->SetSigmaAlpha(0.0);
-    gas_copperplate_opsur->SetMaterialPropertiesTable(opticalprops::Copper());
-    new G4LogicalSkinSurface("GAS_COPPER_PLATE_OPSURF",
-                            copper_plate_logic, gas_copperplate_opsur);
 
     G4double stand_out_length =
       sapphire_window_thickn_ + tpb_thickn_ + optical_pad_thickn_ + pmt_stand_out_;
--- a/source/geometries/Next100FieldCage.cc
+++ b/source/geometries/Next100FieldCage.cc
@@ -53,7 +53,6 @@ Next100FieldCage::Next100FieldCage(G4double grid_thickn):
 
   cathode_int_diam_    (965. * mm),
   cathode_ext_diam_    (1020.* mm),
-
   cathode_thickn_      (13.5  * mm),
   grid_thickn_         (grid_thickn),
   cathode_mesh_diam_   (5.0*mm),
@@ -104,6 +103,7 @@ Next100FieldCage::Next100FieldCage(G4double grid_thickn):
   el_grid_transparency_  (.90),
   max_step_size_ (1. * mm),
   visibility_ (0),
+  grid_visibility_ (0),
   verbosity_(0),
   useDielectricGrid_(0),
   // EL gap generation disk parameters
@@ -124,8 +124,9 @@ Next100FieldCage::Next100FieldCage(G4double grid_thickn):
   msg_ = new G4GenericMessenger(this, "/Geometry/Next100/",
                                 "Control commands of geometry Next100.");
   msg_->DeclareProperty("field_cage_vis", visibility_, "Field Cage Visibility");
+  msg_->DeclareProperty("grid_vis", grid_visibility_, "Grid Visibility");
   msg_->DeclareProperty("field_cage_verbosity", verbosity_, "Field Cage Verbosity");
-  
+
   msg_->DeclareProperty("useDielectricGrid", useDielectricGrid_, "Switch on Fake Grids");
 
   G4GenericMessenger::Command& drift_transv_diff_cmd =
@@ -392,7 +393,8 @@ void Next100FieldCage::BuildCathode()
                                                                grid_thickn_));
 
   // Cathode Mesh
-  G4LogicalVolume* cathode_grid_logic;
+  G4LogicalVolume *cathode_grid_logic;
+  G4LogicalVolume *cathode_hex_logic;
 
   if (useDielectricGrid_){
 
@@ -412,29 +414,20 @@ void Next100FieldCage::BuildCathode()
   else {
       
       // Dist from centre of hex to hex vertex, excluding the land width (circumradius)
-      G4double hex_circumR = cathode_mesh_diam_/std::sqrt(3)*mm;  
+      G4double hex_circumR = cathode_mesh_diam_/std::sqrt(3);  
 
-      // Calculate the number of hexagons needed
-      G4int nHole = (G4int) (std::sqrt(2)*(cathode_int_diam_/2.0) / (cathode_mesh_diam_)*mm);
+      // Total number of hexagons that would fit side-by-side along the diameter
+      G4int nHexDiam = (G4int) ((cathode_int_diam_/2.0) / (hex_circumR)*mm);
     
-      // Define the Stainless steel mesh cylinder to subtract hex pattern from
-      G4Tubs* diel_grid_solid =
-           new G4Tubs("CATHODE_GRID", 0., cathode_int_diam_/2.0 , grid_thickn_/2., 0., twopi);
+      // Define the disk to punch hexagon holes through for the mesh
+      G4Tubs* grid_solid = new G4Tubs("CATHODE_GRID", 0., cathode_int_diam_/2.0 , grid_thickn_/2., 0., twopi);
+      cathode_grid_logic = new G4LogicalVolume(grid_solid, steel_, "Cathode_Mesh_Logic");
 
       // Define a hexagonal prism
       G4ExtrudedSolid* HexPrism = HexCreator->CreateHexagon(grid_thickn_, hex_circumR);
+      cathode_hex_logic  = new G4LogicalVolume(HexPrism, gas_, "Mesh_Hex_Gas");
 
-      // Create the hexagonal mesh
-      G4MultiUnion* MeshUnion = new G4MultiUnion("UnitedMeshPattern");
-      
-      // Crate the Mesh template
-      HexCreator->CreateHexUnion(nHole, cathode_mesh_diam_, grid_thickn_, MeshUnion, HexPrism);
-    
-      // Subtract the template from the mesh cylinder
-      G4SubtractionSolid *meshS = new G4SubtractionSolid("Mesh", diel_grid_solid, MeshUnion);  
-
-      // Create the mesh logical volume
-      cathode_grid_logic = new G4LogicalVolume(meshS, steel_, "CATHODE_GRID");
+      HexCreator->PlaceHexagons(nHexDiam, cathode_mesh_diam_, grid_thickn_, cathode_grid_logic, cathode_hex_logic, cathode_int_diam_);
 
       new G4PVPlacement(0, G4ThreeVector(0., 0., cathode_zpos_), cathode_grid_logic,
                         "CATHODE_GRID", mother_logic_, false, 0, false);
@@ -465,7 +458,11 @@ void Next100FieldCage::BuildCathode()
     cathode_col.SetForceSolid(true);
     cathode_grid_logic->SetVisAttributes(G4VisAttributes::GetInvisible());
     cathode_logic->SetVisAttributes(cathode_col);
+
   }
+
+  if (!grid_visibility_ && !useDielectricGrid_)
+    cathode_hex_logic->SetVisAttributes(G4VisAttributes::GetInvisible());
 
 
   /// Verbosity
@@ -603,7 +600,8 @@ void Next100FieldCage::BuildELRegion()
     el_region->AddRootLogicalVolume(el_gap_logic);
   }
 
-  G4LogicalVolume* el_grid_logic;
+  G4LogicalVolume *el_grid_logic;
+  G4LogicalVolume *el_hex_logic;
 
   // EL Grid -- use Fake Dielectric
   if (useDielectricGrid_){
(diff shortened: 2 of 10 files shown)


paolafer:
Do we want to include this code in the next productions of NEXT-100? Has it been benchmarked?

kvjmistry:
@paolafer 

I have made some updates since this PR that fixes a memory leak and so it renders properly in the interactive display. I will make these changes to this PR and perform some tests as soon as I can.

kvjmistry:
I have rebased to the latest commit, the branch is ready for review. 
Some notes:
- The fake grid can be turned on/off with the macro line `/Geometry/Next100/useDielectricGrid 1`. By default, we use the SS mesh
- The visibility of the mesh is turned off by default and it can be switched on via `/Geometry/Next100/grid_vis 1`
- The reflectivity of SS and Copper was added, and I used a reflectivity of 20%. New measurements may be needed for this number. 

I did not notice much slow down in running with the SS meshes, but it is worth a check on the speed and memory use.

paolafer:
A test is failing, because one of the example macros simulates photons in (0, 0, 0), which now belongs to a volume made of stainless steel, which doesn't have optical properties. Can you simply change the specific vertex to one inside the EL region, for instance? The macro which gives the error is `NEXT100_S2_table.config.mac`.

source/geometries/Next100FieldCage.cc
```diff
@@ -274,7 +283,7 @@ void Next100FieldCage::DefineMaterials()
   steel_ = materials::Steel316Ti();
   // In Geant4 11.0.0, a bug in treating the OpBoundaryProcess produced in the surface makes the code fail.
   // This is avoided by setting an empty G4MaterialPropertiesTable of the G4Material.
-  steel_->SetMaterialPropertiesTable(new G4MaterialPropertiesTable());
+  steel_->SetMaterialPropertiesTable(opticalprops::Steel());
```
paolafer: I'm not sure this is correct. To simulate a reflection on a material, you have to define an optical surface wrapped around it with the desired properties, while here you give the properties to the material itself. Have you checked that reflections do actually happen?

source/geometries/Next100FieldCage.cc
```diff
@@ -375,39 +386,76 @@ void Next100FieldCage::BuildCathode()
...
 
+  }
+  // Use SS hexagonal mesh
+  else {
+      
+      // Dist from centre of hex to hex vertex, excluding the land width (circumradius)
+      G4double hex_circumR = cathode_mesh_diam_/std::sqrt(3);  
+
+      // Total number of hexagons that would fit side-by-side along the diameter
+      G4int nHexDiam = (G4int) ((cathode_int_diam_/2.0) / (hex_circumR)*mm);
```
paolafer: Why do you multiply by `mm`? Isn't `nHexDiam` just a number, with no units?

source/utils/HexagonMeshTools.cc
```diff
@@ -0,0 +1,99 @@
...
+        G4double phi = i*ang;
+        G4double cosphi = std::cos(phi);
+        G4double sinphi = std::sin(phi);
+        polygon[i].set(circumradius*cosphi, circumradius*sinphi);
+    }
+    
+    // Define the hexagon
+    G4TwoVector offsetA(0,0), offsetB(0,0);
+    G4double scaleA = 1, scaleB = 1;
+    return new G4ExtrudedSolid("Extruded", polygon, Thickness, offsetA, scaleA, offsetB, scaleB);
```
paolafer: According to the G4 documentation, the third argument of `G4ExtrudeSolid` is a half thickness. Wouldn't it be better to make it clear in this code? Also, you pass the grid thickness to this function, not half the thickness. Is the thickness arbitrary? Even in that case, I think that it is better to make the code clear and use `thickness/2`. In nexus variable usually have names with lower case letter.

source/geometries/Next100FieldCage.h
```diff
@@ -101,6 +106,9 @@ namespace nexus {
     CylinderPointSampler2020* anode_gen_;
     CylinderPointSampler2020* holder_gen_;
 
+    // Hexagonal Mesh Tool
+    HexagonMeshTools* HexCreator;
```
paolafer: In nexus global variables end with an underscore. Also, we use the convention of lower case letter and `_` for the name of the variables (it may be not completely consistent across the code, but we're trying to follow this rule for new code). 
You can have a look [here](https://github.com/next-exp/nexus/wiki/Coding-conventions-and-style-guide) for the code conventions we're using.

source/geometries/Next100FieldCage.h
```diff
@@ -78,9 +79,13 @@ namespace nexus {
 
     // Visibility of the geometry
     G4bool visibility_;
+    G4bool grid_visibility_;
     // Verbosity of the geometry
     G4bool verbosity_;
 
+    // Use fake mesh
+    G4bool useDielectricGrid_;
```
paolafer: Same as below.

source/utils/HexagonMeshTools.h
```diff
@@ -0,0 +1,42 @@
+// ----------------------------------------------------------------------------
```
paolafer, line 1: On the line of simplifying the code and not using classes when they are not needed, I think these could be two functions. That would be similar to what is done in `IOUtils`.

source/utils/HexagonMeshTools.cc
```diff
@@ -0,0 +1,99 @@
...
+
+  void HexagonMeshTools::PlaceHexagons(G4int nHole, G4double InRadius, G4double Thickness, G4LogicalVolume* DiskLogical, G4LogicalVolume* HexLogical, G4double Mesh_D){
+    
+    // Logical Volume of SS Disk
+
+    // Use cube coordinate system defined in:
+    // https://www.redblobgames.com/grids/hexagons/
+
+    // Dist from hexagon centre to center of hexagon vertex (inc. wire thickness)
+    G4double HexSize = (InRadius + Thickness)/std::sqrt(3.0)*mm;
```
paolafer: Why are you multiplying by `mm`? In principle both `InRadius` and `Thickness` have already their units included.

paolafer: HexSize --> hex_size and the rest of variable within the function.

source/utils/HexagonMeshTools.cc
```diff
@@ -0,0 +1,99 @@
...
+        G4double R = std::sqrt(x*x + y*y);
+        
+        // Only place hexagons within a circle the size of the mesh
+        if (R > Mesh_D/2.0)
+          continue;
+
+        
+        G4RotationMatrix rot = G4RotationMatrix(0, 0, 0);
+        G4Transform3D tr = G4Transform3D(rot, G4ThreeVector(x, y, 0*mm));
+        new G4PVPlacement(0, G4ThreeVector(x, y, 0*mm), HexLogical, "EL_MeshP", DiskLogical, false, 0, false);
```
paolafer: If I understand correctly, these volumes are gas volumes that are placed in the stainless steel volume of the grid. In this case, I would make it clear in the string passed to the placement, because it will be the string used to recognize the volume in the output information. You could use, for instance, `MESH_GAS` or `MESH_HOLE_GAS` or anything that conveys the message. Since it's the same volume for any grids (cathode or EL), I would avoid using the `EL` word. The name of the volumes used in the placements are usually in capital letters (with some exceptions, for instance, the `SiPM` string).

paolafer (changes requested):

paolafer:
Is the grid thickness measured? Right now we are using an arbitrary value, set in the `Next100` class, but it can be changed.

paolafer:
I noticed that the ionization electrons that hit the gate grid continue their path to the EL gap: is that the intended behaviour? Because of that, the lifetime value for this ionization electrons in the EL gap is assigned erroneously, because stainless steel does not have the `attachment` property. That property is read from the first material after the cathode, which is now stainless steel, because the cathode is defined at `z == 0`, exactly between the active volume and the grid volume.

source/geometries/Next100FieldCage.cc
kvjmistry: You are right, I don't know why I did it like this. I'll add a fix for this in my next commits and make sure it works with a test.

source/geometries/Next100FieldCage.cc
kvjmistry: eek, I got lucky that `*mm` multiplies by 1

source/utils/HexagonMeshTools.cc
kvjmistry: Changed variable name `Thickness` -> `half_thickness` and divide mesh thickness by two when it is passed into the function.

source/geometries/Next100FieldCage.h
kvjmistry: Thanks for the reminder, I have updated the variable names where I see them.

source/utils/HexagonMeshTools.cc
kvjmistry: Again, I got lucky here, and `*mm` multiplies by 1. I have removed it.

source/utils/HexagonMeshTools.cc
kvjmistry: fixed

kvjmistry:
> Is the grid thickness measured? Right now we are using an arbitrary value, set in the `Next100` class, but it can be changed.

The grid thickness is 0.13 microns. I have updated the code.

kvjmistry:
> I noticed that the ionization electrons that hit the gate grid continue their path to the EL gap: is that the intended behaviour? Because of that, the lifetime value for this ionization electrons in the EL gap is assigned erroneously, because stainless steel does not have the `attachment` property. That property is read from the first material after the cathode, which is now stainless steel, because the cathode is defined at `z == 0`, exactly between the active volume and the grid volume.

This is a good question. From my simulations, almost all the electrons should reach the EL region. See e.g. https://arxiv.org/abs/2311.03528 Fig 13 where the electrons follow the electric field lines and get funneled into the center of the hexagons.

source/utils/HexagonMeshTools.h
kvjmistry: Done

paolafer:
> > Is the grid thickness measured? Right now we are using an arbitrary value, set in the `Next100` class, but it can be changed.
> 
> The grid thickness is 0.13 microns. I have updated the code.

Do you mean 0.13 millimetres, right?

paolafer:
> > I noticed that the ionization electrons that hit the gate grid continue their path to the EL gap: is that the intended behaviour? Because of that, the lifetime value for this ionization electrons in the EL gap is assigned erroneously, because stainless steel does not have the `attachment` property. That property is read from the first material after the cathode, which is now stainless steel, because the cathode is defined at `z == 0`, exactly between the active volume and the grid volume.
> 
> This is a good question. From my simulations, almost all the electrons should reach the EL region. See e.g. https://arxiv.org/abs/2311.03528 Fig 13 where the electrons follow the electric field lines and get funneled into the center of the hexagons.

In that case, we should give the `attachment` optical property to the stainless steel. One way to do that would be to create a special item of the OpticalMaterialProperty functions with only that property and attach it to the material. I think that  a consequence of that is that any volume made of stainless steel would have that property (not 100% sure, but I remember that we had an issue like this in the past and we had to duplicate the definition of a material). However, I think that there's no harm in having an attachment value in the vessel volume, for instance, since no ionization electrons end up there and, even if they did, they would not create scintillation light.

paolafer:
Running `/geometry/test/run` in the interactive mode several overlap warnings are raised, have you checked that? It doesn't happen with all the holes, only a few of them.

kvjmistry:
> > > Is the grid thickness measured? Right now we are using an arbitrary value, set in the `Next100` class, but it can be changed.
> > 
> > 
> > The grid thickness is 0.13 microns. I have updated the code.
> 
> Do you mean 0.13 millimetres, right?

Yes!

source/geometries/Next100FieldCage.h
```diff
@@ -47,12 +47,12 @@ namespace nexus {
...
     // Dimensions
-    G4double gate_sapphire_wdw_dist_;
+    G4double gate_sapphire_wdw_dist_;    
     const G4double active_diam_, n_panels_, active_ext_radius_;
-    const G4double cathode_int_diam_, cathode_ext_diam_, cathode_thickn_, grid_thickn_;
+    const G4double cathode_int_diam_, cathode_ext_diam_, cathode_thickn_, grid_thickn_, cathode_mesh_diam_;
     const G4double teflon_long_length_, teflon_buffer_length_;
     const G4double teflon_thickn_;
-    const G4double el_gap_length_;
+    const G4double el_gap_length_, el_mesh_diam_, el_mesh_rot_;;
```
paolafer: Remove one semicolon.

paolafer:
Just a detail: my compiler gives a warning about the different order in which `grid_thickn_` and `cathode_mesh_diam_` are declared and initialized. Could you just swap them in either place?

source/geometries/Next100FieldCage.cc
```diff
@@ -425,7 +425,7 @@ void Next100FieldCage::BuildCathode()
 
       PlaceHexagons(n_hex, cathode_mesh_diam_, grid_thickn_, cathode_grid_logic, cathode_hex_logic, cathode_int_diam_);
 
-      new G4PVPlacement(0, G4ThreeVector(0., 0., cathode_zpos_), cathode_grid_logic,
+      new G4PVPlacement(0, G4ThreeVector(0., 0., cathode_zpos_ - cathode_thickn_/2. + grid_thickn_/2.), cathode_grid_logic,
```
paolafer: We could use the same variable for the position of the grid, in both cases (dielectric and SS grid), to avoid possible errors, should the position be modified in the future.
Also, `x` and `y` should also be the same, right?

source/materials/OpticalMaterialProperties.cc
```diff
@@ -1666,33 +1656,21 @@ namespace opticalprops {
...
-      mpt->AddProperty("SPECULARLOBECONSTANT", ENERGIES_2, specularlobe);
-      mpt->AddProperty("SPECULARSPIKECONSTANT",ENERGIES_2, specularspike);
-      mpt->AddProperty("BACKSCATTERCONSTANT",  ENERGIES_2, backscatter);
-      mpt->AddProperty("REFLECTIVITY", ENERGIES, REFLECTIVITY);
+      // Reflectivity
+      std::vector<G4double> refl_energies = {
+              optPhotMinE_, hc_ / (500. * nm), hc_ / (350. * nm),
+                            hc_ / (300. * nm), hc_ / (170. * nm),  optPhotMaxE_ };
+
+      // We assume a reflectivity of the stainless steel of 20% at VUV
```
paolafer: Regarding the reflectivities of steel and copper, so far we have assumed zero all over the place. I don't have a feeling of how much this modification would change the total collected light, but maybe we could make a small production of Kr to compare and be aware of the implications.
I would like to hear @jmalbos (and maybe @jmunozv) opinions here.

paolafer: It could also be a different PR from this one, at least regarding copper's reflectivity, which is not directly related to the grids. Also, you've added a reflective surface to the copper of the EP, but why not to the TP copper plate, also?

kvjmistry: Yes I can move the reflectivities of copper to a different one if that is preferred. As for why only the EP, the TP should be fully covered with the DICE boards whereas there is some exposed copper surface on the EP.

source/geometries/Next100FieldCage.cc
kvjmistry: I've added the variable declaration for the positon so its common between the two. I believe the x and y should be zero still. I have used the GetCoordOrigin().x() and GetCoordOrigin().y() to be the same as the fake grid placement.

source/geometries/Next100FieldCage.cc
```diff
@@ -692,7 +697,20 @@ void Next100FieldCage::BuildELRegion()
 
     new G4PVPlacement(0, G4ThreeVector(0., 0., anode_sim_zpos - grid_thickn_/2.),
                       anode_logic, "ANODE_RING", mother_logic_, false, 0, false);
+
+    /// EL gap.
+    el_gap_solid =
+      new G4Tubs("EL_GAP", 0., gate_ext_diam_/2.,
```
paolafer, line 703: Why is the diameter of the el gap changed from `gate_int_diam_` to `gate_ext_diam_` in the case of the SS grids?

source/geometries/Next100FieldCage.cc
```diff
@@ -403,15 +465,22 @@ void Next100FieldCage::BuildCathode()
     G4VisAttributes grey = nexus::LightGrey();
```
paolafer: Reviewing this piece code, I think that we can simplify it either by taking

```
G4VisAttributes cathode_col = nexus::DarkGrey();
cathode_col.SetForceSolid(true);
cathode_logic->SetVisAttributes(cathode_col);
```

out of the if/else clause, since it is always executed, or decide to visualize the cathode ring only when the visibility is turned on for this class.

source/geometries/Next100FieldCage.cc
```diff
@@ -496,42 +565,159 @@ void Next100FieldCage::BuildBuffer()
...
-  G4Tubs* anode_solid =
+  G4Tubs* anode_solid;
+  G4LogicalVolume* anode_logic;
+  G4double anode_sim_zpos = el_gap_zpos_ - el_gap_length_/2. - anode_ring_thickn/2.;
+
+  /// EL gap.
+  G4Tubs* el_gap_solid;
+  G4LogicalVolume* el_gap_logic;
+
+  // Logical Volumes
```
paolafer: Here you could mimic the other comments with `EL grids` instead of a generic `Logical Volumes`.

source/geometries/Next100FieldCage.cc
```diff
@@ -496,42 +565,159 @@ void Next100FieldCage::BuildBuffer()
...
+    el_gap_logic =
+      new G4LogicalVolume(el_gap_solid, gas_, "EL_GAP");
+
+    new G4PVPlacement(0, G4ThreeVector(0., 0., el_gap_zpos_),
+                      el_gap_logic, "EL_GAP", mother_logic_, false, 0, false);
+
+    /// EL grids
+    G4Material* fgrid_mat = materials::FakeDielectric(gas_, "el_grid_mat");
+    fgrid_mat->SetMaterialPropertiesTable(opticalprops::FakeGrid(pressure_,
+                                                               temperature_,
```
paolafer, line 620: Align the rest of the arguments with `pressure_`.

source/geometries/Next100FieldCage.cc
```diff
@@ -496,42 +565,159 @@ void Next100FieldCage::BuildBuffer()
...
+
+    new G4PVPlacement(0, G4ThreeVector(0., 0., anode_sim_zpos - grid_thickn_/2.),
+                      anode_logic, "ANODE_RING", mother_logic_, false, 0, false);
+
+    /// EL gap.
+    el_gap_solid =
+      new G4Tubs("EL_GAP", 0., gate_ext_diam_/2.,
+                (el_gap_length_ + 2*grid_thickn_)/2., 0, twopi);
+
+    el_gap_logic =
```
paolafer: The definition of the logical volume and the placement are the same for both cases (dielectric or SS grids), right? In that case, we could bring these lines outside the loop.

source/geometries/Next100FieldCage.cc
```diff
@@ -51,17 +52,21 @@ Next100FieldCage::Next100FieldCage(G4double grid_thickn):
...
+  cathode_mesh_diam_   (5.0*mm),
 
   //  teflon_drift_length_ (1178.*mm), //distance from the gate to the beginning of the cathode volume
   teflon_long_length_   (465. * mm),
   teflon_buffer_length_ (241. * mm),
   teflon_thickn_       (5. * mm),
 
   el_gap_length_ (10. * mm),
+  el_mesh_diam_  (2.5 * mm),
+  el_mesh_rot_   (15 * deg),
```
paolafer, line 69: Are the meshes installed at LSC rotated by exactly this angle? Was it measured? Or is this angle fixed by construction?

source/geometries/Next100FieldCage.cc
```diff
@@ -603,11 +769,15 @@ void Next100FieldCage::BuildELRegion()
   if (visibility_) {
     G4VisAttributes light_blue = nexus::LightBlue();
     el_gap_logic->SetVisAttributes(light_blue);
-    diel_grid_logic->SetVisAttributes(G4VisAttributes::GetInvisible());
+    el_grid_logic->SetVisAttributes(G4VisAttributes::GetInvisible());
```
paolafer: If the EL grid is always invisible, you can move it outside the if/else clause, together with the attributes of gate and anode.

source/geometries/Next100EnergyPlane.cc
```diff
@@ -213,6 +213,7 @@ namespace nexus {
     G4LogicalVolume* copper_plate_logic =
       new G4LogicalVolume(copper_plate_solid, copper, "EP_COPPER_PLATE");
 
+
```
paolafer: Is the addition of this line desired? If it's a mistake, I would revert it; this way, we don't see a modification in a file that is not affected by this PR.

source/materials/OpticalMaterialProperties.h
```diff
@@ -83,6 +83,10 @@ namespace opticalprops {
 
   G4MaterialPropertiesTable* PMMA();
 
+  G4MaterialPropertiesTable* Copper();
```
paolafer: Since you left out the steel and copper optical properties from this PR, I would remove also these remaining modifications to the `OpticalMaterialProperties` files.

paolafer (changes requested):

paolafer (approved):
Great job simulating our grids with high precision! Approved!

PR #190, opened 2022-11-18
Emission and absorption fixes for B2 and TPH

soleti:
The B2 and TPH materials I added few weeks ago (PR #184) had the arrays in the inverted order for emission and absorption.


paolafer (approved):
Good catch! Approved.

PR #191, opened 2022-11-19
Fix comma in OpticalMaterialProperties...

soleti:
Was missing a comma...
The code was compiling but crashing at execution time, funny to find the source...


paolafer (approved):
The code compiles and runs. Approved.

PR #192, opened 2023-02-09
Add shielding class for NEW

paolafer:
Since tag `v6_03_00`, the geometry of the lead castle at LSC (the `Next100Shielding` class) has changed quite a lot, after a thorough review and it has also been adapted to the NEXT-100 geometry. As a consequence, it cannot be used as it is in NEXT-White. 
In this PR we add a new class for shielding, to be used only with NEXT-White, importing the `Next100Shielding` class from tag `v6_02_00` and adapting it to the current nexus code.
Of course, this means that the latest additions to the castle geometry are not present when using NEW.


gonzaponte:
For reference, these plots show a proxy[^1] the position of the SiPMs in the current master and with the changes implemented by this PR.

image
image

[^1]: The plot is generated by tracking optical photons that go into the volume `"SiPMSensl"`

gonzaponte (approved):
The shift in the y-axis has been fixed. Approved!

PR #193, opened 2023-02-14
PolishedAl

mariandbt:
Added Polished Aluminium (PolishedAl) to materials

### DIFF
--- a/source/materials/OpticalMaterialProperties.cc
+++ b/source/materials/OpticalMaterialProperties.cc
@@ -735,6 +735,76 @@ namespace opticalprops {
     return mpt;
   }
 
+  /// PolishedAl ///
+  G4MaterialPropertiesTable* PolishedAl()
+  {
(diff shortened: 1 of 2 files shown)


paolafer:
Hi @mariandbt! Welcome to the nexus developers :-)!
I'll comment you a few formal things about commits in nexus, which can be useful also for other newcomers.

- We have the convention of using the imperative mood in the comment of the commit; in your case, it could be, for instance, "Add PolishedAl material". You can have a look [here](https://cbea.ms/git-commit/) at some good practices in writing git commits.
- If you realize you forgot to add something to a commit, you can amend the commit, instead of making a new one with the same comment. In the present case, since you already added a new commit, you can squash the two commits. If you are not familiar with these git actions we can talk offline.
- You can see that the branch has conflicts with the `master` branch. You should try to make your changes always on the latest version of the master, by fetching `master` and rebasing your code to it from time to time. Try to do it now and solve the conflicts.

PR #194, opened 2023-02-16
Possibility of not specifying the persistency manager

paolafer:
This PR removes the need of specifying a persistency manager to run a simulation.
It can be useful for debugging, an example of which has been added in a pair of macros.
Additionally, the names of two configuration parameters of the `PersistencyManager` class have been changed to follow the current convention.


gonzaponte:
@amirbh61 also needs this. Can I help push it forward?

soleti:
@gonzaponte thanks for the reminder, I'll review this by Monday, but feel free to add yourself as a second reviewer if you want.

macros/NEXT100_debug.config.mac
```diff
@@ -0,0 +1,25 @@
...
+
+##### GEOMETRY #####
+/Geometry/Next100/specific_vertex 5 5 0 mm
+
+##### GENERATOR #####
+/Generator/SingleParticle/particle geantino
+/Generator/SingleParticle/min_energy 1 eV
+/Generator/SingleParticle/max_energy 1 eV
+/Generator/SingleParticle/region AD_HOC
+/Generator/SingleParticle/momentum 0 0 1
```
gonzaponte: add a newline at the end of file

gonzaponte: Did you try to shoot something that produced hits? Before I saw this PR existed I started developing something similar in parallel and had a segfault because it attempted to store trajectories.

Is it meant to be used only for non-interacting events or is it more general?
If it is the latter, I suggest shooting an electron in the example macro to cover all cases. It only takes a few milliseconds to run anyway.

macros/nexus_example1.config.mac
```diff
@@ -19,4 +19,4 @@
 
 /control/execute macros/physics/IonizationElectron.mac
 
-/nexus/persistency/outputFile nexus.next
\ No newline at end of file
+/nexus/persistency/output_file nexus.next
```
gonzaponte: add newline at the end of file

source/base/NexusApp.h
```diff
@@ -55,6 +55,8 @@ namespace nexus {
     G4String trkact_name_; ///< Name of the chosen tracking action
     G4String stkact_name_; ///< Name of the chosen stacking action
 
+    G4bool pman_; ///< True if the persistency manager is set
+
```
gonzaponte, line 59: I don't know what's the convention in nexus, but usually in C++ you could simply check if `pm_` is a `nullptr`.

I like that there is a more explicit variable, though, so for me this is fine.

source/persistency/PersistencyManager.cc
```diff
@@ -42,15 +42,15 @@ REGISTER_CLASS(PersistencyManager, PersistencyManagerBase)
 
 
 PersistencyManager::PersistencyManager():
-  PersistencyManagerBase(), msg_(0), ready_(false),
+PersistencyManagerBase(), msg_(0), output_file_("nexus_out"), ready_(false),
```
gonzaponte, line 45: why does it crash if you don't specify a default value?

gonzaponte (commented):
Looks good to me. Just a couple of questions below.

source/base/NexusApp.h
paolafer: If you think this way it is easier to understand what the code does, then I would keep it.

source/base/NexusApp.cc
```diff
@@ -105,13 +105,13 @@ NexusApp::NexusApp(G4String init_macro): G4RunManager(), gen_name_(""),
...
 
-  if (pm_name_ == "") {
-    G4Exception("[NexusApp]", "NexusApp()", FatalException, "A persistency manager must be specified.");
-  }
-  pm_ = ObjFactory<PersistencyManagerBase>::Instance().CreateObject(pm_name_);
-  pm_->SetMacros(init_macro, macros_, delayed_);
 
- // PersistencyManager::Initialize(init_macro, macros_, delayed_);
+  // Set the persistency manager, if needed
+  if (pm_name_ != "") {
```
soleti: I think `!pm_name.empty()` is slightly faster than `!= ""` and the "recommended" way

soleti (commented):
Looks good to me (although I have limited experience with nexus)

source/persistency/PersistencyManager.cc
paolafer: In fact, it doesn't: it doesn't produce an output file at all.

paolafer:
> Looks good to me (although I have limited experience with nexus)

If you're happy with the PR, please, approve it (Review changes --> Approve, in the Files changed tab).

soleti (approved):

Issue #47, opened 2020-11-05
SetELzCoord and GetELzCoord functions from BaseGeometry


jmalbos:
These two functions are not general enough to be part of the base class `BaseGeometry`. On top of that, their use to position a physical volume within its mother results in somewhat obscure code, in my opinion. A simple alternative: providing the origin of coordinates via the constructor as done in the class `Next100TrackingPlane()`, or via a setter specific to the class that the user has to call necessarily.

paolafer:
Yes, I agree.

jmunozv:
Me too

Javier Munoz Vidal
javier.munoz.vidal@gmail.com

paolafer:
PR #196 addresses this issue with a proposal, to be discussed. Please, have a look at the code, whenever you have a chance.

paolafer:
This issue is being addressed in PR #222.

PR #196, opened 2023-02-17
Clean base geometry

paolafer:
This PR addresses issue #47 and eliminates the concept of `EL z coordinate` replacing it with the origin of coordinates, which is more general and can be used with any geometry, when it is not placed in (0, 0, 0).
I have implemented the changes only for NEXT-100, for the moment, to receive feedback. If we consider this approach OK, the next steps would be to remove the setter/getter of `ELzCoord` from `GeometryBase.h` and do the same with the NEW and DEMO geometries. I would probably implement the changes only in DEMO, since NEW is soon to become obsolete; we could simply transfer the `ELzCoord` setter/getter to each NEW class that needs them.
I would like to know @jmalbos and @jmunozv's opinions on this topic, plus anybody else's who wants to contribute.

### DIFF
--- a/source/geometries/GeometryBase.h
+++ b/source/geometries/GeometryBase.h
@@ -40,9 +40,6 @@ namespace nexus {
     /// Returns the 3 dimensions of the geometry (x, y, z)
     G4ThreeVector GetDimensions();
 
-    /// Returns true if geometry has a drift field
-    G4bool GetDrift() const;
-
     // Getter for the starting point of EL generation in z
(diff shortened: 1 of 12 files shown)


paolafer:
The commits of this PR have been added in PR #222. Since the NEXT-100 field cage is displaced from (0, 0, 0) in three coordinates, it is necessary to make these changes now. 
Since I haven't heard complaints since February, I assume that everybody agrees with the new implementation.

PR #197, opened 2023-02-17
Add 2nu ECEC generator

gonzaponte:
This PR is a resurrection of #26, which was closed due to repository configuration issues.

This PR adds a generator for 2-neutrino double electron capture.
Example macros have also been added for reference and testing.

### DIFF
--- a/macros/NEW_2nuECEC.config.mac
+++ b/macros/NEW_2nuECEC.config.mac
@@ -1,0 +1,27 @@
+## ----------------------------------------------------------------------------
+## nexus | NEW_2nuECEC.config.mac
+##
+##  Configuration macro to simulate 2-neutrino ECEC decays
+##  in the NEW detector geometry
+##
+## The NEXT Collaboration
+## ----------------------------------------------------------------------------
+
+/run/verbose 0
+/event/verbose 0
+/tracking/verbose 0
+
+/process/em/verbose 0
+
+/Generator/ECEC/region CENTER
+/Generator/ECEC/atomic_number 52
+/Generator/ECEC/shell K
+
+/Geometry/NextNew/pressure 10. bar
+/Geometry/NextNew/elfield false
+
+/control/execute macros/physics/IonizationElectron.mac
+
+/nexus/persistency/outputFile new_ecec.next
+
+/nexus/random_seed 542862
--- a/source/generators/ECECGenerator.cc
+++ b/source/generators/ECECGenerator.cc
@@ -99,8 +99,6 @@ void ECECGenerator::Initialize() {
   atom_ = new G4UAtomicDeexcitation();
   atom_->SetFluo (true);
   atom_->SetAuger(true);
-  atom_->SetCutForSecondaryPhotons(0);
-  atom_->SetCutForAugerElectrons  (0);
 
   shell_ = atom_->GetAtomicShell(atomic_number_, GetShellID(shell_name_));
 
--- a/source/generators/ECECGenerator.cc
+++ b/source/generators/ECECGenerator.cc
@@ -1,0 +1,157 @@
+// ----------------------------------------------------------------------------
+// nexus | ECECGenerator.cc
+//
+// This class is the primary generator for events consisting in the decay
+// of a double electron capture. The user must specify via configuration
+// parameters the atomic number, mass number and energy level of the
+// daughter isotope of interest.
+//
+// The NEXT Collaboration
+// ----------------------------------------------------------------------------
+
+
+#include "ECECGenerator.h"
+#include "GeometryBase.h"
+#include "DetectorConstruction.h"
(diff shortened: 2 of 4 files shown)


source/generators/ECECGenerator.cc
```diff
@@ -0,0 +1,158 @@
+// ----------------------------------------------------------------------------
+// nexus | ECECGenerator.h
```
paolafer: typo: `.h` --> `.cc`

gonzaponte:
For reference, these are the hit energy spectra obtained with the example macro.

image
image

macros/NEW_2nuECEC.config.mac
```diff
@@ -0,0 +1,23 @@
+### --------------------------------------------------------
+###  This init macro simulates bb0nu decays
```
paolafer: Update this brief description both in the `init` and in the `config` file.

macros/NEW_2nuECEC.config.mac
```diff
@@ -0,0 +1,23 @@
+### --------------------------------------------------------
+###  This init macro simulates 2-neutrino ECEC decays
```
paolafer: I know I'm a bit of a pain, but could you use the same format as the other example macros :-)?

gonzaponte: You can't beat me at that ;). Do you mean like the one I just pushed?

macros/NEW_2nuECEC.config.mac
```diff
@@ -0,0 +1,27 @@
+## ----------------------------------------------------------------------------
+## nexus | NEW_2nuECEC.init.mac
```
paolafer: Perfect! Just a typo: `init` --> `config` here.

gonzaponte: Classic copy-paste mistake :)

source/generators/ECECGenerator.cc
```diff
@@ -0,0 +1,158 @@
...
+
+ECECGenerator::~ECECGenerator()
+{
+  delete msg_;
+  delete shell_;
+  delete atom_;
+}
+
+
+G4AtomicShellEnumerator ECECGenerator::GetShellID(G4String s)
```
paolafer: I get a compiler warning because of the variable name: `warning: declaration shadows a variable in namespace 'CLHEP'`. Could you use another name instead of `s`? Anything that is not related to units of measure.

source/generators/ECECGenerator.cc
```diff
@@ -0,0 +1,158 @@
...
+
+REGISTER_CLASS(ECECGenerator, G4VPrimaryGenerator)
+
+ECECGenerator::ECECGenerator():
+  G4VPrimaryGenerator(),
+  atomic_number_(-1),
+  shell_name_("K"), // K, L1, L2, L3, M1, ..., M5
+  region_("ACTIVE"),
+  msg_ (nullptr),
+
```
paolafer: Remove blank line.

source/generators/ECECGenerator.cc
```diff
@@ -0,0 +1,157 @@
...
+  if (!detconst)
+    G4Exception("[ECECGenerator]", "ECECGenerator()", FatalException,
+                "Unable to load geometry.");
+
+  geom_ = detconst->GetGeometry();
+
+  atom_ = new G4UAtomicDeexcitation();
+  atom_->SetFluo (true);
+  atom_->SetAuger(true);
+  atom_->SetCutForSecondaryPhotons(0);
```
paolafer: Here the class is initialized with two class members (`minGammaEnergy` and `minElectronEnergy`) to zero. However, when you generate the particles in line 149, you comment that 0 can't be used and those variable are modified inside the function. Is there any reason why you should not set `0.1 eV` here?

gonzaponte: I think this was just the first thing I tried and then I realized that I needed to set it to >0 in the GenerateParticles function. At the time I didn't know that GenerateParticles overwrites these values. So probably these two lines can even be removed. What do you prefer?

paolafer: For what I can see in the code, those thresholds are used exclusively in the `GenerateParticles()` function, therefore it should be exactly the same; on top of that, that function requires to be passed those arguments. The whole structure looks quite confusing to me, but it's Geant4... I would remove the setters at the beginning, since they don't do anything.

gonzaponte: As discussed offline, these setters have no effect as the `GenerateParticle` method overrides them. They have been removed.

paolafer (approved):
This PR adds the code for the double electron capture generator, used in the simulations of the NEXT paper on ECEC. It was long due and it's finally approved!

PR #198, opened 2023-02-21
Using different files formats in the output histogram of the validation tracking action

paolafer:
This PR makes it possible to use different formats for the output file containing gamma energies, when using `ValidationTrackingAction`.
While csv format is the default one, ROOT can still be used specifying the .root extension at the end of the file name, in the configuration file.


kvjmistry:
These changes to allow a file name to be passed to the validation tracking action with any format desired is good.

I noticed though that in the macros/validation folder there is only one .init file to run the K40 events, but no others for the other config files. I wonder if it makes sense to give this init file a generic name and then add options in this file for running the other configs in that folder?

paolafer:
I like the idea, but there's a problem. In the automatic tests, when we run all the examples, the code assumes that each `config` file is named exactly the same as the `init` file. Another possibility would be to keep just a pair of macros for one isotope and comment the use of the other three isotopes in the `config` file.

kvjmistry:
Yes that works! 

Probably call the pair something generic like `NEW_valid_isotope.init.mac` and `NEW_valid_isotope.config.mac` or if you have a better name we can use that!

kvjmistry (approved):
Looks good to me! Approved!

PR #199, opened 2023-02-23
Remove superfluous examples of macros

paolafer:
This PR removes two sets of macros, which are considered obsolete now. They are simple examples of the usage of `SingleParticleGenerator`, which is illustrated in a variety of other examples already. The `NEXT100.Neutron` pair of macros is retained to show an example of neutron generation and several examples of muon generation are already there.
Therefore, I prefer to remove these extra examples to speed up the test execution and to keep the examples to a limited number for clarity.


halmamol (approved):
I agree that these 2 set of macros are not necessary considering the existance of other `SingleParticle` generator examples (like the mentioned _NEXT100.Neutron_, or the _NEW_geantino_ pair of macros). I've crosschecked, they are also not referenced on the Wiki so after beign removed we don't need to edit it. Approved!

PR #200, opened 2023-03-02
Store step times in SaveAllSteppingAction

gonzaponte:
Step time is defined as the average between the pre- and post-step time.

### DIFF
--- a/source/actions/SaveAllSteppingAction.cc
+++ b/source/actions/SaveAllSteppingAction.cc
@@ -1,8 +1,19 @@
 // ----------------------------------------------------------------------------
 // nexus | SaveAllSteppingAction.cc
 //
-// This class sets the store_steps flag of the persistency manager
-// so all steps are saved.
+// This class adds a new group and table to the output file, "/DEBUG/steps".
+// This table contains information (position and volume of both the
+// pre- and post-step points, average time, process name and other identifiers)
+// of some steps of the simulation. By default all steps are stored. However,
+// a subset of them can be selected by cherry-picking the volumes and particles
+// involved in the step. This can be achieved with the commands
+// /Actions/SaveAllSteppingAction/select_particle
+// and
+// /Actions/SaveAllSteppingAction/select_volume
+// without the need for re-compilation.
+// It must be noted that the files produced with this action become large
+// very quickly. Therefore, strict filtering and small number of events are
+// encouraged.
 //
 // The NEXT Collaboration
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


source/actions/SaveAllSteppingAction.cc
```diff
@@ -63,6 +63,8 @@ void SaveAllSteppingAction::UserSteppingAction(const G4Step* step)
 
   G4ThreeVector initial_pos = pre ->GetPosition();
   G4ThreeVector   final_pos = post->GetPosition();
+  G4double        step_time = (pre->GetGlobalTime()  +
```
paolafer, line 77: Are you sure that you want the average time of the steps and not the time at the beginning or at the end? What do you use this for?

paolafer:
`
// This class sets the store_steps flag of the persistency manager
// so all steps are saved.`

Could you improve a little the description of the class? It does other things beyond setting a flag.

source/actions/SaveAllSteppingAction.cc
gonzaponte: I don't have a preference. I just want to have a handle on the time of the steps so I can potentially build a waveform out of them. Saving both pre and post seemed a bit too much, but I don't mind doing that and taking the average offline.

paolafer: It's just that in G4 the "borders" of a step are well defined in terms of position, volume... while what happens "inside" the step is sometimes more confuse. However, since yours is the only user case so far and you want the average, we can go ahead with that.

source/actions/SaveAllSteppingAction.cc
```diff
@@ -1,8 +1,19 @@
 // ----------------------------------------------------------------------------
-// nexus | SaveAllSteppingAction.cc
+// nexus | SaveAllSteppingAction.h
```
paolafer: Why did you change the extension? It is `.cc`, right?

source/actions/SaveAllSteppingAction.cc
```diff
@@ -1,8 +1,19 @@
...
+// pre- and post-step points, average time, process name and other identifiers)
+// of some steps of the simulation. By default all steps are stored. However,
+// a subset of them can be selected by cherry-picking the volumes and particles
+// involved in the step. This can be achieved with the commands
+// /Actions/SaveAllSteppingAction/select_particle
+// and
+// /Actions/SaveAllSteppingAction/select_volume
+// without the need for re-compilation.
+// It must be noted that the files produced with this action become large
+// very quicly. Therefore, strict filtering and small number of events are
```
paolafer: Typo: quicly -> quickly.

source/actions/SaveAllSteppingAction.cc
gonzaponte: Yes, the wonders of copy-pasting...

paolafer (approved):
Useful addition in this stepping action, approved!

PR #201, opened 2023-03-06
Fix for Geant4 1.11

atrettin:
This PR adds include lines that are necessary to define the `G4double` type. This is necessary as of Geant4 version > 11.0.

Also, as a small bonus it adds VSCode configurations to gitignore.


atrettin:
closing in order to rename branch

PR #202, opened 2023-03-06
Fix for Geant4 1.11

atrettin:
This PR adds include lines that are necessary to define the G4double type, as necessary as of Geant4 version >11.0.

Also adds .vscode to gitignore to ignore VSCode configuration.

Now with `dev` branch name on my fork.

### DIFF
--- a/source/physics/RadiusDependentDriftField.h
+++ b/source/physics/RadiusDependentDriftField.h
@@ -12,7 +12,6 @@
 
 #include "BaseDriftField.h"
 #include <G4LorentzVector.hh>
-#include <G4Types.hh>
 
 
 namespace nexus {
--- a/source/physics/RadiusDependentDriftField.h
+++ b/source/physics/RadiusDependentDriftField.h
@@ -12,6 +12,7 @@
 
 #include "BaseDriftField.h"
 #include <G4LorentzVector.hh>
+#include <G4Types.hh>
 
 
 namespace nexus {
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 .sconsign.dblite
 source/.DS_Store
 *~
+.vscode
 
 # nexus output files
 *.h5
--- a/source/physics/BaseDriftField.h
+++ b/source/physics/BaseDriftField.h
@@ -11,6 +11,7 @@
 
 #include <G4VUserRegionInformation.hh>
 #include <G4LorentzVector.hh>
+#include <G4Types.hh>
 
 class G4Material;
 


source/physics/RadiusDependentDriftField.h
```diff
@@ -12,6 +12,7 @@
 
 #include "BaseDriftField.h"
 #include <G4LorentzVector.hh>
+#include <G4Types.hh>
```
kvjmistry: Do we need this for it to compile?
This file includes the BaseDriftField.h which already has the header.

kvjmistry:
Ok, if I add the header only to "BaseDriftField.h" it compiles fine using the `cmake`. But for using scons it breaks on the linking, something relating to the visualiser in geant4. Did you compile with scons too?
```
Undefined symbols for architecture arm64:
  "G4ToolsSGOffscreen::G4ToolsSGOffscreen()", referenced from:
      G4VisExecutive::RegisterGraphicsSystems() in nexus.o
ld: symbol(s) not found for architecture arm64
```

atrettin:
> Ok, if I add the header only to "BaseDriftField.h" it compiles fine using the `cmake`. But for using scons it breaks on the linking, something relating to the visualiser in geant4. Did you compile with scons too?
> 
> ```
> Undefined symbols for architecture arm64:
>   "G4ToolsSGOffscreen::G4ToolsSGOffscreen()", referenced from:
>       G4VisExecutive::RegisterGraphicsSystems() in nexus.o
> ld: symbol(s) not found for architecture arm64
> ```

That is odd, because the error also happens in RadiusDependenDriftField (and given that both use the `Geant4double` type, it would be weird if the import was only required for one of them).

I did compile with some sort of visualization, but I'm not sure if it was "scons" specifically.

In the earlier PR, the workflow also ran without problems, so it should compile fine?

kvjmistry:
Ah thats because the RadiusDependantDriftField.h has this line: #include "BaseDriftField.h"
So you only need to add it to BaseDriftField.h. Are you saying it only compiles if you add it to both?

atrettin:
> Ah thats because the RadiusDependantDriftField.h has this line: #include "BaseDriftField.h"
> So you only need to add it to BaseDriftField.h. Are you saying it only compiles if you add it to both?

Ah, sorry, that makes sense of course. I didn't see that. Of course, then it is only required in one of them.

kvjmistry:
For the compilation, can you try using the command `scons` in the main directory and see if it compiles? You may need to pass it some extra vars too like this
```
scons GEANT4_BINDIR=$G4INSTALL/bin GSL_BINDIR=$GSL_PATH/bin HDF5_DIR=$HDF5_PATH -j4
```

atrettin:
> For the compilation, can you try using the command `scons` in the main directory and see if it compiles? You may need to pass it some extra vars too like this
> 
> ```
> scons GEANT4_BINDIR=$G4INSTALL/bin GSL_BINDIR=$GSL_PATH/bin HDF5_DIR=$HDF5_PATH -j4
> ```

I will, tomorrow! :)

atrettin:
It works for me, in both cases where I add or remove the line in `RadiusDependentDriftField`. I installed Geant4 with the settings: ```{"GEANT4_USE_OPENGL_X11": "ON", "GEANT4_INSTALL_DATA": "ON", "GEANT4_USE_QT": "ON"}```

atrettin:
But if we have these two build methods, it would be a good idea to add both of them to the GitHub workflow (but I think that goes beyond the scope of this PR).

atrettin:
Speaking of which, should we bump the Geant4 version in the workflow in `docker/geant4/Dockerfile`?

paolafer:
> Speaking of which, should we bump the Geant4 version in the workflow in docker/geant4/Dockerfile?

Before changing the recommended G4 version, we usually do a validation, to check that the relevant physics stays the same across versions. If you're interested in doing this, we can talk offline.

paolafer:
> But if we have these two build methods, it would be a good idea to add both of them to the GitHub workflow (but I think that goes beyond the scope of this PR).

Yes, we can do it in a different PR.

kvjmistry (approved):
This PR adds a header include and updates the .gitignore file. The header include is for compilation with G4v11.1.0 or greater. There are still issues with this geant4 version and nexus (subject to another PR in the future), but this PR can be merged in since it's simple. 

Congrats on your first contribution to nexus Alex!

PR #203, opened 2023-03-08
Fix order of energies

paolafer:
The energy vector of the wavelength shifter property had a bug, being the last value of energy lower than the second-to-last one (energies in property vectors must be in increasing order). This PR fixes the bug and replaces the last value of absorption length, which was infinite, with a value closer to the real value appearing in the reference plot. In my opinion this is a more correct approach, since we know that at that energy the absorption length is not infinite.

### DIFF
--- a/source/materials/OpticalMaterialProperties.cc
+++ b/source/materials/OpticalMaterialProperties.cc
@@ -867,7 +867,7 @@ namespace opticalprops {
       h_Planck * c_light / (250. * nm),  h_Planck * c_light / (230. * nm),
       h_Planck * c_light / (210. * nm),  h_Planck * c_light / (190. * nm),
       h_Planck * c_light / (170. * nm),  h_Planck * c_light / (150. * nm),
-      h_Planck * c_light / (100. * nm),  optPhotMaxE_
+      h_Planck * c_light / (108. * nm),  optPhotMaxE_
     };
 
     std::vector<G4double> WLS_absLength = {
@@ -879,7 +879,7 @@ namespace opticalprops {
       400. * nm,     400. * nm,     // 250 , 230 nm
       350. * nm,     250. * nm,     // 210 , 190 nm
       350. * nm,     400. * nm,     // 170 , 150 nm
-      400. * nm,     noAbsLength_   // 100 nm
+      400. * nm,     400. * nm   // 100 nm
     };
 
     //for (int i=0; i<WLS_abs_entries; i++)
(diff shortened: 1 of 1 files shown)


halmamol:
I'm having issues trying to read the absorption length plot (to see where do you get the 108 nm value, basically). Could you point it to me offline maybe?

source/materials/OpticalMaterialProperties.cc
```diff
@@ -879,7 +879,7 @@ namespace opticalprops {
       400. * nm,     400. * nm,     // 250 , 230 nm
       350. * nm,     250. * nm,     // 210 , 190 nm
       350. * nm,     400. * nm,     // 170 , 150 nm
-      400. * nm,     noAbsLength_   // 100 nm
+      400. * nm,     400. * nm   // 100 nm
```
halmamol: to keep consistency with previous lines, you should change 100nm to 108nm on the commented part (so it's easy to read the plots)

halmamol (commented):
I also wonder, do we need to keep what is between the commented lines 827 and 856?

paolafer:
> I'm having issues trying to read the absorption length plot (to see where do you get the 108 nm value, basically). Could you point it to me offline maybe?

The 108 nm value is arbitrary, I just wanted to obtain an energy lower than `optPhotMaxE_` and as close as possible to it. Since in that range the absorption length is always 400 nm, according to the plot, the exact value is not relevant. In fact, I think we can simplify it even more, removing that value completely.

paolafer:
> I also wonder, do we need to keep what is between the commented lines 827 and 856?

That was kept by @jmunozv, when we changed the way we implemented the property to avoid secondary wavelength shifting. I think we can keep it, for the moment, since the wavelength shifting process can still be improved.

halmamol:
> The 108 nm value is arbitrary, I just wanted to obtain an energy lower than `optPhotMaxE_` and as close as possible to it. Since in that range the absorption length is always 400 nm, according to the plot, the exact value is not relevant. In fact, I think we can simplify it even more, removing that value completely.

I see what you mean, and I like how you simplified it now (including the commented text).

halmamol (approved):
This PR fixes a bug on the wavelength shifter energies. It removes the second to last value, so now the energies are on increasing order (as it should be on vectors). The value is removed since on that region the absortion values are constant and it is not necessary to add a point there. In addition it adds some comments on the lines for future reading. Good catch, I approve it!

PR #204, opened 2023-03-19
CORSIKA generator

soleti:
This PR adds a new generator which takes as input binary files generated by the [CORSIKA cosmic-ray generator](https://www.iap.kit.edu/corsika/index.php). An example of CORSIKA binary file useful for testing can be found [here](https://www.dropbox.com/s/j585iqr287xokxs/DAT000001?dl=0). The code is adapted from https://github.com/soleti/corsika2RooTracker.


paolafer:
Why don't you add a test using this generator? I think that would be useful.

soleti:
> Why don't you add a test using this generator? I think that would be useful.

I add a macro with a generate the cosmic rays taken from the CORSIKA binary file above a generic xenon sphere.

PR #205, opened 2023-03-20
Use total drift length of EL gap to generate EL photons

paolafer:
Due to a change in `Geant4-11.1`, the step length returned by the ionization electron crossing the EL gap is 0 instead of the gap length. An effect of this is that the number of EL photons is 0, as well.
This PR modifies the way we calculate the number of EL photons, using the actual length of the gap instead of the step length, to overcome this problem.

### DIFF
--- a/source/physics/BaseDriftField.h
+++ b/source/physics/BaseDriftField.h
@@ -40,6 +40,7 @@ namespace nexus {
       GeneratePointAlongDriftLine(const G4LorentzVector&, const G4LorentzVector&) = 0;
 
     virtual G4double LightYield() const;
+    virtual G4double GetTotalDriftLength() const;
 
   private:
     void Print() const;
@@ -52,6 +53,8 @@ namespace nexus {
   inline BaseDriftField::~BaseDriftField() {}
 
   inline G4double BaseDriftField::LightYield() const {return 0.;}
+
+  inline G4double BaseDriftField::GetTotalDriftLength() const {return 0.;}
 
   inline void BaseDriftField::Print() const {}
 
--- a/source/physics/Electroluminescence.cc
+++ b/source/physics/Electroluminescence.cc
@@ -79,7 +79,7 @@ Electroluminescence::PostStepDoIt(const G4Track& track, const G4Step& step)
 
   // Get the light yield from the field
   const G4double yield = field->LightYield();
-  G4double step_length = step.GetStepLength();
+  G4double step_length = field->GetTotalDriftLength();
 
   if (yield <= 0.)
     return G4VDiscreteProcess::PostStepDoIt(track, step);
--- a/source/physics/UniformElectricDriftField.cc
+++ b/source/physics/UniformElectricDriftField.cc
@@ -115,4 +115,11 @@ namespace nexus {
   }
 
 
+
+  G4double UniformElectricDriftField::GetTotalDriftLength() const
+  {
+    return std::abs(anode_pos_ - cathode_pos_);
+  }
+
+
 } // end namespace nexus
--- a/source/physics/UniformElectricDriftField.h
+++ b/source/physics/UniformElectricDriftField.h
@@ -64,6 +64,8 @@ namespace nexus {
     void SetNumberOfPhotons(G4double);
     G4double GetNumberOfPhotons() const;
 
+    virtual G4double GetTotalDriftLength() const;
+
   private:
     /// Returns true if coordinate is between anode and cathode
     G4bool CheckCoordinate(G4double);


kvjmistry (commented):
Currently testing this new calculation of the EL gap length with the current approved release G4 v11.0.2 using the `DEMOPP_fullKr.init.mac` macro as a test case (so we can work with a reproducible macro). I notice one thing: the new and old methods of calculating the drift length in the EL gap disagree. The old method of getting the step produces a variable EL gap size whereas the new method produces a fixed value. I belive this difference is due to diffusion in the gap where the `x` and `y` coordinate difference is taken into account. The new method is only assuming a step length along `z`.

A print of the step sizes with v11.0.2 shows the following:
```
Debug EL, Step Length NEW: 309.55
Debug EL, Step Length OLD: 1

Debug EL, Step Length NEW: 9.8
Debug EL, Step Length OLD: 9.81966

Debug EL, Step Length NEW: 309.55
Debug EL, Step Length OLD: 1

Debug EL, Step Length NEW: 9.8
Debug EL, Step Length OLD: 9.88621
```
Not totally sure what the values at 300 are, but ignoring those, you can see the different step sizes being calculated in the old method. The new method is all the same value.

Testing with the new G4 release using v11.1.0 I indeed find the step sizes are all 0 in the old method showing the change in behaviour:

```
Debug EL, Step Length NEW: 309.55
Debug EL, Step Length OLD: 0

Debug EL, Step Length NEW: 9.8
Debug EL, Step Length OLD: 0

Debug EL, Step Length NEW: 309.55
Debug EL, Step Length OLD: 0

Debug EL, Step Length NEW: 9.8
Debug EL, Step Length OLD: 0

Debug EL, Step Length NEW: 309.55
Debug EL, Step Length OLD: 0

Debug EL, Step Length NEW: 9.8
Debug EL, Step Length OLD: 0

Debug EL, Step Length NEW: 309.55
Debug EL, Step Length OLD: 0

Debug EL, Step Length NEW: 9.8
Debug EL, Step Length OLD: 0
```

paolafer:
Thanks, Krishan! I think you're right and the small difference of calculated step length in the old version is due to diffusion. On the other hand, I noticed that the step length of the drift part is miscalculated (1 is given, in your example), although that value is not used anywhere in the simulation, therefore we never realized or tried to fix it. All considered, I think that returning the actual value of the gap is more correct. The small differences due to diffusion should not introduce a significant error in the EL yield; however, we should compare the two distributions of the yield as it is calculated in the `Electroluminescence` class, using the old method and the new one, to quantify this.

kvjmistry:
Just one note: 
I am debugging the yields and one thing that isn't quite clear in the code is the unit of the yield value that comes out from `field->LightYield()`. It appears to be in photons/mm rather than photons/cm. You can see this as we multiply the number by 9.8 to get the mean yield and 9.8 is units of mm. The `XenonELLightYield` function in `XenonProperties.cc` indicates in the comment that the yield is in photons/cm.

Just mentioning this as I think this should be updated. Maybe return as `yield/cm` and fix the units in the calculation of the mean yield? or update some of the comments?

paolafer:
> Just one note: I am debugging the yields and one thing that isn't quite clear in the code is the unit of the yield value that comes out from `field->LightYield()`. It appears to be in photons/mm rather than photons/cm. You can see this as we multiply the number by 9.8 to get the mean yield and 9.8 is units of mm. The `XenonELLightYield` function in `XenonProperties.cc` indicates in the comment that the yield is in photons/cm.
> 
> Just mentioning this as I think this should be updated. Maybe return as `yield/cm` and fix the units in the calculation of the mean yield? or update some of the comments?

Nexus uses the Geant4 internal units, which are `mm` in the case of length, so I think that we can update the comment, removing  the mention to `cm`.

kvjmistry:
Here is my comparison using the same Kr macro, 300 events:
- v11.0.2 "old" is using the old method of getting the step length `step.GetStepLength();`
- v11.0.2  "new" is using the new method of the step length `field->GetTotalDriftLength()`
- v11.1.0 "new" is using the new g4 version and the new method of the step length `field->GetTotalDriftLength()` 

image

I fit the distributions, both the new implementations regardless of g4 versions I deem are the same. Compared with the old method, there is a difference in the overall yield distribution due to the difference in how the gap length is calculated.

```
Mean OLD:  1311386
Sigma OLD:  1.1%

Mean NEW:  1297357
Sigma NEW:  0.9%
```
The percent difference in mean of new and old methods is 1.1%

paolafer:
> Here is my comparison using the same Kr macro, 300 events:
> 
> * v11.0.2 "old" is using the old method of getting the step length `step.GetStepLength();`
> * v11.0.2  "new" is using the new method of the step length `field->GetTotalDriftLength()`
> * v11.1.0 "new" is using the new g4 version and the new method of the step length `field->GetTotalDriftLength()`
> 
> image
> 
> I fit the distributions, both the new implementations regardless of g4 versions I deem are the same. Compared with the old method, there is a difference in the overall yield distribution due to the difference in how the gap length is calculated.
> 
> ```
> Mean OLD:  1311386
> Sigma OLD:  1.1%
> 
> Mean NEW:  1297357
> Sigma NEW:  0.9%
> ```
> 
> The percent difference in mean of new and old methods is 1.1%

Thanks, Krishan! I'd say that this difference won't impact on our total energy resolution, taking into account the different contributions, as explained, for instance, [here](https://www.sciencedirect.com/science/article/abs/pii/S0168900209002897), but we should bring the issue up at the next meeting to take a decision.

kvjmistry (approved):

PR #209, opened 2023-05-25
Update scintillation time constants and yield for gaseous Ar

Aretno:
The values are extracted from 'Triplet Lifetime in Gaseous Argon' by Michael Akashi-Ronquest. The slow component is greatly slower than the previous value in our simulation. The scintillation yield has also been adjusted, being the slow component 6.6 times larger than the fast.


paolafer (approved):
This PR updates Ar optical properties: approved!

PR #210, opened 2023-07-14
Change step size to reasonable value

paolafer:
This PR changes the value of the step size in a couple of examples to 1 mm, which is more accurate than the old value.

### DIFF
--- a/macros/NEXT100.config.mac
+++ b/macros/NEXT100.config.mac
@@ -16,7 +16,7 @@
 
 ##### GEOMETRY #####
 /Geometry/Next100/elfield false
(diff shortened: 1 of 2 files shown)


FabianKellerer (commented):
Has the impact of this change on the performance been investigated?

paolafer:
Yes, in fact several studies have been done, the most recent ones being those by Ben, which point to a better behaviour with 1 mm. 1 mm is the value used in all NEXT productions; these macros weren't updated at the time we made the decision and they kept the old value.

FabianKellerer (approved):
Looks good to me!

PR #211, opened 2023-10-04
Modify code and tests to avoid creating output files when tests run

paolafer:
This PR modifies tests and a few classes to prevent tests from creating permanent files.
Now said files (hdf5 and csv) are created in a temporary directory, which is deleted at the end of the test run.


jwaiton (approved):
### Approved

All tests pass. I receive some warnings related to deprecated functions which we've assumed are due to python package differences. These don't appear to be a cause for concern.

Output from tests contained within tmp/ directory on my computer, which is regularly wiped. So works as expected :+1: 

OS: Ubuntu 22.04.3 LTS x86_64
Kernel: 6.2.0-34-generic

PR #214, opened 2023-10-17
Stop using bitwise operators with Booleans

paolafer:
This PR fixes the use of bitwise operators with booleans. Although in nexus the behaviour was the expected one, it could be misleading, and it raised compilation warnings.


atrettin (approved):
This looks straight forward enough! Yes, `&&` should always be used when combining logical expressions to ensure correct short-circuiting behavior, although these expressions look like it wouldn't make a lot of difference. I approve these changes.

PR #215, opened 2023-10-27
Adjust/update Next100 geometry

jwaiton:
- Altered `Next100OpticalGeometry.cc` to match the more up to date distances between the sapphire windows, the EL gate, and the tracking plane that were introduced to `Next100.cc` in [this PR](https://github.com/next-exp/nexus/pull/121) but were not carried over.
- Changed length of inner copper shielding to match drawings (copper should not overlap beyond the staves on the energy plane-side of the detector).
- Adjusted positions of calibration and feed-through ports to match the expected locations in the detector.

On my machine both `pytest` and the Catch2 tests passed (with the pytest containing warnings that appeared unrelated to this change).

General summary of geometry changes can be seen here.
[geometry_changes.pdf](https://github.com/next-exp/nexus/files/13190745/geometry_changes.pdf)

### DIFF
--- a/source/geometries/Next100Ics.cc
+++ b/source/geometries/Next100Ics.cc
@@ -54,8 +54,7 @@ namespace nexus {
 
   void Next100Ics::Construct()
   {
-    G4double length = gate_tp_distance_ + gate_sapphire_wdw_dist_;
-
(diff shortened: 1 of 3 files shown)


paolafer (approved):
This PR fixes the position of the hole in the ICS, as well as its dimensions. Good job, @jwaiton!

Issue #216, opened 2023-11-06
FusedSilica and FakeFusedSilica refraction indices not valid


soleti:
The formula for the refractive indices for `FusedSilica` and `FakeFusedSilica` is valid only up to ~10.7 eV. However, the values are calculated in the code up to `optPhotMaxE_`, which is set to 11.5 eV:

https://github.com/next-exp/nexus/blob/5389016be8a34058c745d53f181be3ab66cdf8a0/source/materials/OpticalMaterialProperties.h#L88

Thus, this value:

https://github.com/next-exp/nexus/blob/5389016be8a34058c745d53f181be3ab66cdf8a0/source/materials/OpticalMaterialProperties.cc#L82

is `nan` for points above 10.7 eV.

paolafer:
Solved in #217.

PR #217, opened 2023-11-06
Fix FusedSilica and FakeFusedSilica refractive indices

soleti:
This PR fixes issue #216. Now, the refractive indices for `FusedSilica` and `FakeFusedSilica` are calculated only up to 10.7 eV, which is the upper limit for the formula used in the code.

### DIFF
--- a/source/materials/OpticalMaterialProperties.cc
+++ b/source/materials/OpticalMaterialProperties.cc
@@ -50,7 +50,7 @@ namespace opticalprops {
     // The range is chosen to be up to ~10.7 eV because Sellmeier's equation
     // for fused silica is valid only in that range
     const G4int ri_entries = 200;
-    G4double eWidth = (optPhotMaxE_ - optPhotMinE_) / ri_entries;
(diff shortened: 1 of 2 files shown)


paolafer:
The problem is that we can have photons with higher energies simulated (e.g., Cherenkov photons), so we should define a refractive index up to 11.5 eV in some way. Are there not any measurements at that energy?

soleti:
I see, then maybe setting the refractive index between 10.7 and 11.5 eV at the value obtained at 10.7 eV? I couldn't find measurements below 115 nm (which is 10.7 eV).

paolafer:
> I see, then maybe setting the refractive index between 10.7 and 11.5 eV at the value obtained at 10.7 eV? I couldn't find measurements below 115 nm (which is 10.7 eV).

Fine for me!

paolafer (approved):
This PR fixes a wrong behaviour of the refractive index of quartz at high energies. Approved!

PR #218, opened 2023-11-06
Update Field Cage class

paolafer:
This PR reviews the field cage thoroughly, using the latest drawings as a reference, with measurements taken by Sara C. and Jordi T.
The main changes compared with the previous version are the following:
- Move the cathode grid to the edge of the cathode ring, facing the active volume (following pictures taken during the installation).  
- Use real dimensions and placements of the staves (ring holders), decoupling them from the teflon reflector panels.
- Encapsulate the stave code in one class, glueing together the 3 parts that makes it up (drift, buffer and cathode parts).
- Update dimensions of poly wrap, teflon reflector panels and cathode, gate and anode ring (the latter is simplified).
- Fix the distances between gate and tracking plane and between gate and teflon panels.

### DIFF
--- a/source/geometries/Next100.cc
+++ b/source/geometries/Next100.cc
@@ -37,7 +37,7 @@ namespace nexus {
     // Lab dimensions
     lab_size_ (5. * m),
 
-    // common used variables in geometry components
+    // common variables used in geometry components
     grid_thickness_ (0.1 * mm),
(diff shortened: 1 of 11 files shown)


jwaiton (approved):
All tests passed and vertex generation/ordering of components within the detector appear consistent with relation to code and physical expectations. 

The inclusion of staves into a separate class and as a union of the three objects previously produced separately is a very nice change. Good job! 

Approved :)

PR #219, opened 2023-11-07
Review mask dimensions according to Sara's drawings

paolafer:
It turned out that the mask dimensions were following an obsolete technical drawing with round holes. I attach the latest drawing, provided by Sara, which is implemented in this PR.
[03-95 Mask Hama 6 mm PASANTE - .pdf](https://github.com/next-exp/nexus/files/13277419/03-95.Mask.Hama.6.mm.PASANTE.-.pdf)


jwaiton (approved):
Adjustments to SiPM board geometry match those from more recent drawings, and passed all tests locally on my machine. Approved! :)

PR #220, opened 2023-11-07
Add inner flange

paolafer:
This PR adds a piece of the vessel flange on the side of the energy plane (EP), which was not simulated previously. 
Its sits between the EP copper plate and the inner copper shielding (ICS). 
In the space occupied by this new volume there were previously a part of the ICS bars and gas; therefore it may be relevant for the shielding of background.

### DIFF
--- a/source/geometries/Next100.cc
+++ b/source/geometries/Next100.cc
@@ -42,6 +42,7 @@ namespace nexus {
     //ep_fc_distance_ (24.8 * mm), /// this value was measured by Sara in the step file
     gate_tracking_plane_distance_(25. * mm + grid_thickness_), // Jordi = 1.5 (distance TP plate-anode ring) + 13.5 (anode ring thickness) + 10 (EL gap)
     gate_sapphire_wdw_distance_  (1458.8 * mm - grid_thickness_), // Jordi
+    ics_ep_lip_width_ (55. * mm),
 
     specific_vertex_{},
     lab_walls_(false)
@@ -65,10 +66,10 @@ namespace nexus {
   hallA_walls_ = new LSCHallA();
 
   // Vessel
-  vessel_ = new Next100Vessel();
+  vessel_ = new Next100Vessel(ics_ep_lip_width_);
 
   // Internal copper shielding
-  ics_ = new Next100Ics();
+  ics_ = new Next100Ics(ics_ep_lip_width_);
 
   // Inner Elements
   inner_elements_ = new Next100InnerElements(grid_thickness_);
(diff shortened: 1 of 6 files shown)


jwaiton:
When trying to look at the new geometry, the interactive mode of nexus hangs on a black screen when I attempt to open it.

The command I type is:
`bin/nexus -n 1 -i macros/NEXT100.geantino.init.mac`
Where this macro is defined [here](https://github.com/jwaiton/next_misc/blob/main/nexus_PR/NEXT100.geantino.init.mac) and its config [here](https://github.com/jwaiton/next_misc/blob/main/nexus_PR/NEXT100.geantino.config.mac).

Removing the interactive mode (-i) allows the software to run unhindered. Switching to a macro of a differing geometry (such as `NEW_geantino.init.mac`) allows for the interactive mode to work, and reverting to my master branch allows the same macro to function as expected.

Is there a particular reason why this may be the case? I assumed it may be an issue with my macro, but I'm unsure due to the macro working in my master branch.

One test in `pytest` also fails, that being `macros_test.py::test_run_fast_examples`.

paolafer:
Hi @jwaiton! The links to the macros are not working for me. 

However, I'm aware of an issue in the visualization of the inner copper shielding after the modification of this PR (at least on my laptop). I haven't been able to figure out the reason, but if you set the visibility of the ICS class to false, the rest of volumes are visualized correctly. I'm surprised it fails with the geantino macro, though, because it should have the visibility of ICS set to false (can you confirm that?).
The NEW macro works because the code doesn't include the NEXT-100 ICS class.

Regarding the failure of the test, can you copy the error and identify where it fails (running which macro)? Automatic tests pass on my laptop and also on the GitHub server, though.

jwaiton:
Apologies @paolafer, I thought I made that repo public but I hadn't. The macros are now available to see :)

I have personally altered the visibility of the ICS class to visually inspect the changes, hence why it must be crashing. I can confirm that removing the ICS class visibility allows it to work as expected.

For the pytest error, I tried to determine it through the implementation of `pytest` but the error appears to be cut off before it's output is completed. The output is as follows:

`=========================== short test summary info ============================
FAILED tests/pytest/macros_test.py::test_run_fast_examples - subprocess.CalledProcessError: Command '['/home/e78368jw/Documents/NEXT_COD...
============ 1 failed, 36 passed, 163 warnings in 548.86s (0:09:08) ============
`

I resultantly tried to read through the `macros_test.py` code but I couldn't find any unique insight into the error source. Is there an output log somewhere that might return more information about the error?

paolafer:
The error message doesn't help... What you can do is to look at the output to screen while the fast macros tests are running: you will see the name of the macros appearing on screen and, as soon as the problematic one crashes, the tests should fail. Can you try that?

jwaiton:
The failure of pytest was due to a mistake on my part, apologies.

I've continued looking into this. I found no overlaps between the ICS and anything else that may be causing issues, and checking the vertices explicitly outputs what's expected (ep lip apparent in the ICS in the expected location).

I'll try playing around with it a bit to see if I isolate the issue more specifically.

jwaiton (approved):
This PR adds the vessel flange to the simulation. Tests have been completed to ensure that the shape of the ICS/vessel in the simulation is as desired.

There are issues with the visualisation tied directly to the subtraction of the ICS in this change (long load times for visualising the ICS), but there doesn't appear to be a reasonable solution to this, and the simulation produces the expected outputs from this change.

Good job @paolafer!

PR #222, opened 2023-11-14
FC displacement

paolafer:
This PR displaces the field cage volumes in `xy`, trying to reproduce the real displacement at LSC as much as possible. This displacement is caused by having two HDPE layers only in one sector of the cylinder, and not all over the place.
Modifications have been made in order to have the `ACTIVE` volume centred in the TPC axis.

### DIFF
--- a/source/geometries/GeometryBase.h
+++ b/source/geometries/GeometryBase.h
@@ -55,6 +55,12 @@ namespace nexus {
     /// Setter for the starting point of EL generation in z
     void SetELzCoord(G4double z);
 
+    /// Getter for the origin of coordinates
+    G4ThreeVector GetCoordOrigin() const;
+
+    /// Setter for the origin of coordinates
+    void SetCoordOrigin(G4ThreeVector origin);
+
     /// Translates position to G4 global position
     void CalculateGlobalPos(G4ThreeVector& vertex) const;
 
@@ -90,12 +96,13 @@ namespace nexus {
     G4ThreeVector dimensions_; ///< XYZ dimensions of a regular geometry
     G4bool drift_; ///< True if geometry contains a drift field (for hit coordinates)
     G4double el_z_; ///< Starting point of EL generation in z
+    G4ThreeVector coord_origin_; ///< Origin of coordinates of the mother volume
   };
 
 
   // Inline definitions ///////////////////////////////////
 
-  inline GeometryBase::GeometryBase(): logicVol_(0), span_(25.*m), drift_(false), el_z_(0.*mm) {}
+  inline GeometryBase::GeometryBase(): logicVol_(0), span_(25.*m),  el_z_(0.*mm), coord_origin_(G4ThreeVector(0., 0., 0.)) {}
 
   inline GeometryBase::~GeometryBase() {}
 
@@ -128,6 +135,10 @@ namespace nexus {
   inline G4double GeometryBase::GetELzCoord() const {return el_z_;}
 
   inline void GeometryBase::SetELzCoord(G4double z) {el_z_ = z;}
+
+  inline G4ThreeVector GeometryBase::GetCoordOrigin() const {return coord_origin_;}
+
+  inline void GeometryBase::SetCoordOrigin(G4ThreeVector origin) {coord_origin_ = origin;}
 
   // This methods is to be used only in the Next1EL and NEW geometries
   inline void GeometryBase::CalculateGlobalPos(G4ThreeVector& vertex) const
--- a/source/geometries/GeometryBase.h
+++ b/source/geometries/GeometryBase.h
@@ -46,9 +46,6 @@ namespace nexus {
     /// Returns the 3 dimensions of the geometry (x, y, z)
     G4ThreeVector GetDimensions();
 
-    /// Returns true if geometry has a drift field
-    G4bool GetDrift() const;
-
     // Getter for the starting point of EL generation in z
     G4double GetELzCoord() const;
 
@@ -78,9 +75,6 @@ namespace nexus {
     /// Sets the span (maximum dimension) of the geometry
     void SetSpan(G4double);
 
-    /// Sets the drift variable to true if a drift field exists
-    void SetDrift(G4bool);
-
     /// Sets the 3 dimensions of the geometry (x, y, z)
     void SetDimensions(G4ThreeVector dim);
 
@@ -94,7 +88,6 @@ namespace nexus {
     G4LogicalVolume* logicVol_; ///< Pointer to the logical volume
     G4double span_; ///< Maximum dimension of the geometry
     G4ThreeVector dimensions_; ///< XYZ dimensions of a regular geometry
-    G4bool drift_; ///< True if geometry contains a drift field (for hit coordinates)
     G4double el_z_; ///< Starting point of EL generation in z
     G4ThreeVector coord_origin_; ///< Origin of coordinates of the mother volume
   };
@@ -127,10 +120,6 @@ namespace nexus {
   inline void GeometryBase::SetDimensions(G4ThreeVector dim) {  dimensions_ = dim; }
 
   inline  G4ThreeVector GeometryBase::GetDimensions()  { return dimensions_; }
-
-  inline void GeometryBase::SetDrift(G4bool drift) { drift_ = drift; }
-
-  inline G4bool GeometryBase::GetDrift() const { return drift_; }
 
   inline G4double GeometryBase::GetELzCoord() const {return el_z_;}
 
--- a/source/geometries/Next100.cc
+++ b/source/geometries/Next100.cc
@@ -123,7 +123,11 @@ namespace nexus {
     this->SetLogicalVolume(lab_logic_);
 
     // VESSEL (initialize first since it defines EL position)
+    // The z coord origin is not set, because it is not defined, yet
     vessel_->SetELtoTPdistance(gate_tracking_plane_distance_);
+    vessel_->SetCoordOrigin(G4ThreeVector(fc_displ_x_,
+                                          fc_displ_y_,
+                                          0.));
     vessel_->Construct();
     G4LogicalVolume* vessel_logic = vessel_->GetLogicalVolume();
     G4LogicalVolume* vessel_internal_logic  =
--- a/source/geometries/Next100.cc
+++ b/source/geometries/Next100.cc
@@ -318,7 +318,7 @@ namespace nexus {
 		  "Unknown vertex generation region!");
     }
 
-    return vertex + G4ThreeVector(0., 0., -gate_zpos_in_vessel_);
+    return vertex - coord_origin_;
   }
 
 } //end namespace nexus
--- a/source/geometries/Next100.cc
+++ b/source/geometries/Next100.cc
@@ -47,7 +47,8 @@ namespace nexus {
     fc_displ_y_ (-6.4 * mm), // displacement of the field cage volumes from 0
 
     specific_vertex_{},
-    lab_walls_(false)
+    lab_walls_(false),
+    print_(false)
   {
 
     msg_ = new G4GenericMessenger(this, "/Geometry/Next100/",
@@ -56,7 +57,9 @@ namespace nexus {
     msg_->DeclarePropertyWithUnit("specific_vertex", "mm",  specific_vertex_,
       "Set generation vertex.");
 
-    msg_->DeclareProperty("lab_walls", lab_walls_, "Placement of Hall A walls");
+    msg_->DeclareProperty("lab_walls", lab_walls_, "Placement of Hall A walls.");
+
+    msg_->DeclareProperty("print_sipms", print_, "Print SiPM positions.");
 
   // The following methods must be invoked in this particular
   // order since some of them depend on the previous ones
@@ -181,11 +184,28 @@ namespace nexus {
                         "LEAD_BOX", lab_logic_, false, 0);
     }
 
+    if (print_) {
+      std::vector<G4ThreeVector> sipm_pos = inner_elements_->GetSiPMPosInGas();
+      G4int n_sipm = 0;
+      G4int b = 1;
+      for (unsigned int i=0; i<sipm_pos.size(); i++) {
+        G4ThreeVector pos = sipm_pos[i] - coord_origin_;;
+        G4int id = 1000 * b + n_sipm;
+        G4cout << "SiPM " << id << ": " << pos.x() << ", "<< pos.y() << G4endl;
+        n_sipm++;
+        if (id % 1000 == 63) {
+          n_sipm = 0;
+          b++;
+        }
+      }
+    }
+
     //// VERTEX GENERATORS
     lab_gen_ =
       new BoxPointSampler(lab_size_ - 1.*m, lab_size_ - 1.*m,
                           lab_size_  - 1.*m, 1.*m,
                           G4ThreeVector(0., 0., 0.), 0);
+
   }
 
 
--- a/source/geometries/Next100.cc
+++ b/source/geometries/Next100.cc
@@ -136,13 +136,11 @@ namespace nexus {
       vessel_->GetInternalPhysicalVolume();
     G4ThreeVector vessel_displacement =
       shielding_->GetAirDisplacement(); // explained below
-    displ_from_origin_ =
-      G4ThreeVector(-fc_displ_x_, -fc_displ_y_, -vessel_->GetGateZpos());
+
+    coord_origin_ = G4ThreeVector(fc_displ_x_, fc_displ_y_, vessel_->GetGateZpos());
 
     // SHIELDING
-    shielding_->SetCoordOrigin(G4ThreeVector(fc_displ_x_,
-                                             fc_displ_y_,
-                                             vessel_->GetGateZpos()));
+    shielding_->SetCoordOrigin(coord_origin_);
     shielding_->Construct();
     G4LogicalVolume* shielding_logic     = shielding_->GetLogicalVolume();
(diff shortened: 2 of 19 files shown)


source/geometries/GeometryBase.h
```diff
@@ -40,15 +40,18 @@ namespace nexus {
     /// Returns the 3 dimensions of the geometry (x, y, z)
     G4ThreeVector GetDimensions();
 
-    /// Returns true if geometry has a drift field
-    G4bool GetDrift() const;
```
MiryamMV, line 50: Is this related to the field cage displacement?

source/geometries/Next100.cc
```diff
@@ -43,6 +43,8 @@ namespace nexus {
     gate_tracking_plane_distance_(25. * mm + grid_thickness_), // Jordi = 1.5 (distance TP plate-anode ring) + 13.5 (anode ring thickness) + 10 (EL gap)
     gate_sapphire_wdw_distance_  (1458.8 * mm - grid_thickness_), // Jordi
     ics_ep_lip_width_ (55. * mm), // length of the step cut out in the ICS, in the EP side
+    fc_displ_x_ (-3.7 * mm), // displacement of the field cage volumes from 0
+    fc_displ_y_ (-6.4 * mm), // displacement of the field cage volumes from 0
```
MiryamMV, line 47: How have these displacements been determined? We're they measured in the lab?

source/geometries/Next100.cc
```diff
@@ -130,49 +136,49 @@ namespace nexus {
...
       new G4PVPlacement(0, castle_pos, shielding_logic,
                         "LEAD_BOX", hallA_logic_, false, 0);
-      new G4PVPlacement(0, gate_pos - castle_pos, hallA_logic_,
+      new G4PVPlacement(0, -coord_origin_ - castle_pos, hallA_logic_,
                         "Hall_A", lab_logic_, false, 0, false);
     }
     else {
-      new G4PVPlacement(0, gate_pos, shielding_logic, "LEAD_BOX", lab_logic_,
-                        false, 0);
+      new G4PVPlacement(0, -coord_origin_, shielding_logic,
```
MiryamMV, line 183: This is me not understanding the the geometry, why do we add a minus sign to `coord_origin_`?

source/geometries/Next100.cc
```diff
@@ -265,8 +271,7 @@ namespace nexus {
 		  "Unknown vertex generation region!");
     }
 
-    G4ThreeVector displacement = G4ThreeVector(0., 0., -gate_zpos_in_vessel_);
-    vertex = vertex + displacement;
+    vertex = vertex - coord_origin_;
```
MiryamMV, line 294: Same here. Why a minus and not a plus?

source/geometries/Next100FieldCage.cc
```diff
@@ -72,8 +72,8 @@ Next100FieldCage::Next100FieldCage(G4double grid_thickn):
   // cathode_sapphire_wdw_dist_(257.8 * mm),
 
   // external to teflon (hdpe + rings + holders)
-  hdpe_tube_int_diam_ (1093. * mm),
-  hdpe_tube_ext_diam_ (1105.4 * mm),
+  hdpe_tube_int_diam_ (1093. * mm - 11. * mm), // we keep the thickness, but we make it smaller to be closer to the staves
+  hdpe_tube_ext_diam_ (1105.4 * mm - 11. * mm),
```
MiryamMV: Is this 11 derived from the values of the displacement?

source/geometries/Next100Vessel.h
```diff
@@ -96,6 +101,14 @@ namespace nexus {
     gate_tp_distance_ = distance;
   }
 
+  inline void Next100Vessel::SetGateZpos(G4double pos){
+    gate_z_pos_ = pos;
+  }
+
+  inline G4double Next100Vessel::GetGateZpos(){
```
MiryamMV, line 108: This for sure has a reason but I can't get my head around it. Why do we define this here and not in the Next100FieldCage class, for example.

source/geometries/Next100EnergyPlane.cc
```diff
@@ -207,7 +207,8 @@ namespace nexus {
     G4double stand_out_length =
       sapphire_window_thickn_ + tpb_thickn_ + optical_pad_thickn_ + pmt_stand_out_;
 
-    copper_plate_posz_ = GetELzCoord() + gate_sapphire_wdw_dist_ + stand_out_length + copper_plate_thickn_/2.;
+    copper_plate_posz_ = GetCoordOrigin()[2]
```
MiryamMV: In other clases (like Next100FieldCage.cc L216-224) you use `.z()` to extract the z component of the G4ThreeVector. Is it possible to do the same here?

source/geometries/Next100FieldCage.cc
```diff
@@ -330,9 +333,16 @@ void Next100FieldCage::BuildActive()
   drift_region->AddRootLogicalVolume(active_logic);
 
   /// Vertex generator
-  active_gen_ = new CylinderPointSampler2020(0., active_ext_radius_, active_length_/2.,
+  active_gen_radius_ =
+    std::max(gate_int_diam_/2. - GetCoordOrigin().x(),
+             gate_int_diam_/2. - GetCoordOrigin().y());
+
+  active_gen_radius_ = std::max(active_gen_radius_, active_ext_radius_);
```
MiryamMV, line 340: I don't understand. Won't the external radius always be greater than the internal radius?

source/geometries/Next100FieldCage.cc
```diff
@@ -516,7 +534,7 @@ void Next100FieldCage::BuildELRegion()
   if (elfield_) {
     /// Define EL electric field
     UniformElectricDriftField* el_field = new UniformElectricDriftField();
-    G4double global_el_gap_zpos = el_gap_zpos_ - GetELzCoord();
+    G4double global_el_gap_zpos = el_gap_zpos_ - GetCoordOrigin()[2];
```
MiryamMV: Same as before. I personally find it more intuitive.

source/geometries/Next100FieldCage.cc
```diff
@@ -757,17 +786,20 @@ void Next100FieldCage::BuildFieldCage()
...
   }
 
   // Ring vertex generator
   G4double ring_gen_lenght = first_ring_buff_z_pos + ring_thickn_/2. - (posz - ring_thickn_/2.);
   G4double ring_gen_zpos = first_ring_buff_z_pos + ring_thickn_/2. - ring_gen_lenght/2.;
-  ring_gen_ = new CylinderPointSampler2020(ring_int_diam_/2., ring_ext_diam_/2., ring_gen_lenght/2.,
-                                           0., twopi, nullptr,
-                                           G4ThreeVector(0., 0., ring_gen_zpos));
+  ring_gen_ =
+    new CylinderPointSampler2020(ring_int_diam_/2., ring_ext_diam_/2., ring_gen_lenght/2.,
```
MiryamMV: This is not related to the PR but I just realized that `ring_gen_lenght` has a typo. And it might be spreaded throughout the whole code.

source/geometries/GeometryBase.h
paolafer: This was an obsolete function, which was not used practically in the code. Since this PR adds a displacement in the 3 dimensions (instead of only in `z`), I took the chance of restructuring the code using a generic 3D displacement. This was something I had already started in a previous PR, which I have closed now, since it is being implemented here.

source/geometries/Next100.cc
paolafer: I have used the measurements of the distances between the ICS bars and the teflon panels as a reference. These distances can be found in the picture [here](https://github.com/next-exp/nexus/wiki/The-NEXT-100-geometry). The `fc_displ_x_` and `fc_displ_y_` variables have been adjusted to get as closest to the real measurements as possible, without producing overlaps in the geometry. The result is not perfect, but close enough.

source/geometries/Next100.cc
paolafer: Because if the field cage is displaced by a quantity from the origin, you want to place the whole volume that contains everything in `-` that quantity, to have the field cage in `(0, 0, 0)` as a result.

source/geometries/Next100.cc
paolafer: Same as above. In the code there was a `+` before simply because the `displacement` variable was already defined with a minus sign.

source/geometries/Next100FieldCage.cc
paolafer: No, `11 mm` is the distance we need to reduce the diameter of, to get the hdpe tube basically touching the teflon panels. In the picture taken in the lab, the two objects look like touching each other, while in the technical drawings it is not the case. I wanted to keep track of the measurements of the technical drawings, therefore I did keep the numbers in the code and simply subtract a fixed, _ad hoc_, quantity.

source/geometries/Next100Vessel.h
paolafer: Because the position of the gate with respect to the center of the gas (which is used everywhere else in the code) is calculated in the `Next100Vessel` class, therefore we need a function to get it from that class only.

source/geometries/Next100EnergyPlane.cc
paolafer: Yes, I must have missed it. Done!

source/geometries/Next100FieldCage.cc
paolafer: Done.

source/geometries/Next100FieldCage.cc
paolafer: Fixed!

source/geometries/Next100FieldCage.cc
paolafer: The thing is that the active volume is made of the union of two pieces, a polyhedron and a cylinder. On top of that, these pieces are not glued centered on the same axis, but displaced due to the field cage displacement. Therefore, depending on the value of the displacement, it can be that one needs a bigger radius for the generator, than the external radius of the polyhedron. Notice that the generator is a cylinder that needs to overcover all the active volume, to keep the code simple.

source/geometries/Next100.cc
MiryamMV: I understand now.

source/geometries/Next100FieldCage.cc
MiryamMV: I don't get it. The teflon panels shouldn't touch the HDPE tube because they have the holder/staves in between.

source/geometries/Next100OpticalGeometry.cc
```diff
@@ -34,11 +34,13 @@ namespace nexus {
     // common variables used in geometry components
     grid_thickness_ (0.1 * mm),
     gate_tracking_plane_distance_(25. * mm + grid_thickness_),
-    gate_sapphire_wdw_distance_  ((1458.8 - 1.3) * mm - grid_thickness_),
+    gate_sapphire_wdw_distance_  (1458.8 * mm - grid_thickness_),
```
MiryamMV, line 37: Why were the 1.3mm removed and what was the purpose of it initially?

source/geometries/Next100OpticalGeometry.cc
```diff
@@ -126,19 +128,19 @@ namespace nexus {
                 "Unknown kind of gas, valid options are: naturalXe, enrichedXe, depletedXe.");
   }
 
-  G4double gas_size = lab_size - 1.*cm;
+  G4double gas_size = lab_size - 10.*cm;
```
MiryamMV, line 131: Was this increased to avoid possible overlaps?

source/geometries/Next100TrackingPlane.cc
```diff
@@ -84,7 +84,7 @@ void Next100TrackingPlane::Construct()
   G4LogicalVolume* copper_plate_logic =
     new G4LogicalVolume(copper_plate_solid, copper, copper_plate_name);
 
-  G4double copper_plate_zpos = GetELzCoord() - gate_tp_dist_ - copper_plate_thickness_/2.;
+  G4double copper_plate_zpos = GetCoordOrigin()[2] - gate_tp_dist_ - copper_plate_thickness_/2.;
```
MiryamMV: Same as in previous comments.

source/geometries/Next100TrackingPlane.cc
```diff
@@ -100,7 +100,7 @@ void Next100TrackingPlane::Construct()
   sipm_board_geom_->Construct();
   G4LogicalVolume* sipm_board_logic = sipm_board_geom_->GetLogicalVolume();
 
-  G4double zpos = GetELzCoord() - gate_tp_dist_ + sipm_board_geom_->GetThickness()/2.;
+  G4double zpos = GetCoordOrigin()[2] - gate_tp_dist_ + sipm_board_geom_->GetThickness()/2.;
```
MiryamMV: And same here.

source/geometries/Next100FieldCage.cc
paolafer: Sorry, I meant the staves, not the teflon panels.

source/geometries/Next100FieldCage.cc
MiryamMV: Thanks!

source/geometries/Next100FieldCage.cc
MiryamMV: In this STEP file it seems like they're touching: 
[https://viewer.autodesk.com/id/dXJuOmFkc2sub2JqZWN0czpvcy5vYmplY3Q6YTM2MHZpZXdlci1wcm90ZWN0ZWQvdDE3MDE4NjM3NjhfNTE1MDE0ZGMtMzA1OC00ZWUwLWFjNTQtZDZmZTc2NDllODQ0LnN0ZXA?sheetId=ZTgzMzNkMjUtZDhhOC00NmFjLThlMTYtMWYzODdkMGU5NDZi]

source/geometries/Next100OpticalGeometry.cc
paolafer: It was just a fine tuning based on the technical drawings. After the whole review, it turned out that the correct distance was 1458.8 mm.

source/geometries/Next100OpticalGeometry.cc
paolafer: Yes, after the displacement of the FC, `1 cm` was too small, and the displacement produced overlaps.

source/geometries/Next100FieldCage.cc
paolafer: There is an incongruence somewhere, because the 2D drawings say that the internal diameter is 1093. mm. I'll have a look!

source/geometries/Next100TrackingPlane.cc
paolafer: Done.

source/geometries/Next100TrackingPlane.cc
paolafer: Done.

source/geometries/Next100FieldCage.cc
paolafer: Here you can find the technical drawing of hdpe, slide 4: 
[NEXT-100_111522_HDPE_and_staves_PLANOS FABRICACIÓN.pdf](https://github.com/next-exp/nexus/files/13635208/NEXT-100_111522_HDPE_and_staves_PLANOS.FABRICACION.pdf)

As you can see, the internal diameter is 1093 mm. However, with that diameter hdpe is not touching the staves, provided that the active diameter is 984 mm (which has not been measured directly and I don't know what the techincal drawings say).

paolafer: If we leave the hdpe diameter as it is in the drawings, the distance would be like in the picture:
Screenshot 2023-12-11 at 14 49 05

Take into account that hdpe is a cylinder, although in the picture it is shown as a polyhedron.

paolafer: Shall we keep it like this?

paolafer: I realized that the dimensions of the hdpe were those of the original external wrap. I've changed them to those of the internal one, since they are almost touching the staves. I've also updated the distance between two opposite teflon panels, according to the drawings. We know from in-situ measurements that the field cage is not a perfect cylinder, but it is slightly deformed. However, it is impossible to simulate that correctly, therefore we decided to keep the simulation simple.

source/geometries/Next100FieldCage.cc
```diff
@@ -72,8 +72,8 @@ Next100FieldCage::Next100FieldCage(G4double grid_thickn):
   // cathode_sapphire_wdw_dist_(257.8 * mm),
 
   // external to teflon (hdpe + rings + holders)
-  hdpe_tube_int_diam_ (1093. * mm - 11. * mm), // we keep the thickness, but we make it smaller to be closer to the staves
-  hdpe_tube_ext_diam_ (1105.4 * mm - 11. * mm),
+  hdpe_tube_int_diam_ (1081. * mm),
```
MiryamMV, line 75: Maybe it's irrelevant, but the thickness before  and after is not exactly the same.

paolafer: No, it's not the same. It turns out that the external hdpe cylinder is 6 mm-thick, while the internal one is 6.5 mm-thick, according to the drawing attached above.
Since my understanding is that the internal one has been kept, I think that now it is done correctly.

MiryamMV: Great!

paolafer:
Hi @MiryamMV, @jmalbos, what's the status of your review?

MiryamMV (approved):
This PR implements the changes necessary to simulate the field cage displacement successfully. I approve it. Good job @paolafer !

PR #223, opened 2023-11-16
Add memory header

paolafer:
This header is necessary in some compilers to include reference to smart pointers. Some compilers issue an error if it is missing.


jmbenlloch (approved):
This PR fixes an issue with some compilers required at our clusters.

PR #224, opened 2023-11-17
Add Th calibration source

paolafer:
This PR adds the Th-228 calibration source purchased for NEXT-100. It implements the geometry of the encapsulation, following the [technical drawing](https://github.com/next-exp/nexus/files/13391518/Th228_source.pdf).
The source is screwed to a plastic stick, which is not simulated; however, the hole in the capsule, which hosts the stick, is simulated.
The diameter of the active spot, as well as its distance from the top of the capsule are not measured, therefore they have been estimated by eye.

The possibility of using the old source employed in NEXT-White has also been added. The two sources cannot be simulated at the same time.

In order not to further complicate an already complex class, when either source is placed in the geometry (via a configuration parameter), it is placed in all four ports. The vertex generation is independent of the placement of the actual volume, therefore it is not affected by this approximation.

### DIFF
--- a/macros/calibration/NEW_external_source.init.mac
+++ b/macros/calibration/NEW_external_source.init.mac
@@ -1,0 +1,28 @@
+## ----------------------------------------------------------------------------
+## nexus | NEW_disk_source.init.mac
+##
+## Initialization macro to simulate the activity of a specific calibration
+## source in the NEXT-WHITE detector. The source is placed outside the
+## axial or the lateral port, with lead collimator and external scintillator.
+##
+## The NEXT Collaboration
+## ----------------------------------------------------------------------------
+
+### PHYSICS
+/PhysicsList/RegisterPhysics G4EmStandardPhysics_option4
+/PhysicsList/RegisterPhysics G4DecayPhysics
+/PhysicsList/RegisterPhysics G4RadioactiveDecayPhysics
+/PhysicsList/RegisterPhysics NexusPhysics
+/PhysicsList/RegisterPhysics G4StepLimiterPhysics
+
+/nexus/RegisterGeometry NextNew
+
+/nexus/RegisterGenerator IonGenerator
+
+/nexus/RegisterPersistencyManager PersistencyManager
+
+/nexus/RegisterTrackingAction DefaultTrackingAction
+/nexus/RegisterEventAction DefaultEventAction
+/nexus/RegisterRunAction DefaultRunAction
+
+/nexus/RegisterMacro macros/calibration/NEW_external_source.config.mac
--- a/macros/calibration/NEXT100_cal.config.mac
+++ b/macros/calibration/NEXT100_cal.config.mac
@@ -15,6 +15,7 @@
 /process/em/verbose 0
 
 ##### GEOMETRY #####
+/Geometry/Next100/pressure 10 bar
 /Geometry/Next100/elfield false
 /Geometry/Next100/max_step_size 1. mm
 /Geometry/Next100/th_source next100 #next_white
@@ -22,7 +23,7 @@
 ##### GENERATOR #####
 /Generator/IonGenerator/atomic_number 81
 /Generator/IonGenerator/mass_number 208
-/Generator/IonGenerator/region PORT_2b
+/Generator/IonGenerator/region PORT_2b # PORT_2a, PORT_1a, PORT_1b
 
 ##### PHYSICS #####
 ## No full simulation
--- a/macros/calibration/NEXT100_cal.config.mac
+++ b/macros/calibration/NEXT100_cal.config.mac
@@ -1,0 +1,34 @@
+## ----------------------------------------------------------------------------
+## nexus | NEXT100_cal.config.mac
+##
+## Configuration macro to simulate decays of the Th calibration source
+## from one of the vessel ports in the NEXT-100 detector.
+##
+## The NEXT Collaboration
+## ----------------------------------------------------------------------------
+
+##### VERBOSITY #####
+/run/verbose 0
+/event/verbose 0
+/tracking/verbose 0
+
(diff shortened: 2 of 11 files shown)


paolafer:
Hi @bpalmeiro, what's the status of this review?

macros/calibration/NEW_external_source.init.mac
bpalmeiro, line 1: Sorry if I'm missing something, but what is the point of adding/modifying a macro for NEW for something that has been acquired this year?

macros/calibration/NEXT100_cal.config.mac
```diff
@@ -0,0 +1,34 @@
...
+/run/verbose 0
+/event/verbose 0
+/tracking/verbose 0
+
+/process/em/verbose 0
+
+##### GEOMETRY #####
+/Geometry/Next100/elfield false
+/Geometry/Next100/max_step_size 1. mm
+/Geometry/Next100/th_source next100 #next_white
```
bpalmeiro: why this comment? The macro is explicitly about N100

macros/calibration/NEXT100_cal.config.mac
```diff
@@ -0,0 +1,34 @@
...
+## ----------------------------------------------------------------------------
+
+##### VERBOSITY #####
+/run/verbose 0
+/event/verbose 0
+/tracking/verbose 0
+
+/process/em/verbose 0
+
+##### GEOMETRY #####
```
bpalmeiro, line 17: Should we explicitly say the pressure? Just to avoid problems with default values.

source/geometries/Next100Vessel.cc
```diff
@@ -472,15 +547,17 @@ namespace nexus {
...
 
     else if (region == "PORT_1a"){
-      vertex = port_gen_->GenerateVertex(VOLUME);
-
+      if (th_source_ == "next100") {
+        vertex = th_port_gen_->GenerateVertex(VOLUME);
+      } else if (th_source_ == "next_white") {
+        vertex = th_white_port_gen_->GenerateVertex(VOLUME);
+      }
       vertex = vertex.rotateX( 90. * deg);
```
bpalmeiro, line 561: If I understand correctly these lines, for each port, it defines a vertex based on the detector and then adapts it and places it. However, this makes some NEW ports to be placed very far from the active volume. Am I missing anything?

macros/calibration/NEXT100_cal.config.mac
```diff
@@ -0,0 +1,34 @@
...
+/Generator/IonGenerator/region PORT_2b
+
+##### PHYSICS #####
+## No full simulation
+/PhysicsList/Nexus/clustering          false
+/PhysicsList/Nexus/drift               false
+/PhysicsList/Nexus/electroluminescence false
+
+##### PERSISTENCY #####
+/nexus/persistency/output_file Next100_th_cal
```
bpalmeiro: Maybe, here, it's helpful (since this is an example) to add a comment with the ports available. No strong opinions.

bpalmeiro (commented):
The PR is in a very mature state, almost ready for revision. I left a couple of questions, I'm sure prompted mainly by my ignorance of the code and conventions.

macros/calibration/NEW_external_source.init.mac
paolafer: The modification of the NEW macros was prompted by the fact that I realized that those macros weren't run automatically with the tests because the `init` macros were missing. Without these macros, I didn't see that the new code was breaking the old one and this is why I decided to add them in this PR, although they are related to NEW.

macros/calibration/NEXT100_cal.config.mac
paolafer: I understand that this can be confusing, but the `next100`/`next_white` parameter refers to the kind of source to be used. Both of them will be used in NEXT100, and I just call the old one "the NEXT-White source" and the new one "the NEXT100 source". If you have any suggestions to make this clearer, please, tell me.

macros/calibration/NEXT100_cal.config.mac
paolafer: Done.

macros/calibration/NEXT100_cal.config.mac
paolafer: Done.

source/geometries/Next100Vessel.cc
paolafer: This code is only used in NEXT100 (it is in the NEXT100Vessel class). I think that the misunderstanding comes from the `next_white` name of the source. As I mentioned in another comment, if you can see another way of identifying the two sources, please, let me know!

macros/calibration/NEW_external_source.init.mac
bpalmeiro: Fine with me then!

macros/calibration/NEXT100_cal.config.mac
bpalmeiro: Ok, gotcha. It's clear once you know... but I agree it's prompt confusion at first glance. Not having any brilliant advice myself, I think we should avoid the NEW/N100 naming. Maybe by the year or something of that sort?

paolafer: I tried with old/new, is that better?

bpalmeiro: I believe so. It can go like that for now. My only concern is that it is a temporal patch since we may acquire more sources in the future, and this convention runs short. But, again, it's fine with me for the scope of this PR.

bpalmeiro (approved):
The PR is clear and adds the recently acquired source for calibration. In addition, fixes some problems with NEW tests and provides a cleaner macro for the vessel scripts. Good work!

PR #225, opened 2023-11-23
Add possibility of getting more precision in verbosity

paolafer:
This PR allows the user to get the desired precision in the distances appearing in the output when tracking verbosity is used. 
It is particularly useful in the debugging of new geometries.


jwaiton (approved):
This PR adds the ability for users to set tracking precision. This is done by an additional argument entered at the CLI:

`/bin/nexus -n 1 -p 5 example.init.mac`

Where the `-p` flag is for precision, and 5 is the precision defined in significant figures.

Approved! Good work @paolafer :)

Issue #142, opened 2022-02-28
Review of the temperature-dependent density code


paolafer:
To be able to use the code that calculates the gas density as a function of pressure and temperature, further review is required.

paolafer:
This code needs to be reviewed completely now, following PR #242.

PR #227, opened 2023-11-27
Add epoxy optical properties

paolafer:
This PR adds new optical properties to be used in the epoxy resine in front of the NEXT-100 SiPMs. This is the best we can do with the information we have. 
The datasheet of the SiPM model can be found here:
[K30-I50426_specification sheet(S13372-1350TE).pdf](https://github.com/next-exp/nexus/files/13479087/K30-I50426_specification.sheet.S13372-1350TE.pdf)

I have also reorganized the code to try to reduce its length and to optimize it.

### DIFF
--- a/source/materials/OpticalMaterialProperties.cc
+++ b/source/materials/OpticalMaterialProperties.cc
@@ -598,7 +598,7 @@ namespace opticalprops {
 
   /// Gaseous xenon ///
   G4MaterialPropertiesTable* GXe(G4double pressure,
-                                G4double temperature,
+                                 G4double /*temperature*/,
                                 G4int    sc_yield,
                                 G4double e_lifetime)
   {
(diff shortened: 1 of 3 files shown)


source/materials/OpticalMaterialProperties.cc
```diff
@@ -535,7 +596,7 @@ namespace opticalprops {
 
 
 
-  /// Gaseous Xenon ///
+  /// Gaseous xenon ///
   G4MaterialPropertiesTable* GXe(G4double pressure,
                                 G4double temperature,
```
soleti: This gives an unused parameter warning, do we want to keep it or maybe add a dummy function as a placeholder?

paolafer: I know, we have an open issue (#142) about the review of the temperature-dependent code which hasn't been done yet. In the meanwhile, we can comment the parameter, maybe?

source/materials/OpticalMaterialProperties.cc
```diff
@@ -18,7 +18,7 @@ using namespace nexus;
 using namespace CLHEP;
```
soleti, line 17: It doesn't let me add a comment above, but it looks like we don't need `assert.h` on line 15. I tried to comment it out and it compiles and runs fine, can you double check?

paolafer: Correct! I eliminated it.

soleti (approved):
Looks good to me, nice refactoring 👍

PR #233, opened 2024-01-23
Remove `sipm_pos` from if-loop.

MiryamMV:
This PR defines `sipm_pos` outside the `verbosity_` if-loop. This corrects all the overlaps in this volume encountered when running the overlap checks.


paolafer (approved):
The overlap is fixed, I approve it!

PR #235, opened 2024-01-24
Fix vertices

paolafer:
This PR removes a vertex which did not behave as intended (and it was not used) and fixes the vertex placed in the center of the NEXT100 geometry, which is now displaced from xy = (0, 0).
Finally, a region that was not used anymore has been removed.


MiryamMV (approved):
Everything seems correct. I approve.

PR #236, opened 2024-01-24
Geant4-v11.2

paolafer:
This PR modifies the code to work with Geant4-v11.2.
Beyond the removal of an unnecessary forward declaration, an explicit dependence on Qt is added in the `SConstruct` file, due to a bug in Geant4 (https://bugzilla-geant4.kek.jp/show_bug.cgi?id=2589). 
I'll mark this PR as a draft, because if they solve the bug before we decide to start using Geant4-v11.2, there will be no need of this change.
The v11.2.1 patch solves the issue, therefore this explicit dependence is removed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.cern.ch/next/nexus-base:2024-01-25
+FROM registry.cern.ch/next/nexus-base:2024-02-20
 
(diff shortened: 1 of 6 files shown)


jmbenlloch (approved):
This PR updates nexus to version 11.2.1

PR #237, opened 2024-01-25
Add Qt to docker image used in github action

jmbenlloch:
This PR modifies the docker container used in the Github action to include Qt as a dependency and uses a Geant4 compilation with Qt enabled.

### DIFF
--- a/docker/geant4/Dockerfile
+++ b/docker/geant4/Dockerfile
@@ -3,7 +3,7 @@ FROM ubuntu:jammy
 ENV G4_VERSION v11.0.2
 
 RUN apt-get update -y && \
-    apt-get install -y cmake curl g++ libexpat1-dev
+    apt-get install -y cmake curl g++ libexpat1-dev libqt5core5a qtbase5-dev
 RUN mkdir -p /opt/geant4/src && \
     mkdir -p /opt/geant4/build && \
     mkdir -p /opt/geant4/install && \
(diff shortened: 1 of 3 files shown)


docker/geant4/Dockerfile
```diff
@@ -14,6 +14,7 @@ RUN mkdir -p /opt/geant4/src && \
     cmake -DCMAKE_INSTALL_PREFIX=/opt/geant4/install \
           -DGEANT4_BUILD_MULTITHREADED=ON \
           -DGEANT4_INSTALL_EXAMPLES=OFF \
+          -DGEANT4_USE_QT=ON \
```
paolafer, line 17: According to the G4 instructions, one should also add the `-DCMAKE_PREFIX_PATH=/path/to/Qt/installation` flag. Does it work without it?

jmbenlloch: Since Qt was in a standard path, cmake was able to find it automatically. In any, case, I have check it running the docker container and, as you can see, it has Qt enabled and the proper paths were included:

```
root@6554bee89f16:/opt/geant4/install/bin# geant4-config  --cflags
-DG4VIS_USE_OPENGL -DG4UI_USE_TCSH -DG4UI_USE_QT -DG4VIS_USE_OPENGLQT -I/usr/include/x86_64-linux-gnu/qt5/ -I/usr/include/x86_64-linux-gnu/qt5/QtCore -I/usr/lib/x86_64-linux-gnu/qt5//mkspecs/linux-g++ -I/usr/include/x86_64-linux-gnu/qt5/QtGui -I/usr/include/x86_64-linux-gnu/qt5/QtWidgets -I/usr/include/x86_64-linux-gnu/qt5/QtOpenGL -I/usr/include/x86_64-linux-gnu/qt5/QtPrintSupport -W -Wall -pedantic -Wno-non-virtual-dtor -Wno-long-long -Wwrite-strings -Wpointer-arith -Woverloaded-virtual -Wno-variadic-macros -Wshadow -pipe -pthread -ftls-model=initial-exec -std=c++17 -I/opt/geant4/install/bin/../include/Geant4
```

paolafer (approved):
This PR allows us to check any compilation errors due to Qt automatically; it can be useful, for instance, when we change the G4 version.
Approved!

PR #239, opened 2024-01-30
Clean up samplers

paolafer:
This PR addresses some points of issue #2, with a review of the samplers used in vertex generation.

- The `MuonsPointSampler` class is deleted, since the same sampling can be achieved with a box sampler.
- The class used currently for the cylindric sampler is renamed to `CylinderPointSampler` and the old one to `CylinderPointSamplerLegacy`.
- A new class for the box sampler is introduced, similar to the new one for the cylinder, following the G4 constructor of `G4Box` and with the possibility of passing a physical volume.
- New tests are added.
- `Enum` are used for regions, instead of `strings`.
- The `SpherePointSampler` class is also modified to adapt it to the G4 constructor.


source/utils/BoxPointSampler.h
```diff
@@ -9,28 +9,32 @@
...
+namespace nexus {
+  
   class BoxPointSampler
   {
   public:
     /// Constructor
-    BoxPointSampler(G4double inner_x, G4double inner_y, G4double inner_z,
-                    G4double thickness,
+    BoxPointSampler(G4double half_inner_x, G4double half_inner_y,
+                    G4double half_inner_z, G4double thickness=0,
```
atrettin, line 27: Just out of interest: Why is this "half" everywhere now? I see many factors of two or half have been added in the code to get the same result.

paolafer: The guide line here is to mimic the constructor of the correspondent solid volumes, in this case, `G4Box` (see, for instance, https://geant4.kek.jp/lxr/source/geometry/solids/CSG/include/G4Box.hh#L59).

atrettin: I see. I'm not a huge fan of having all these divisions by two everywhere in the code now, but I suppose that makes sense if we want to match our convention to that of G4. Thank you for the explanation.

source/geometries/Next1EL.cc
```diff
@@ -1371,7 +1373,7 @@ G4ThreeVector Next1EL::GenerateVertex(const G4String& region) const
       }
   } else if (region == "MUONS") {
     //generate muons sampling the plane
-    vertex = muons_sampling_->GenerateVertex();
+    vertex = muons_sampling_->GenerateVertex("INSIDE");
```
atrettin, line 1376: Why doesn't this use an enum?

atrettin: Nevermind, I see that this is calling the legacy sampler where it is indeed a string.

atrettin (approved):
I have compiled and run `nexus` with these changes and looked at all the files. The re-naming of the classes and change of conventions have been done consistently as far as I can see.

PR #240, opened 2024-02-05
Assign attachment to electric field

paolafer:
With this PR we change the way we simulate the electron lifetime in our gas (a.k.a. `attachment`). 
In the current code, the value is read by the `ATTACHMENT` optical property of the material where the ionization electrons start their drift. However, this is a problem if we simulate the grids as made of steel instead of as made of a thin dielectric gas with the same properties as xenon, as done so far.
One way of solving this issue is to assign the lifetime property directly to the electric field; it is still read from the optical property of the xenon gas, but now it is decoupled from the specific material the grids are made of.

I also changed the term `attachment` to `lifetime` in the code, because I think it describes better what it represents.

### DIFF
--- a/source/physics/UniformElectricDriftField.cc
+++ b/source/physics/UniformElectricDriftField.cc
@@ -64,7 +64,6 @@ namespace nexus {
 
     G4ThreeVector position;
     G4double time;
-    G4double time_diff;
 
     for (G4int i=0; i<3; i++) {
       if (i != axis_)  {     // Transverse coordinate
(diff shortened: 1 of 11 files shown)


source/physics/UniformElectricDriftField.cc
```diff
@@ -84,23 +87,18 @@ namespace nexus {
     // Set the new time and position of the drifting charge
     xyzt.set(time, position);
 
+    G4double rnd = -lifetime_ * log(G4UniformRand());
```
kvjmistry, line 90: Do we need a catch here in case of log(0)?

paolafer: It seems that `0` is excluded, see, for instance, https://geant4-forum.web.cern.ch/t/what-is-the-range-of-numbers-generated-by-g4uniformrand/5187/2.

source/physics/UniformElectricDriftField.cc
```diff
@@ -84,23 +87,18 @@ namespace nexus {
...
+    if (xyzt.t() > rnd) step_length = 0.;
```
paolafer: I'm not sure about this line. I've run two small Kr productions with a short lifetime with the current code and with these changes and they don't match (the current code gives the correct value of the lifetime, but this one doesn't). I have to figure that out before approving the PR.

paolafer: Solved.

kvjmistry (approved):
I think this all looks good to me!

PR #241, opened 2024-02-15
Register slow mark and simplify folder structure

paolafer:
The `slow` mark is registered to eliminate a warning appearing during the run of automatic tests.
The full `pytest` folder has been extracted from the `tests` folder, since no other tests are foreseen in the near future (and the Catch2 tests live elsewhere already, since they need to be compiled).


gonzaponte (approved):
Clear and straightforward!

PR #242, opened 2024-02-19
Fix Xe density

enakshidey:
In the Material.cc file the calculation of the Xe density did not take into account the isotopic composition of Xe for different Xe types (e.g depleted, enriched, natural). The density was only function of pressure and directly taken from the NIST data table. But percentage contribution from different isotopes should also be included for different Xe types. So I have modified the density calculation including the isotopic composition.

### DIFF
--- a/source/materials/MaterialsList.cc
+++ b/source/materials/MaterialsList.cc
@@ -14,6 +14,7 @@
 #include <G4Material.hh>
 #include <G4Element.hh>
 #include <G4NistManager.hh>
+
 #include <vector>
 #include <utility> 
 
@@ -23,7 +24,7 @@ using namespace CLHEP;
 namespace materials {
 
   // Function to calculate gas density based on isotopic composition
-  double CalculateGasDensityFromIsotopicComposition(double pressure, double temperature, const std::vector<std::pair<int, double>>& isotopicComposition) {
+  G4double CalculateGasDensityFromIsotopicComposition(G4double pressure, G4double temperature, const std::vector<std::pair<int, double>>& isotopicComposition) {
     const double R = 8.314; // Ideal gas constant in J/(mol·K)
     double averageMolarMass = 0.0; // in g/mol
 
@@ -38,7 +39,7 @@ namespace materials {
     averageMolarMass /= 1000.0;
 
     // Calculate density using the ideal gas law: ρ = PM / RT
-    double density = (pressure * averageMolarMass) / (R * temperature); // Result in kg/m^3
+    G4double density = (pressure * averageMolarMass) / (R * temperature); // Result in kg/m^3
     return density;
   }
 
@@ -51,7 +52,7 @@ namespace materials {
     G4String name = "GXe";
     G4Material* mat = G4Material::GetMaterial(name, false);
 
-    if (mat == nullptr) {
+    if (mat == 0) {
       G4NistManager* nist = G4NistManager::Instance();
       mat = new G4Material(name, density, 1, kStateGas, temperature, pressure);
       G4Element* Xe = nist->FindOrBuildElement("Xe");
@@ -201,17 +202,17 @@ namespace materials {
   }
 
   G4Material* GXeAr(G4double pressure, G4double temperature, G4double percXe) 
+  {
+    G4String name = "GXeAr";
+
+    G4Material* mat = G4Material::GetMaterial(name, false);
+
     std::vector<std::pair<int, double>> isotopicComposition = {
       {124, 0.0952}, {126, 0.089}, {128, 1.9102}, {129, 26.4006},
       {130, 4.071}, {131, 21.2324}, {132, 26.9086}, {134, 10.4357}, {136, 8.8573}
     };
 
     G4double GXeNatural_density = CalculateGasDensityFromIsotopicComposition(pressure * 1e5, temperature, isotopicComposition);
-    
-  {
-    G4String name = "GXeAr";
-
-    G4Material* mat = G4Material::GetMaterial(name, false);
 
     if (mat == 0) {
 
@@ -253,15 +254,16 @@ namespace materials {
   G4Material* GXeHe(G4double pressure,
           G4double temperature,
           G4double percXe, G4int mass_num)
+    
+  {
+    G4String name = "GXeHe";
+
     std::vector<std::pair<int, double>> isotopicComposition = {
       {129, 0.0656392}, {130, 0.0656392}, {131, 0.234361},
       {132, 0.708251}, {134, 8.6645}, {136, 90.2616}
     };
 
     G4double GXeEnriched_density = CalculateGasDensityFromIsotopicComposition(pressure * 1e5, temperature, isotopicComposition);
-    
-  {
-    G4String name = "GXeHe";
 
     G4Material* mat = G4Material::GetMaterial(name, false);
 
--- a/source/materials/MaterialsList.cc
+++ b/source/materials/MaterialsList.cc
@@ -14,54 +14,65 @@
 #include <G4Material.hh>
 #include <G4Element.hh>
 #include <G4NistManager.hh>
+#include <vector>
+#include <utility> 
 
 using namespace nexus;
 using namespace CLHEP;
 
-
 namespace materials {
-  G4Material* GXe(G4double pressure, G4double temperature)
-  {
-    G4Material* mat = GXe_bydensity(GXeDensity(pressure),
-                                    temperature, pressure);
 
+  // Function to calculate gas density based on isotopic composition
+  double CalculateGasDensityFromIsotopicComposition(double pressure, double temperature, const std::vector<std::pair<int, double>>& isotopicComposition) {
+    const double R = 8.314; // Ideal gas constant in J/(mol·K)
+    double averageMolarMass = 0.0; // in g/mol
+
+    for (const auto& iso : isotopicComposition) {
+      int massNumber = iso.first;
+      double percentage = iso.second; 
+      double molarMass = XenonMassPerMole(massNumber); // This function needs to provide the molar mass for each isotope
+      averageMolarMass += molarMass * percentage;
+    }
+
+    // Convert average molar mass to kg/mol for the density calculation
+    averageMolarMass /= 1000.0;
+
+    // Calculate density using the ideal gas law: ρ = PM / RT
+    double density = (pressure * averageMolarMass) / (R * temperature); // Result in kg/m^3
+    return density;
+  }
+
+  G4Material* GXe(G4double pressure, G4double temperature) {
+    G4Material* mat = GXe_bydensity(GXeDensity(pressure), temperature, pressure);
     return mat;
   }
 
-
-  G4Material* GXe_bydensity(G4double density,
-                                          G4double temperature,
-                                          G4double pressure)
-  {
+  G4Material* GXe_bydensity(G4double density, G4double temperature, G4double pressure) {
     G4String name = "GXe";
-
     G4Material* mat = G4Material::GetMaterial(name, false);
 
-    if (mat == 0) {
+    if (mat == nullptr) {
       G4NistManager* nist = G4NistManager::Instance();
-
-      mat = new G4Material(name, density, 1,
-        kStateGas, temperature, pressure);
-
+      mat = new G4Material(name, density, 1, kStateGas, temperature, pressure);
       G4Element* Xe = nist->FindOrBuildElement("Xe");
-
-      mat->AddElement(Xe,1);
+      mat->AddElement(Xe, 1);
     }
 
     return mat;
   }
 
+  G4Material* GXeEnriched(G4double pressure, G4double temperature) {
+    std::vector<std::pair<int, double>> isotopicComposition = {
+      {129, 0.0656392}, {130, 0.0656392}, {131, 0.234361},
+      {132, 0.708251}, {134, 8.6645}, {136, 90.2616}
+    };
 
-  G4Material* GXeEnriched(G4double pressure, G4double temperature)
-  {
-    G4double gas_density = GXeDensity(pressure);
+    G4double gas_density = CalculateGasDensityFromIsotopicComposition(pressure * 1e5, temperature, isotopicComposition);
     G4Material* mat = GXeEnriched_bydensity(gas_density, temperature, pressure);
-
     return mat;
   }
 
-
-  G4Material* GXeEnriched_bydensity(G4double density,
+   G4Material* GXeEnriched_bydensity(G4double density,
               G4double temperature,
               G4double pressure)
   {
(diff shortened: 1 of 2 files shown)


paolafer:
Hi @enakshidey, is the PR ready for review or do you still need some work to do? I notice that the code as it is isn't compiling.

paolafer:
Also, please, have a look at the NEXUS [coding convention](https://github.com/next-exp/nexus/wiki/Coding-conventions-and-style-guide) and try to follow them in the code style.

enakshidey:
Hi @paolafer I still need to modify few things there. It's not yet ready.

enakshidey:
IsotopicComposition
Previously in the Materials.cc code it was not taking in to account the isotopic masses of various Xe isotopes. So the density calculation was incorrect. I have included the isotopic masses in the density calculation. And in the plot we can see the blob energy does not depend on what gas you put at a particular pressure.

paolafer:
Can you please rebase your branch on the master? For some reason, it doesn't compile on my laptop, although it does run on the github action.

source/materials/MaterialsList.cc
```diff
@@ -15,53 +15,79 @@
...
 
-
 namespace materials {
-  G4Material* GXe(G4double pressure, G4double temperature)
-  {
-    G4Material* mat = GXe_bydensity(GXeDensity(pressure),
-                                    temperature, pressure);
+    // Calculation of Compressibilty factor from https://next.ific.uv.es/DocDB/0011/001183/001/GasDensity_2021-04-23.pdf
+    G4double CompressibilityFactor(G4double pressure, G4double temperature, double R) {
+        double T_r = temperature/289.733;
```
paolafer: You should use the unit with temperature, too.

source/materials/MaterialsList.cc
```diff
@@ -15,53 +15,79 @@
...
 using namespace CLHEP;
 
-
 namespace materials {
-  G4Material* GXe(G4double pressure, G4double temperature)
-  {
-    G4Material* mat = GXe_bydensity(GXeDensity(pressure),
-                                    temperature, pressure);
+    // Calculation of Compressibilty factor from https://next.ific.uv.es/DocDB/0011/001183/001/GasDensity_2021-04-23.pdf
+    G4double CompressibilityFactor(G4double pressure, G4double temperature, double R) {
```
paolafer: R is not used in the function, is it?

source/materials/MaterialsList.cc
```diff
@@ -15,53 +15,79 @@
...
+        double T_r = temperature/289.733;
+        double P_r = pressure/ (58.420*bar);
+        double B0 = 0.083 - (0.422 / pow(T_r, 1.6));       
+        double z = 1 + B0*(P_r/T_r);
+        return z;       
+    }
+
+    // Function to calculate gas density based on Xe isotopic composition
+    G4double CalculateGasDensityFromIsotopicComposition(G4double pressure, G4double temperature, const std::vector<std::pair<int, double>>& isotopicComposition) {
```
paolafer: If these values need to be in some specific units (K, bar...) I would specify all of those, even if they are the basic units of Geant4, for clarity.

source/materials/MaterialsList.cc
```diff
@@ -15,53 +15,79 @@
...
+        double average_molar_mass = 0.0; // in g/mol
+
+        for (const auto& iso : isotopicComposition) {
+            int mass_number = iso.first;
+            double percentage = iso.second/100;
+            double molar_mass = XenonMassPerMole(mass_number) / (g / mole);
+            average_molar_mass += molar_mass * percentage;
+        }
+
+        average_molar_mass /= 1000.0;
```
paolafer: Why is this divided by 1000? I would avoid magic numbers, if possible or I would write a comment on that.

source/materials/MaterialsList.cc
```diff
@@ -98,15 +124,18 @@ namespace materials {
     return mat;
   }
 
-  G4Material* GXeDepleted(G4double pressure, G4double temperature)
-  {
-    G4double gas_density = GXeDensity(pressure);
-    G4Material* mat = GXeDepleted_bydensity(gas_density, temperature, pressure);
+  G4Material* GXeDepleted(G4double pressure, G4double temperature) {
+    std::vector<std::pair<int, double>> isotopicComposition = {
+      {124, 0.102}, {126, 0.201}, {128, 3.065}, {129, 24.900}, {130, 5.361},
```
paolafer: The composition of enriched and depleted xenon is repeated twice in the code. You can define it once and reuse it.

source/materials/MaterialsList.cc
```diff
@@ -121,21 +150,29 @@ namespace materials {
         kStateGas, temperature, pressure);
 
 
-      G4Element* Xe = new G4Element("GXeDepleted", "Xe", 5);
-
+      G4Element* Xe = new G4Element("GXeDepleted", "Xe", 9);
+      
+      G4Isotope* Xe124 = new G4Isotope("Xe129", 54, 124, XenonMassPerMole(124));
```
paolafer: The isotopic number is not the same in the string and in the other arguments.

source/materials/MaterialsList.cc
```diff
@@ -795,6 +845,7 @@ namespace materials {
   }
 
 
+
```
paolafer: If these lines were added by error, please remove them.

source/materials/MaterialsList.cc
```diff
@@ -186,17 +225,24 @@ namespace materials {
     return mat;
   }
 
-  G4Material* GXeAr(G4double pressure, G4double temperature, G4double percXe)
+  G4Material* GXeAr(G4double pressure, G4double temperature, G4double percXe)
```
paolafer, line 203: Remove blank space at the end.

source/materials/MaterialsList.cc
```diff
@@ -15,55 +15,85 @@
...
+            int mass_number = iso.first;
+            double percentage = iso.second/100;
+            double molar_mass = XenonMassPerMole(mass_number) / (g / mole);
+            average_molar_mass += molar_mass * percentage;
+        }
 
+        average_molar_mass /= 1000.0; //convert in kg/mol
+
+        double z = CompressibilityFactor(pressure, temperature);
+        double density = (pressure/hep_pascal * average_molar_mass) / (z * R * temperature); // Result in kg/m^3
```
paolafer: I'm confused. In the comment in line 36 you say the pressure is in bar, and then you divide by pascal. Why is that so?

paolafer:
I'm afraid I'm a bit lost with the units over the code. Does the user need to provide pressure and temperature in specific units or not? Could you add a test to ensure that the calculations are done correctly?

source/materials/MaterialsList.cc
```diff
@@ -15,55 +15,85 @@
...
+    }
+
+  G4Material* GXe(G4double pressure, G4double temperature) {
+    std::vector<std::pair<int, double>> isotopicComposition = {
+      {124, 0.095}, {126, 0.089}, {128, 1.910},
+      {129, 26.401}, {130, 4.071}, {131, 21.232}, 
+      {132, 26.909}, {134,10.436}, {136, 8.857}
+    };
+
+    G4double gas_density = CalculateGasDensityFromIsotopicComposition(pressure , temperature, isotopicComposition);
```
paolafer: Move the comma after `pressure` close to it.

source/materials/MaterialsList.cc
```diff
@@ -15,55 +15,85 @@
...
+    G4Material* mat = GXe_bydensity(gas_density, temperature, pressure);
+    std::cout << "press, temp,den: "<<pressure << ", "<<temperature <<", " <<gas_density << std::endl;
```
paolafer: Please, remove the printout.

source/materials/MaterialsList.cc
```diff
@@ -15,55 +15,85 @@
...
+        double average_molar_mass = 0.0; // in g/mol
+
+        for (const auto& iso : isotopicComposition) {
+            int mass_number = iso.first;
+            double percentage = iso.second/100;
+            double molar_mass = XenonMassPerMole(mass_number) / (g / mole);
+            average_molar_mass += molar_mass * percentage;
+        }
 
+        average_molar_mass /= 1000.0; //convert in kg/mol
```
paolafer: Instead of dividing by 1000 to obtain kg/mol, why don't you directly return the molar mass in kg/mol in line 44 like this:

`double molar_mass = XenonMassPerMole(mass_number) / (kg / mole);`

? Am I missing something?

source/materials/MaterialsList.cc
```diff
@@ -15,55 +15,85 @@
...
 using namespace CLHEP;
 
-
 namespace materials {
-  G4Material* GXe(G4double pressure, G4double temperature)
-  {
-    G4Material* mat = GXe_bydensity(GXeDensity(pressure),
-                                    temperature, pressure);
+    // Calculation of Compressibilty factor from https://next.ific.uv.es/DocDB/0011/001183/001/GasDensity_2021-04-23.pdf
+    G4double CompressibilityFactor(G4double pressure, G4double temperature) {
```
paolafer, line 26: If I understand correctly, it's this formula the one that needs the pressure in bar and the temperature in kelvin, is that right? In that case, I would make clear that the code transform whatever T/P is passed by the user in those units, like this:

`double T_r = (temperature/kelvin)/289.733;`
`double P_r = (pressure/bar)/58.420;`

I think that this way it's clearer.

source/materials/MaterialsList.cc
```diff
@@ -15,55 +15,85 @@
...
+        double T_r = temperature/(289.733*kelvin);
+        double P_r = pressure/ (58.420*bar);
+        double B0 = 0.083 - (0.422 / pow(T_r, 1.6));   //Second virial coefficient    
+        double z = 1 + B0*(P_r/T_r);  //Compressibility factor
+        return z;       
+    }
+
+    // Function to calculate gas density based on Xe isotopic composition
+    G4double CalculateGasDensityFromIsotopicComposition(G4double pressure, G4double temperature,
+     const std::vector<std::pair<int, double>>& isotopicComposition)   //pressure in bar , temperature in K
```
paolafer: Here the units can be whatever, right? They don't need to be bar or kelvin, if they are transformed within the `CompressibilityFactor()` function.

paolafer (changes requested):

source/materials/MaterialsList.cc
gonzaponte: The pressure and temperature in the input will be expressed in the internal geant units. For example, 12345 * torr would work as well as 67.8 * bar. The caller should provide values with units, not assume one.

source/materials/MaterialsList.cc
gonzaponte: Indeed, as explained above, all numbers should come with units.

source/materials/MaterialsList.cc
```diff
@@ -15,55 +15,85 @@
...
+      {
+        const double R = 8.314; // Ideal gas constant in J/(mol·K)
```
gonzaponte: Following the same reasoning,
`R = 8.314 * joule / mole / kelvin`

source/materials/MaterialsList.cc
gonzaponte: I don't see a reason to keep changing the units. whatever mass per mole `XenonMassPerMole` returns should be expressed in the internal geant4 units. If a formula uses a quantity expressed in some specific units, it should be done in the formula, but this is rare.

source/materials/MaterialsList.cc
gonzaponte: Agreed. There are probably an even number of factors cancelling out somewhere.

paolafer: I see that the pressure here must be in pascal, because you're using a value for R in `J/(mol*K)`. For the same reason, the temperature must be provided in Kelvin, is that correct? Since the conversion is done internally in the code, the user can still provide temperature and pressure in whatever units, right? In that case, I would remove the comment in line 36 (since the units can be whatever) and change:

`double density = (pressure/hep_pascal) * average_molar_mass / (z * R * (temperature/kelvin));`

Do you agree or am I missing something?

source/materials/MaterialsList.cc
```diff
@@ -15,55 +15,85 @@
...
+      mat->AddElement(Xe, 1);
     }
 
     return mat;
   }
+  
+  G4Material* GXeEnriched(G4double pressure, G4double temperature) {
+    std::vector<std::pair<int, double>> isotopicComposition = {
+      {129, 0.0656392}, {130, 0.0656392}, {131, 0.234361},
+      {132, 0.708251}, {134, 8.6645}, {136, 90.2616}
```
gonzaponte: These numbers should probably be multiplied by `perCent`

gonzaponte (commented):
My two cents

paolafer:
Hi @enakshidey! I've talked to the other reviewer (@gonzaponte) and we agreed on the following suggestions (in case you found our individual comments confusing - which they are, sorry for that!).

There's no need of changing the units within the functions, as long as all the variables have the appropriate ones. 
Therefore, you just need to define `const double R = 8.314 * joule / (mole*kelvin); ` in the `CalculateGasDensityFromIsotopicComposition()` function and 

`double T_r = temperature/(289.733*kelvin);`
`double P_r = pressure/(58.420*bar);`

in the `CompressibilityFactor()` function.
Once you have that, the density can be just defined as `double density = (pressure * average_molar_mass) / (z * R * temperature);`. 
Do you agree?

Also, we would define directly `isotopicCompisition` as `124, 0.095 * perCent`. This way, you can drop the division by 100 in line 43 and the multiplication by `perCent` whenever you add an isotope to the material.

Notice that the tests are failing with the actual version of the code. I'm not exactly sure why, but the problem seems to be that there's no light reaching both SiPMs and PMTs in the events generated in the tests, which may be caused by a wrong implementation of the density.

enakshidey:
Thank you both @paolafer @gonzaponte  for the inputs. I know why the test failed. I was checking the density value in normal unit kg/m^3 to make sense if everything is correct and forgot to change it in the G4units back. But you are right I have done few unnecessary unit conversion. I will remove those things as you have mentioned and submit the modified code as soon as possible. 

And extremely sorry for all these inconveniences.

source/materials/MaterialsList.cc
kvjmistry: As I understand this the inputs of pressure and temperature are in G4 units (as that is what gets put into the config file). The /pascal is to use pascal units in this calculation.

paolafer: The thing is that you don't need to use pascal units if you use `R` also in Geant4 units.

kvjmistry: Ah I see what you are saying, should be simple to fix this.

source/materials/MaterialsList.cc
```diff
@@ -15,55 +15,83 @@
...
+        double T_r = temperature/(289.733*kelvin);
+        double P_r = pressure/ (58.420*bar);
+        double B0 = 0.083 - (0.422 / pow(T_r, 1.6));   //Second virial coefficient    
+        double z = 1 + B0*(P_r/T_r);  //Compressibility factor(in G4unit)
+        return z;       
+    }
+
+    // Function to calculate gas density based on Xe isotopic composition
+    G4double CalculateGasDensityFromIsotopicComposition(G4double pressure, G4double temperature,
+     const std::vector<std::pair<int, double>>& isotopicComposition)   //pressure in G4unit , temperature in G4unit
```
paolafer: I would drop the comments of G4unit throughout the class, since it's the default everywhere in nexus.

paolafer:
Good job, @enakshidey! There are a few more style-related comments, after which I would approve this PR.

paolafer (approved):
The PR adapts the code to calculate Xe density taking into account the isotope composition of the gas. Excellent job, @enakshidey! I approve it.

PR #244, opened 2024-02-21
Add necessity of env variable for Qt headers

paolafer:
This PR fixes an issue with how `scons` deals (or, better, doesn't) with the new `-F`  flags which fixes Qt compilation starting from geant4-v11.2.1. 
The user will need to define a new environment variable, `QT_DIR`, in order to be able to do only incremental compilations and not be forced to compile every time from scratch when using `scons`.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -154,6 +154,12 @@ vars.AddVariables(
                  'Path to hdf5 installation directory.',
                  NULL_PATH),
 
+    ## Qt
+
+    PathVariable('QT_DIR',
+                 'Path to Qt installation directory.',
+                 NULL_PATH),
+
     ## installation directory
 
     PathVariable('PREFIX',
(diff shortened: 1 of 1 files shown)


soleti:
I've tested this and it works on macOS 14.2.1 with clang-1500.1.0.2.5. Qt5 was installed with `brew`. The commands I used were:
```bash
export QT_DIR=$(brew --cellar qt@5)/5.15.12
scons HDF5_DIR=$(brew --cellar hdf5)/1.14.3/ -j 32
```

paolafer:
Thanks, @soleti! I've assigned you as a reviewer, you can approve the PR whenever you can!

soleti (approved):
This fix works. Incremental compilation has been verified on macOS 14.2.1 and clang-1500.1.0.2.5.

PR #248, opened 2024-02-26
Add reflections on copper and SS

kvjmistry:
This PR adds the reflections on the copper on the energy plane and on the stainless steel meshes. 

Absolute reflectivities are currently implemented without any angular dependence. References are provided in the comments.

### DIFF
--- a/source/geometries/Next100EnergyPlane.cc
+++ b/source/geometries/Next100EnergyPlane.cc
@@ -219,6 +219,16 @@ namespace nexus {
     copper_plate_posz_ = GetCoordOrigin().z()
       + gate_sapphire_wdw_dist_ + stand_out_length + copper_plate_thickn_/2.;
 
+    // Add optical surface
+    G4OpticalSurface* gas_copperplate_opsur = new G4OpticalSurface("GAS_COPPER_PLATE_OPSURF");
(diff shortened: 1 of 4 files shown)


paolafer:
Do you know or have a feeling of how much of specular/diffusive components these reflectivities have?

kvjmistry:
Scanning a bit more on the net. 

I found this for polished copper and PTFE:
https://www.sciencedirect.com/science/article/pii/S016890020701056X 
There is an angular dependence on the reflectivity and depends on the oxidation level of the copper

This note shows the reflectance of VUV on SS at different angles:
https://www-eng.lbl.gov/~shuman/NEXT/MATERIALS&COMPONENTS/VUV_light-sources/VUV_reflectance_172_Siena06.pdf
Although the reflectivity measurement is in disagreement with the other work. 

Maybe our own measurements would be useful here if we have the means to do it?

paolafer:
Yeah, I think that our own measurement would be the best way to do it. @jmalbos said he would talk with people in Canfranc to obtain a sample of the metals we're using and he would do the measurement at IFIC.
In the meanwhile, I will approve this PR to have something to start with.

paolafer (approved):
Reflectivities on metals implemented, to  the best of our current knowledge. Approved!

PR #249, opened 2024-02-27
Polygon sampler

kvjmistry:
This PR adds the option of generating events in a polygon with n sides. This is useful for generating in geometries such as the light tube. As an example I have used this generator as the default in DEMO.

The plots below show generation of Kr83 events in the INSIDE (gas), VOLUME (in the teflon) and INNER_SURF (on teflon surface)

image
image
image

I mark as draft as I don't think the translations or rotations are working and needs speed benchmarking.


paolafer:
Does this class replace also `HexagonPointSampler`? I think we can remove the class in that case. It is only used in `Next1EL`  with two regions, `ACTIVE` and `RESTRICTED`. The first region should be covered by your new sampler, and the second one is just a particular case of the first one, I think. At any rate, it is an obsolete class and we can remove the `RESTRICTED` region, if needed.

kvjmistry:
The speed improvement is not that different than what is currently done in NEXT-100. So, I will leave that as it is (the volume checker in NEXT-100 code ensures the right region is sampled anyway). Removed the hexagon point sampler in place of this code. 
Ready for review!

source/utils/PolygonPointSampler.cc
```diff
@@ -0,0 +1,261 @@
...
+      G4bool valid_sample = false;
+      std::vector<G4double> point;
+
+      while (!valid_sample){
+        
+        G4double internal_radius = min_radius_ * std::cos(M_PI/n_sides_);
+
+        G4double phi = GetPhi();
+        G4double rad = GetRadius(internal_radius, max_radius_);
+        G4double x = rad * cos(phi);
```
paolafer: You're declaring the `x` and `y` variables twice here.

source/utils/PolygonPointSampler.cc
```diff
@@ -0,0 +1,261 @@
...
+    G4double x = 0.;
+    G4double y = 0.;
+    G4double z = 0.;
+
+    // Center of the chamber
+    if (region == CENTER) {
+      x = y = z = 0.;
+    }
+
+    // Center of the chamber
```
paolafer: Change this comment (I guess it was just a repetition of the previous one).

source/utils/PolygonPointSampler.cc
```diff
@@ -0,0 +1,261 @@
...
+        valid_sample= true;
+        valid_sample = CheckXYBoundsPolygon(point, polygon_inner_);
+      }
+
+      x = point[0];
+      y = point[1];
+      z = GetLength(half_length_);
+    }
+
+    // Generating from inside the cylinder (between min_radius and max_radius)
```
paolafer: Same here (maybe cylinder --> polygon?)

source/utils/PolygonPointSampler.cc
```diff
@@ -0,0 +1,261 @@
...
+
+    // Center of the chamber
+    if (region == INSIDE) {
+      // sample position based on interior radius
+      G4bool valid_sample = false;
+      std::vector<G4double> point;
+      while (!valid_sample){
+        point = {2.0 * min_radius_ * (G4UniformRand() - 0.5),
+                 2.0 * min_radius_ * (G4UniformRand() - 0.5)};
+        valid_sample= true;
```
paolafer: Why do you assign two different values to the same variable in adjacent lines? Wouldn't the second assignment always overcome the first one?

source/utils/PolygonPointSampler.cc
```diff
@@ -0,0 +1,261 @@
...
+      // 2. Check if the point is below the maximum y coordinate of the edge
+      // 3. Check if the point is to the left of the maximum x coordinate of the edge
+      if ( (y > std::min(p1[1], p2[1])) && (y <= std::max(p1[1], p2[1])) && (x <= std::max(p1[0], p2[0])) ) {
+
+        // Calculate the x-intersection of the line connecting the point to the edge
+        G4double x_intersection = (y - p1[1]) * (p2[0] - p1[0]) / (p2[1] - p1[1]) + p1[0];
+
+        // Check if the point is on the same line as the edge or to the left of
+        // the x-intersection
+        if (p1[0] == p2[0] || x <= x_intersection) {
```
paolafer: I don't understand the comment to the first condition: if `p1` and `p2` are two subsequent vertices of the polygon, the condition of them having the same `x` coordinate doesn't mean that the point is on the same line as the edge, does it? I don't understand the purpose of the `p1[0] == p2[0]` condition. It looks like it's the case where an edge is vertical.

source/utils/PolygonPointSampler.cc
```diff
@@ -0,0 +1,261 @@
...
+
+        // point = {2.0 * max_radius_ * (G4UniformRand() - 0.5),
+        //          2.0 * max_radius_ * (G4UniformRand() - 0.5)};
+
+        point = {x,y};
+
+        G4bool inside_inner = CheckXYBoundsPolygon(point, polygon_inner_);
+        G4bool inside_outer = CheckXYBoundsPolygon(point, polygon_outer_);
+
+        // Want outside inner and inside outter
```
paolafer: Typo: outter --> outer.

paolafer (changes requested):

source/utils/PolygonPointSampler.cc
kvjmistry: fixed

source/utils/PolygonPointSampler.cc
kvjmistry: fixed

source/utils/PolygonPointSampler.cc
kvjmistry: fixed

source/utils/PolygonPointSampler.cc
kvjmistry: fixed

source/utils/PolygonPointSampler.cc
kvjmistry: removed that condition, didnt seem to do anything

source/utils/PolygonPointSampler.cc
kvjmistry: fixed

source/geometries/Next1EL.cc
```diff
@@ -804,8 +804,7 @@ void Next1EL::BuildFieldCage()
   // WORLD system of reference
   active_position_.set(0.,0.,fieldcage_position_.z()+posz);
 
-  hexrnd_ = new HexagonPointSampler(active_diam_/2., active_length_, 0.,
-				    active_position_);
+  active_gen_ = new PolygonPointSampler(active_diam_/2.,active_diam_/2. + ltube_diam_/2. + tpb_thickn_, active_length_, 6, nullptr, active_position_);
```
paolafer: I think that there's something wrong in the value of the maximum radius in this case: `ltube_diam_` has the same value as `active_diam_`.

kvjmistry: Sorry, I didn't see this comment till you mentioned it today, fixed it now!

paolafer (approved):
Nice generalization for a polygon point sampler! Approved!

PR #250, opened 2024-03-08
Fix latest tag

paolafer:
This PR fixes two things:
- Turns an exception into a warning when looking for Qt libraries through an environment variable. This variable is only needed in some MacOS versions, so we don't want an error on any machines.
- Changes the metal reflectivity from 100% Lambertian to 75% specular/25% Lambertian. This seems a reasonable assumption, until we obtain our own measurements.


gonzaponte (approved):
Clear, approved.

PR #251, opened 2024-03-22
Improve the error message when using opt photons as primary

paolafer:
Optical photons' trajectories are not stored in the simulation when using `DefaultTrackingAction`, to avoid creating huge output files. If one wants to save them, `OpticalTrackingAction` should be used instead, only for small productions and debugging.
On the other hand, `DefaultEventAction` tries to access the trajectory container, to look for the hits associated with every particle and make a cut on the total deposited energy.
As a consequence, the combination of `DefaultTrackingAction` and `DefaultEventAction` produces a segmentation fault in simulations of optical photons as primary particles, since no other particles are present in the simulation.
For the same reason, using `DefaultEventAction` but not `DefaultTrackingAction` in standard simulations also causes a segmentation fault.
This PR tries to improve the error message to cover these two cases, which are the most common causes of that particular segmentation fault.


Aretno (approved):
The error message is now much clearer and points directly to the issue. Nice!

PR #252, opened 2024-04-09
Fix compilation error

kvjmistry:
This PR fixes an issue with the compilation of nexus on some machines that require an extra header.


paolafer (approved):
Neat and simple, approved!

PR #253, opened 2024-04-17
Implement a different sampling region for the S2 light table / psf

gonzaponte:
This PR splits the EL generators in two: one for PMTs and one for SiPMs. The former generates vertices in a circle with the same diameter as the gate ring. This is the same generator as before but it hardwires the generation area to the gate ring instead of using a parameter. The generator for SiPMs produces vertices in a unit cell centered at the origin. Both cases can be restricted to slices of the EL gap (this was already implemented).

Below are plots showing the distributions of the vertices of the (new) SiPM generator:
in x-y
s2_sipms_lt

in z (full EL gap)
s2_sipms_lt_z_full

in z (partial EL gap)
s2_sipms_lt_z_slice

### DIFF
--- a/source/geometries/Next100.cc
+++ b/source/geometries/Next100.cc
@@ -244,6 +244,7 @@ namespace nexus {
              (region == "LIGHT_TUBE") ||
              (region == "HDPE_TUBE") ||
              (region == "EL_GAP_PMT") ||
+             (region == "EL_GAP_SIPM") ||
              (region == "EP_COPPER_PLATE") ||
              (region == "SAPPHIRE_WINDOW") ||
              (region == "OPTICAL_PAD") ||
--- a/source/geometries/Next100.cc
+++ b/source/geometries/Next100.cc
@@ -243,7 +243,7 @@ namespace nexus {
              (region == "XENON") ||
              (region == "LIGHT_TUBE") ||
              (region == "HDPE_TUBE") ||
-             (region == "EL_GAP") ||
+             (region == "EL_GAP_PMT") ||
              (region == "EP_COPPER_PLATE") ||
              (region == "SAPPHIRE_WINDOW") ||
              (region == "OPTICAL_PAD") ||
--- a/source/geometries/Next100.cc
+++ b/source/geometries/Next100.cc
@@ -243,8 +243,8 @@ namespace nexus {
              (region == "XENON") ||
              (region == "LIGHT_TUBE") ||
              (region == "HDPE_TUBE") ||
-             (region == "EL_GAP_PMT") ||
-             (region == "EL_GAP_SIPM") ||
+             (region == "S2_PMT_LT") ||
+             (region == "S2_SIPM_PSF") ||
              (region == "EP_COPPER_PLATE") ||
              (region == "SAPPHIRE_WINDOW") ||
              (region == "OPTICAL_PAD") ||
--- a/source/geometries/Next100FieldCage.cc
+++ b/source/geometries/Next100FieldCage.cc
@@ -738,6 +738,8 @@ void Next100FieldCage::BuildELRegion()
                                - el_gap_length_ * el_gap_slice_min_ // start of slice
                                - el_gap_slice_thickness/2.;         // center of slice
 
+  // Here vertices are generated in a cylinder with the same diameter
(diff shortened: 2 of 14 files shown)


source/geometries/Next100.cc
```diff
@@ -243,7 +243,8 @@ namespace nexus {
              (region == "XENON") ||
              (region == "LIGHT_TUBE") ||
              (region == "HDPE_TUBE") ||
-             (region == "EL_GAP") ||
+             (region == "EL_GAP_PMT") ||
```
paolafer: I would use a name that tells you straight away that this region is used for light tables/PSF. Otherwise, it could be misleading, what do you think? What about `PMT_LIGHT_TABLE_S2`? and `SIPM_PSF_S2`? Are they too cumbersome?

source/geometries/Next100FieldCage.cc
```diff
@@ -170,36 +170,19 @@ Next100FieldCage::Next100FieldCage(G4double grid_thickn):
...
+  G4GenericMessenger::Command& el_gap_slice_min_cmd =
+    msg_->DeclareProperty("el_gap_slice_min", el_gap_slice_min_,
+                          "Lower limit (fraction) to the EL gap slice in which vertices are generated.");
+  el_gap_slice_min_cmd.SetParameterName("el_gap_slice_min", false);
+  el_gap_slice_min_cmd.SetRange("el_gap_slice_min >= 0.0 &&"
+                                "el_gap_slice_min <= 1.0");
+
+  G4GenericMessenger::Command& el_gap_slice_max_cmd =
+    msg_->DeclareProperty("el_gap_slice_max", el_gap_slice_max_,
+                          "Uppeer limit (fraction) to the EL gap slice in which vertices are generated.");
```
paolafer: Typo: Uppeer --> Upper.

source/geometries/Next100FieldCage.cc
```diff
@@ -170,36 +170,19 @@ Next100FieldCage::Next100FieldCage(G4double grid_thickn):
...
-  el_gap_gen_disk_zmin_cmd.SetRange("el_gap_gen_disk_zmin>=0.0 && el_gap_gen_disk_zmin<=1.0");
-
-  G4GenericMessenger::Command& el_gap_gen_disk_zmax_cmd =
-    msg_->DeclareProperty("el_gap_gen_disk_zmax", el_gap_gen_disk_zmax_,
-                          "Maximum Z range of the EL gap vertex generation disk.");
-  el_gap_gen_disk_zmax_cmd.SetParameterName("el_gap_gen_disk_zmax", false);
-  el_gap_gen_disk_zmax_cmd.SetRange("el_gap_gen_disk_zmax>=0.0 && el_gap_gen_disk_zmax<=1.0");
+  G4GenericMessenger::Command& el_gap_slice_min_cmd =
+    msg_->DeclareProperty("el_gap_slice_min", el_gap_slice_min_,
+                          "Lower limit (fraction) to the EL gap slice in which vertices are generated.");
```
paolafer: What about `fraction of the whole length`? And starting from which side? Maybe we can make ir clearer in the description.

source/geometries/Next100.cc
gonzaponte: I'm fine with it and I don't have a strong preference. What do you prefer? Here are some options:
- `PMT_LIGHT_TABLE_S2` / `SIPM_PSF_S2`
- `S2_PMT_LIGHT_TABLE` or `S2_PMT_LT` / `S2_SIPM_PSF`
- `S2_PMT_LT` / `S2_SIPM_PSF`
- any other variation

paolafer: I like `S2_PMT_LT / S2_SIPM_PSF`!

paolafer:
I realized that we have an obsolete pair of macros as an example for the S2 generation in NEXT100. Maybe you could modify them, so we also have an example for these new vertex regions.

paolafer:
With the change of macros names, you should also modify `macros_test.py`, in the `pytest` folder, to include the strings `PSF` and `LT` in the list of "slow" jobs.

paolafer (approved):
A more correct way of generating EL tables, with the same statistic for each distance in the PSF and a more robust code for PMT light table. Approved!

PR #254, opened 2024-04-19
Use dielectric grids for automatic overlap checks

paolafer:
The automatic tests that perform the overlap checks were taking ages to be executed, since the implementation of stainless steel grids as the default option in NEXT-100. This PR replaces the macros used for the overlap checks with others that use dielectric grids (the other option).

### DIFF
--- a/macros/NEXT100opt_ovlp.config.mac
+++ b/macros/NEXT100opt_ovlp.config.mac
@@ -1,0 +1,34 @@
+## ----------------------------------------------------------------------------
+## nexus | NEXT100opt_ovlp.config.mac
+##
+## Initialization macro to simulate Kr-83 decays in the NEXT-100 geometry
+## with fast simulation and grids implemented as dielectric,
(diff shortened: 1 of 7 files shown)


gonzaponte (approved):
Great job!

macros/NEXT100opt_ovlp.config.mac
```diff
@@ -0,0 +1,36 @@
...
+
+/process/em/verbose 0
+
+##### GEOMETRY #####
+/Geometry/Next100/elfield false
+/Geometry/Next100/pressure 10. bar
+/Geometry/Next100/max_step_size 1. mm
+/Geometry/Next100/use_dielectric_grid true
+
+/process/optical/processActivation Cerenkov false
```
gonzaponte: Oops, it seems to be complaining about this. I guess it's because there is no optical physics.

paolafer: Yeah, I just removed that line.

PR #255, opened 2024-04-22
Warn users of deprecated classes

gonzaponte:
- ELLookupTable
- RadiusDependentDriftField
- WavelengthShifting


paolafer:
I'm thinking that we should also remove the addition of the `WavelengthShifting` process from `NexusPhysics`, so there's no doubt that we are not using it anymore. Can you do that?

gonzaponte:
Yes. Why is it that we don't see this process even though it is added to the list?

paolafer:
Because our custom-made process uses properties with different names than those defined in the `TPB` material. Right now there are no optical material properties with the appropriate parameters, so that process never gets to do anything.

source/physics/RadiusDependentDriftField.cc
```diff
@@ -3,6 +3,11 @@
 //
 // Drift field varying with radial coordinate.
 //
+//
```
paolafer: Maybe, rather than deprecated, this class is still not implemented, right?

paolafer (approved):
This PR clarifies which classes are not being used in nexus, to avoid confusion for users.

PR #256, opened 2024-04-26
Add protection for wrong keys in object factory

gonzaponte:
If the wrong name is used for, say, a generator, the user gets a wild segfault with zero information about what went wrong. This PR adds protection against bad strings and provides a more helpful message.

Bear in mind that part of the message is compiler-dependent. There are no guarantees that `typeid(x).name()` gives a human-readable output. Thankfully, most compilers do provide a sensible output, so this might be a good enough compromise between simplicity and correctness.

I also added a test that verifies that the exception is raised.


paolafer (approved):
Nice utility to make debugging easier! It works fine! Approved.

PR #257, opened 2024-06-13
Argon

kvjmistry:
This PR includes some updates to the NEXT100 geometry and argon properties relevant for the LPR argon run. I have also included some example configuration files with baseline parameters for running in the LPR mode. 

Since the NEXT100 geometry had the drift velocities hard-coded. I have added a configurable parameter for this now.


source/materials/MaterialsList.cc
```diff
@@ -187,13 +187,18 @@ namespace materials {
     G4Material* mat = G4Material::GetMaterial(name, false);
 
     if (mat == 0) {
-      G4NistManager* nist = G4NistManager::Instance();
 
       mat = new G4Material(name, ArgonDensity(pressure), 1,
         kStateGas, temperature, pressure);
 
-      G4Element* Ar = nist->FindOrBuildElement("Ar");
+      G4Element* Ar = new G4Element(name, "Ar", 1);
```
paolafer: Why do you think we need to change that? Aren't we using natural argon? Or is Geant4-defined argon wrong?

paolafer:
The optical material properties of some materials, such as the sapphire windows or the PMT windows, are not defined for energies greater than 10.3 eV and 10.7 eV, respectively, because the formulas we're using are not valid beyond that point. This could be a problem because we are simulating argon scintillation light up to 11.2 eV; if a photon at the end of the spectrum hits a material without optical properties for its energy, it will have an unexpected behaviour.

macros/NEXT100_Ar_alpha_full.config.mac
```diff
@@ -0,0 +1,56 @@
...
+##### GEOMETRY #####
+/Geometry/Next100/elfield true
+/Geometry/Next100/EL_field 6.7 kV/cm
+/Geometry/Next100/max_step_size 1. mm
+/Geometry/Next100/pressure 4.2 bar
+/Geometry/Next100/gas GAr
+/Geometry/PmtR11410/time_binning 25. nanosecond
+/Geometry/Next100/sipm_time_binning  1. microsecond
+
+## Geant4 only permits units of mm/ns via the macro
```
paolafer: You can also define a new unit, if you prefer.

source/materials/MaterialsList.cc
kvjmistry: I was going to include all the isotopes of ar, but their natural percentages are too low. Reverted back to just Ar 40 through G4.

macros/NEXT100_Ar_alpha_full.config.mac
kvjmistry: Didnt know thats how it worked, updated now

kvjmistry:
> The optical material properties of some materials, such as the sapphire windows or the PMT windows, are not defined for energies greater than 10.3 eV and 10.7 eV, respectively, because the formulas we're using are not valid beyond that point. This could be a problem because we are simulating argon scintillation light up to 11.2 eV; if a photon at the end of the spectrum hits a material without optical properties for its energy, it will have an unexpected behaviour.

I will see if I can find any references for relevant materials.

kvjmistry:
> > The optical material properties of some materials, such as the sapphire windows or the PMT windows, are not defined for energies greater than 10.3 eV and 10.7 eV, respectively, because the formulas we're using are not valid beyond that point. This could be a problem because we are simulating argon scintillation light up to 11.2 eV; if a photon at the end of the spectrum hits a material without optical properties for its energy, it will have an unexpected behaviour.
> 
> I will see if I can find any references for relevant materials.

@paolafer I couldn't find much, it might be because the transmittance of 128nm light through sapphire and UV FS tanks very significantly at these wavelengths. The plot below is from the crab paper:
image

How is it best to deal with this?

paolafer:
If the CRAB plot is the only piece of information we have at high energies, maybe we can add a few points more to those properties to match that transmittance values. We would need to do a small simulation and vary the refraction index and the absorption length to match the desired transmittance. What do people think?

macros/NEXT100_Ar_alpha_full.config.mac
```diff
@@ -0,0 +1,56 @@
...
+/Generator/SingleParticle/min_energy 5.6 MeV
+/Generator/SingleParticle/max_energy 5.6 MeV
+/Generator/SingleParticle/region ACTIVE
+
+##### ACTIONS #####
+/Actions/DefaultEventAction/min_energy 50 keV
+/Actions/DefaultEventAction/max_energy 6 MeV
+
+##### PHYSICS #####
+## No full simulation
```
paolafer: This comment is misleading, since this is actually full simulation. Please, change it,

source/geometries/Next100Vessel.cc
```diff
@@ -342,16 +342,24 @@ namespace nexus {
...
       G4Exception("[Next100Vessel]", "Construct()", FatalException,
 		  "Unknown kind of xenon, valid options are: "
                   "natural, enriched, depleted, or XeHe.");
     }
 
+    // Gas Properties
+    if (gas_ == "GAr"){
+      vessel_gas_mat->SetMaterialPropertiesTable(opticalprops::GAr(sc_yield_, e_lifetime_));
+    }
+    else{
```
paolafer: A blank space is missing after `else`.

kvjmistry:
> If the CRAB plot is the only piece of information we have at high energies, maybe we can add a few points more to those properties to match that transmittance values. We would need to do a small simulation and vary the refraction index and the absorption length to match the desired transmittance. What do people think?

I also found this paper which has the transmission and reflectance for quartz and sapphire for a few more thicknesses and angles of incidence. https://link.springer.com/article/10.1134/S0020441206030195
image

Transmittance is small (e.g. from CRAB measurements) or basically zero in these plots for FS and Sapphire. Reflectance ranges from 20-30% for FS, and 2-20% for Sapphire. I need to brush up on my optics, can we calculate the refractive index from these numbers or does it need to be done anlytically?

paolafer:
I'm not sure about how to do it.

soleti:
I think you should be able to calculate the refractive index starting from the reflectance at a specific angle using the Fresnel equations no?

kvjmistry:
In this last update I have udpated the refractive index for the out of range value used in the Semellier Eq for FS and Sapphire. 

Currently we use the last recorded value in the Semellier equation. I have bumped this up to value of 10 to make sure the transmission goes to zero. This is based on the paper cited in the comment/cited above. 

I have also rebased the branch to master.

source/geometries/Next100FieldCage.cc
```diff
@@ -742,7 +743,15 @@ void Next100FieldCage::BuildELRegion()
...
-    el_field->SetLightYield(XenonELLightYield(ELelectric_field_, pressure_));
+    
+    // Decide whether to use argon or xenon LY
+    if (gas_->GetName() == "GAr"){
+      el_field->SetLightYield(ArgonELLightYield(ELelectric_field_, pressure_));
+    }
+    else {
+      el_field->SetLightYield(XenonELLightYield(ELelectric_field_, pressure_));
+    }
+
```
gonzaponte, line 767: Should we protect against invalid names like everywhere else?

source/geometries/Next100Vessel.cc
```diff
@@ -342,16 +342,24 @@ namespace nexus {
...
+    // Gas Properties
+    if (gas_ == "GAr"){
+      vessel_gas_mat->SetMaterialPropertiesTable(opticalprops::GAr(sc_yield_, e_lifetime_));
+    }
+    else {
     vessel_gas_mat->SetMaterialPropertiesTable(opticalprops::GXe(pressure_,
                                                                  temperature_,
                                                                  sc_yield_,
                                                                  e_lifetime_));
+    }
```
gonzaponte, line 362: This is an issue for a different PR, but it seems that we are defining the same gas in different places, perhaps we should revisit this and propagate the gas material from the top down.

source/materials/ArgonGasProperties.cc
```diff
@@ -24,29 +24,49 @@ namespace nexus {
...
+                                    { 5.0  * bar, 8.2268 * kg/m3},
+                                    { 6.0  * bar, 9.8788 * kg/m3},
+                                    { 7.0  * bar, 11.533 * kg/m3},
+                                    { 8.0  * bar, 13.189 * kg/m3},
+                                    { 9.0  * bar, 14.848 * kg/m3},
+                                    { 10.0 * bar, 16.508 * kg/m3},
+                                    { 15.0 * bar, 24.843 * kg/m3},
+                                    { 20.0 * bar, 33.231 * kg/m3},
+                                    { 30.0 * bar, 50.155 * kg/m3}};
+    G4bool found = false;
```
gonzaponte: you could initialize density to 0 and then check if it's still 0 later so you don't need to keep track of a separate variable

source/materials/OpticalMaterialProperties.cc
```diff
@@ -65,13 +65,20 @@ namespace opticalprops {
...
-    ri_energy.push_back(optPhotMaxE_);          // This sets the refractive index between optPhotFusedSilicaMaxE_ and
-    rIndex.push_back(rIndex[rIndex.size()-1]);  // optPhotMaxE_ to the value obtained at optPhotFusedSilicaMaxE_
+
+    // The rindex is not defined for sapphire beyond optPhotFusedSilicaMaxE_ = 10.7 eV
+    // Measurements https://link.springer.com/article/10.1134/S0020441206030195 imply
+    // that the transmission goes to zero for energies higher than this (>120 nm)
+    // We set n to be an arbitarily high value of 10 so the value is
+    // implemented in the simulation (transmission goes to zero)
+    ri_energy.push_back(optPhotMaxE_);
+    rIndex.push_back(10);
```
gonzaponte: If I recall correctly, G4 interpolates between values, so the refractive index between `optPhotFusedSilicaMaxE_ ` and `optPhotMaxE_ ` will grow linearly. Is this what we want or do we want it to be high immediately after `optPhotFusedSilicaMaxE_ `?

source/materials/OpticalMaterialProperties.cc
```diff
@@ -431,14 +438,20 @@ namespace opticalprops {
...
+
+    // The rindex is not defined for sapphire beyond optPhotSapphireMaxE_ = 10.3 eV
+    // Measurements https://link.springer.com/article/10.1134/S0020441206030195 imply
+    // that the transmission goes to zero for energies higher than this (>120 nm)
+    // We set n to be an arbitarily high value of 10 so the value is
+    // implemented in the simulation (transmission goes to zero)
     ri_energy.push_back(optPhotMaxE_);
-    rIndex.push_back(rIndex[rIndex.size()-1]);
+    rIndex.push_back(10);
+
```
gonzaponte, line 453: same comment as above.

macros/NEXT100_Ar_alpha_full.config.mac
```diff
@@ -0,0 +1,64 @@
...
+##### PERSISTENCY #####
+/nexus/persistency/output_file NEXT100_Ar_alpha
+
+/nexus/random_seed             100
+/nexus/persistency/start_id    100
+
+## eventType options: bb0nu, bb2nu, background
+/nexus/persistency/event_type background
+
+/process/optical/processActivation Cerenkov false
```
gonzaponte: Missing a newline at the end of file.

gonzaponte (commented):
Looks good. Just a couple of questions.

source/geometries/Next100FieldCage.cc
kvjmistry: Added protection. I realize though for the XeHe case we are adopting the pure xe scintillation light. For tonne-scale simulation in the future, we will want to adjust this to account for the small reduction in EL light yield when we add He.

source/geometries/Next100Vessel.cc
kvjmistry: agreed. I will keep a note of this improvement.

source/materials/OpticalMaterialProperties.cc
kvjmistry: It now sets the rindex max at optPhotSapphireMaxE_, then the rindex is set to 10 for anthing higher than this now.

source/materials/ArgonGasProperties.cc
kvjmistry: check against density = 0 now

source/materials/OpticalMaterialProperties.cc
kvjmistry: It now sets the rindex max at optPhotFusedSilicaMaxE_, then the rindex is set to 10 for anything higher than this now.

macros/NEXT100_Ar_alpha_full.config.mac
kvjmistry: fixed

gonzaponte (approved):
Important update on gas properties. Good job!

PR #260, opened 2024-09-13
Generic material properties

### LINKED ISSUE #12: Ionization energy & Fano Factor of Xenon gas
Currently these values are hard-coded into IonizationClustering.cc

These Xenon Gas properties should be set in OpticalMaterialProperties.cc, and IonizationClustering.cc get them from there. There is commented code ready to do this.

At the same time, we should check that the values set are the most updated ones.

paolafer:
This PR drops the `optical` part of the name of the OpticalMaterialProperties files and relative namespace. Since we use more properties that are not just optical (such as scintillation and TPB emission spectra, yield, attachment...) this seems more logical. 
Two more non-optical properties, namely the Fano factor and the ionization energy for gaseous xenon,  are added to the property table instead of being coded in the `IonizationClustering` class.
This PR closes #12.

### DIFF
--- a/source/materials/MaterialProperties.cc
+++ b/source/materials/MaterialProperties.cc
@@ -1,12 +1,12 @@
 // ----------------------------------------------------------------------------
-// nexus | OpticalMaterialProperties.cc
+// nexus | MaterialProperties.cc
 //
-// Optical properties of relevant materials.
+// Properties of relevant materials.
 //
 // The NEXT Collaboration
 // ----------------------------------------------------------------------------
 
-#include "OpticalMaterialProperties.h"
+#include "MaterialProperties.h"
 #include "XenonProperties.h"
 #include "SellmeierEquation.h"
 
@@ -16,7 +16,7 @@
 using namespace nexus;
 using namespace CLHEP;
 
-namespace opticalprops {
+namespace materialprops {
 
   G4MaterialPropertiesTable* Vacuum()
   {
@@ -114,7 +114,7 @@ namespace opticalprops {
 
     G4MaterialPropertiesTable* mpt = new G4MaterialPropertiesTable();
 
-    G4MaterialPropertiesTable* fused_sil_pt = opticalprops::FusedSilica();
+    G4MaterialPropertiesTable* fused_sil_pt = materialprops::FusedSilica();
     mpt->AddProperty("RINDEX", fused_sil_pt->GetProperty("RINDEX"));
 
     // ABSORPTION LENGTH (Set to match the transparency)
@@ -735,7 +735,7 @@ namespace opticalprops {
     G4MaterialPropertiesTable* mpt = new G4MaterialPropertiesTable();
 
     // PROPERTIES FROM XENON
-    G4MaterialPropertiesTable* xenon_pt = opticalprops::GXe(pressure, temperature, sc_yield, e_lifetime);
+    G4MaterialPropertiesTable* xenon_pt = materialprops::GXe(pressure, temperature, sc_yield, e_lifetime);
 
     mpt->AddProperty("RINDEX",        xenon_pt->GetProperty("RINDEX"));
     mpt->AddProperty("SCINTILLATIONCOMPONENT1", xenon_pt->GetProperty("SCINTILLATIONCOMPONENT1"));
@@ -820,6 +820,7 @@ namespace opticalprops {
        hc_ / (717.13442 * nm),  hc_ / (597.48892 * nm),  hc_ / (477.84343 * nm),
        hc_ / (418.02068 * nm),  hc_ / (358.19793 * nm),  hc_ / (293.94387 * nm)
     };
+
     std::vector<G4double> REFLECTIVITY = {
       .99088, .99082, .98925,
       .98623, .98611, .98163,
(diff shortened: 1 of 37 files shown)


gonzaponte:
I've edited your comment to [close the related issue automatically](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword):

`closes issue #12 `-> `closes #12`

source/physics/IonizationClustering.cc
```diff
@@ -109,15 +109,13 @@ namespace nexus {
...
-    //  return G4VRestDiscreteProcess::PostStepDoIt(track, step);
-
-    //G4double ioni_energy = mpt->GetConstProperty("IONIZATIONENERGY");
-    //G4double fano_factor = mpt->GetConstProperty("FANOFACTOR");
-    G4double ioni_energy = 22.4 * eV;
-    G4double fano_factor = .15;
+    G4MaterialPropertiesTable* mpt =
+      track.GetMaterial()->GetMaterialPropertiesTable();
+    if (!mpt)
+      return G4VRestDiscreteProcess::PostStepDoIt(track, step);
```
gonzaponte: Shouldn't this be an error?

paolafer: Yes, it would be probably better.

gonzaponte: Also can it be done at class initialization instead of at every call to `PostStepDoIt`?

paolafer: You can't do it in the class initialization, because you need access to the track to know in which material you are passing through at each moment.

source/materials/MaterialProperties.cc
```diff
@@ -708,6 +715,8 @@ namespace opticalprops {
     LXe_mpt->AddConstProperty("SCINTILLATIONYIELD1", .03);
     LXe_mpt->AddConstProperty("SCINTILLATIONYIELD2", .97);
     LXe_mpt->AddConstProperty("ATTACHMENT", 1000.*ms, 1);
+    LXe_mpt->AddConstProperty("IONIZATIONENERGY",    21.9 * eV, 1);
+    LXe_mpt->AddConstProperty("FANOFACTOR",               0.29, 1);
```
gonzaponte: These values correspond to
https://journals.aps.org/pra/abstract/10.1103/PhysRevA.46.1676
which are GXe measurements.
For LXe, Doke measured Wi=15.6 eV in https://journals.aps.org/pra/abstract/10.1103/PhysRevA.12.1771
The same paper gives fano factors of the order of 0.05, which seem too low, but in
https://www.sciencedirect.com/science/article/abs/pii/0168900294012369
they also measure 0.033.

source/physics/IonizationClustering.cc
```diff
@@ -109,15 +109,17 @@ namespace nexus {
...
-    G4double ioni_energy = 22.4 * eV;
-    G4double fano_factor = .15;
+    G4String mat_name = track.GetMaterial()->GetName();
+    G4MaterialPropertiesTable* mpt =
+      track.GetMaterial()->GetMaterialPropertiesTable();
+    if (!mpt) {
+      G4cout << "Material properties missing for material " << mat_name << G4endl;
+      G4Exception("[IonizationClustering]", "PostStepDoIt()", FatalException,
+		  "Material properties table not defined in material!");
+      }
```
gonzaponte: Since the material name is used only within the `if` statement, let's put it there.

source/materials/MaterialProperties.cc
paolafer: I've changed the ionization energy. However, I'm puzzled about the Fano factor ( in NEXT papers we used the value 20 sometimes...) : maybe we should review also the value we use in xenon, then.

source/physics/IonizationClustering.cc
paolafer: Fixed.

source/materials/MaterialProperties.cc
paolafer: See, for instance, https://journals-aps-org.ezproxy.cern.ch/prb/pdf/10.1103/PhysRevB.68.054201.

paolafer: See, for instance 
[PhysRevB.68.054201.pdf](https://github.com/user-attachments/files/17001738/PhysRevB.68.054201.pdf), where they talk about discrepancies between calculations and measurements.

gonzaponte: > in NEXT papers we used the value 20 sometimes...

In gas? The discrepancy seems to be only in liquid. For gas I thought we always used 0.15, although the first paper above measured 0.29.

> [PhysRevB.68.054201.pdf](https://github.com/user-attachments/files/17001738/PhysRevB.68.054201.pdf), where they talk about discrepancies between calculations and measurements.

Not sure what to make of this... Should we use F=20 then?

paolafer: > > in NEXT papers we used the value 20 sometimes...
> 
> In gas? The discrepancy seems to be only in liquid. For gas I thought we always used 0.15, although the first paper above measured 0.29.

I was referring to liquid. Yes, we used 0.15, but it's probably time to review it, also.
> 
> > [PhysRevB.68.054201.pdf](https://github.com/user-attachments/files/17001738/PhysRevB.68.054201.pdf), where they talk about discrepancies between calculations and measurements.
> 
> Not sure what to make of this... Should we use F=20 then?

gonzaponte: I believe simulations in LXe need to go in a different direction because we need to introduce the anticorrelation between ionization and scintillation. This requires adapting nexus to this case, which may not be trivial. Also, we might want to use NEST (by the way, they use F=0.03 for LXe).

PR #262, opened 2025-01-29
Update EL gain equation

kvjmistry:
This PR fixes the EL gain calculation in nexus which was using an incorrect intercept for higher pressures.


gonzaponte (approved):
Perfect.

PR #264, opened 2025-03-19
Add option for aluminized end in GenericWLSFiber

soleti:
This PR adds the possibility of adding an aluminized coating at the end of a WLS fiber.


kvjmistry (approved):
This change adds the option for an aluminized end for WLS fibers. Everything looks good to merge, so approved.

PR #265, opened 2025-03-26
Work around Geant4 limitations for reading int64

gonzaponte:
The following command
`/nexus/persistency/start_id <a number greater than 2**32-1> `
throws an exception. This comes from a (stupid) limitation in Geant4 that doesn't allow parsing 64-bit integers from values passed by commands. We work around this limitation by using a function from the standard library to parse this number.


kvjmistry (approved):
Changes look good with me! Thanks for doing this fix.

PR #266, opened 2025-06-13
Add NEXT100 generator for events on cathode surface

kvjmistry:
This PR adds a generator to make events in a 1 micron thickness, 1 micron in front of the cathode mesh. This is for modelling the plated-out Rn events. 

According to the geometry, the cathode mesh sits at `z = 1187.37 mm`. 100k test events were generated with the following coordinates:

var | initial_x | initial_y | initial_z
-- |-- | -- | --
min| -482.316803 | -482.406372 | 1187.302979
max| 482.427917 | 482.360321 | 1187.305054

The z positions are just in front of the cathode mesh surface. 

image

### DIFF
--- a/source/geometries/Next100FieldCage.cc
+++ b/source/geometries/Next100FieldCage.cc
@@ -480,13 +480,13 @@ void Next100FieldCage::BuildCathode()
                                            GetCoordOrigin().y(),
                                            cathode_zpos_));
 
-  // Generate in a 1 micron thick disk in front of cathode
+  // Generate in a small 1 micron thick disk in front of cathode
   cathode_surf_gen_ =
     new CylinderPointSampler(0.0, cathode_int_diam_/2.,
-                             1.0*micrometer,0., twopi, nullptr,
+                             0.5*micrometer,0., twopi, nullptr,
                              G4ThreeVector(GetCoordOrigin().x(),
                                            GetCoordOrigin().y(),
-                                           cathode_grid_zpos - grid_thickn_ - 1*micrometer));
+                                           cathode_grid_zpos - grid_thickn_ - 1.0*micrometer));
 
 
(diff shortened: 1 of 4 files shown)


source/geometries/Next100FieldCage.cc
```diff
@@ -480,6 +480,14 @@ void Next100FieldCage::BuildCathode()
                                            GetCoordOrigin().y(),
                                            cathode_zpos_));
 
+  // Generate in a 1 micron thick disk in front of cathode
+  cathode_surf_gen_ =
+    new CylinderPointSampler(0.0, cathode_int_diam_/2.,
+                             1.0*micrometer,0., twopi, nullptr,
```
gonzaponte: Not that it matters, but to get 1 um, you should specify 0.5*micrometer here.

source/geometries/Next100FieldCage.cc
```diff
@@ -480,6 +480,14 @@ void Next100FieldCage::BuildCathode()
...
+                             G4ThreeVector(GetCoordOrigin().x(),
+                                           GetCoordOrigin().y(),
+                                           cathode_grid_zpos - grid_thickn_ - 1*micrometer));
```
gonzaponte: and here.

2 um is fine, though :)

gonzaponte (approved):
As I've annotated, the generator is actually 2 um thick, not 1. I don't think this matters, so I approve it anyway. Feel free to change it, if you prefer it to be 1 um.

source/geometries/Next100FieldCage.cc
kvjmistry: ah yes, the half-width! 
2 micron is fine shouldn't make much of a difference. I'll update the thickness so it matches with the desired shift.

PR #267, opened 2025-10-30
Update SiPM mask hole dimensions

gonzaponte:
As reported by @halmamol, the size of the holes in the SiPM masks is 6x6 mm2, not 6x5 mm2


kvjmistry (approved):
Looks good with me!

PR #268, opened 2025-11-01
Reduce PTFE reflectivity by 4 percent

kvjmistry:
Based on studies with alpha particles, the PTFE reflectivity is better matched in NEXT-100 to 94%.
https://next.ific.uv.es/cgi-bin/DocDB/private/ShowDocument?docid=1739


gonzaponte (approved):
Modifies the reflectivity of PTFE for visible wavelengths only. 👍

PR #269, opened 2025-11-01
Update NEXT100 TPB thickness

kvjmistry:
Based on https://next.ific.uv.es/cgi-bin/DocDB/private/ShowDocument?docid=1775 the TPB thickness should be closer to 3-4 microns.  3.5 micron is input as an average.

### DIFF
--- a/source/geometries/Next100EnergyPlane.cc
+++ b/source/geometries/Next100EnergyPlane.cc
@@ -57,7 +57,7 @@ namespace nexus {
 
     sapphire_window_thickn_ (6. * mm),
     optical_pad_thickn_ (1.0 * mm),
-    tpb_thickn_ (1.*micrometer),
+    tpb_thickn_ (5.4*micrometer),
 
     pmt_stand_out_ (2. * mm), // length that PMTs stand out of copper, in the front
     pmt_base_diam_ (46.8 * mm),
--- a/source/geometries/Next100FieldCage.cc
+++ b/source/geometries/Next100FieldCage.cc
@@ -92,7 +92,7 @@ Next100FieldCage::Next100FieldCage(G4double grid_thickn):
   num_drift_rings_ (48),
   num_buffer_rings_ (4),
 
-  tpb_thickn_ (1 * micrometer),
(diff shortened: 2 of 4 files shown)


gonzaponte:
Unfortunately, the thickness of TPB is defined independently in multiple places. There are ***at least*** two more definitions: one in NEXT100EnergyPlane and one in NEXT100SiPMBoard. 

Of course, the right thing to do is to have a single variable and propagate it, but if we want to push the release forward I'd be happy to postpone it, provided we ensure all variables are updated.

Also, can you check that modifying this variable does not introduce volume overlaps?

kvjmistry:
I can check the overlaps, will update here when ive checked that. 

I think it makes sense to have different implementations of the TPB thickness, though, since the coatings were done independently, we should check with IFIC on the EP/TP coatings if there are measurements. This measurement should be just for the light tube.

gonzaponte:
Ok. I agree.

kvjmistry:
As far as overlaps go, we should be good:
```
Checking overlaps for volume LIGHT_TUBE_DRIFT:0 (G4Polyhedra) ... OK! 
Checking overlaps for volume DRIFT_TPB:0 (G4Polyhedra) ... OK! 
Checking overlaps for volume LIGHT_TUBE_BUFFER:0 (G4Polyhedra) ... OK! 
Checking overlaps for volume BUFFER_TPB:0 (G4Polyhedra) ... OK! 
```
I am checking with Miryam about the coatings for EP/TP.

kvjmistry:
I got some values from Quim. 
The way it is done is the DICE, Masks, and PMT Windows were done in batches. Within each batch there are several monitors to measure the TPB uniformity across the evarporation plate (Crystal 1 - Crystal 4). The image shows the average for each batch. Then the global average. Do you think using the global average works?
image

gonzaponte:
I think so, in principle above 2 microns there should be very little difference.

kvjmistry:
Updated Dice, Mask and PMT window TPB thickness according to tables above. Overlaps have been checked and seem ok.

gonzaponte (approved):
Looks good to me!

PR #270, opened 2025-11-01
Add AD HOC generator position for NEXT100

kvjmistry:
Add AD HOC generator option for NEXT-100 geometry to make it easier to generate events with any specified position in the active volume.

### DIFF
--- a/source/geometries/Next100FieldCage.cc
+++ b/source/geometries/Next100FieldCage.cc
@@ -115,7 +115,8 @@ Next100FieldCage::Next100FieldCage(G4double grid_thickn):
   // EL gap generation disk parameters
   el_gap_slice_min_(0.), el_gap_slice_max_(1.),
   sipm_pitch_(0),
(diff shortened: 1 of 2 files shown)


gonzaponte:
This option exists already and it's implemented in NEXT100.cc (and NEXT100OpticalGeometry.cc). In fact, the function you added is never called, because the root geometry is NEXT100 (or NEXT100OpticalGeometry), not NEXT100FieldCage.

In case it's not clear, the way this works is that the root geometry propagates the "region" argument to some sub-geometry, which does the same, until one component returns a value. In general, if you add a region identifier for random position generation, you would need to add this propagation all the way to the root geometry (or geometries), but for this specific label, the method is implemented directly in the root geometry because it is not specific to any sub-geometry.

kvjmistry:
Oops, I have no idea why I didn't check for the root AD_HOC generation. Closing this PR.

PR #272, opened 2025-12-15
Fix distance between dice boards

gonzaponte:
The difference in sipm x-positions between adjacent dices was not the same as within a board (15.55 mm). The geometry was previously tweaked in #170 and #219, but somewhere along the way we forgot to ensure the distances were correct.


halmamol (approved):
This PR corrects the wrong distance between TP diceboards caused by several changes on geometry reviews. Now the pitch between SiPMs is consistent (15.55mm), the sensor plane is centered in (0,0) and numbers match correctly with the engineer drawings. Good job!

PR #273, opened 2026-01-19
Delay full simulation until deposited energy is known

gonzaponte:
Full simulation is computationally very expensive. The main reason is light tracking, as a large number of optical photons can be produced per event. Moreover, nexus is capable of disregarding (not saving) events depending on their energy. However, this is criterion is only checked at the end of the full event simulation, which includes the propagation of photons. This can result in a long simulation that ends up being thrown away.

In this PR we introduce a stacking action that delays the propagation of photons (and ionization electrons, which turn will only produce more optical photons) until all particles that can deposit energy are tracked. The result is a two-stage simulation similar to what we have in what we call fast-sim:
- a (fast) first stage in which the "main" particles are tracked, which allows to compute the deposited energy
- a (slow) second stage in which optical photons are produced[^1] and tracked

with the second stage performed only when the event meets the deposited energy criterion.

[^1]: S1 photons are actually produced in the first stage, but not tracked immediately. S2 photons are generated during the second stage, when the ionization electrons delayed from the first stage are tracked.

### DIFF
--- a/macros/NEXT100.init.mac
+++ b/macros/NEXT100.init.mac
@@ -22,6 +22,8 @@
 /nexus/RegisterRunAction DefaultRunAction
 /nexus/RegisterEventAction DefaultEventAction
 /nexus/RegisterTrackingAction DefaultTrackingAction
+/nexus/RegisterStackingAction DefaultStackingAction
+/nexus/RegisterSteppingAction DefaultSteppingAction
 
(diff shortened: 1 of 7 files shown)


kvjmistry (approved):
This is a really nice upgrade to the nexus simulation!
Code has been tested and runs as expected.
Excellent job @gonzaponte!

PR #275, opened 2026-04-23
Extend TPB WLS absorption range

gonzaponte:
TPB has an absorption spectrum up to ~430 nm which overlaps with the emission spectrum (Fig 11 in [this paper](https://arxiv.org/pdf/1709.05002)). The current implementation only adds the absorption data up to ~380 nm and sets a virtually infinite absorption length for longer wavelengths. This was done intentionally to avoid secondary absorptions and reemissions. We know this is not correct and does not match the data both from the detector and other dedicated setups. As a result, the data-mc agreement in the NEXT-100 detector is good!


kvjmistry (approved):
This is an important fix to the simulation. Good catch!

PR #276, opened 2026-05-25
Fix XeHe molar fraction material construction

jerenner:
Fixes a bug in the creation of the Xe/He mixture, identified by @martinperezmaneiro 

G4Material::AddElement interprets the numeric fraction argument as a mass fraction. GXeHe was using the requested Xe percentage directly, which made 85/15 Xe/He become 85/15 by mass rather than by molar fraction.

Compute the enriched-xenon molar mass, form the Xe/He mixture molar mass from the requested molar fractions, and derive the mass fractions passed to Geant4 from those molar fractions. Use the same mixture molar mass for the gas density calculation so the material density and composition are internally consistent.

### DIFF
--- a/source/geometries/Next100Vessel.cc
+++ b/source/geometries/Next100Vessel.cc
@@ -73,7 +73,7 @@ namespace nexus {
     // Visibility
     visibility_(0),
     gas_("enrichedXe"),
-    xe_he_xe_type_("naturalXe"),
+    xe_he_type_("naturalXe"),
     helium_mass_num_(4),
     xe_perc_(100.),
     th_source_("no_source"),
@@ -105,10 +105,10 @@ namespace nexus {
     msg_->DeclareProperty("gas", gas_, "Gas being used");
     msg_->DeclareProperty("XePercentage", xe_perc_, "Percentage of xenon used in mixtures");
     msg_->DeclareProperty("helium_A", helium_mass_num_, "Mass number for helium used, 3 or 4");
-    G4GenericMessenger::Command& xe_he_xe_type_cmd =
-      msg_->DeclareProperty("XeHeXenonType", xe_he_xe_type_,
+    G4GenericMessenger::Command& xe_he_type_cmd =
+      msg_->DeclareProperty("XeHeXenonType", xe_he_type_,
                             "Xenon type used in XeHe mixtures.");
-    xe_he_xe_type_cmd.SetCandidates("naturalXe enrichedXe depletedXe");
+    xe_he_type_cmd.SetCandidates("naturalXe enrichedXe depletedXe");
 
     G4GenericMessenger::Command& pressure_cmd =
       msg_->DeclareProperty("pressure", pressure_, "Xenon pressure");
@@ -346,7 +346,7 @@ namespace nexus {
       vessel_gas_mat =  materials::GXeDepleted(pressure_, temperature_);
     } else if  (gas_ == "XeHe") {
       vessel_gas_mat = materials::GXeHe(pressure_, 300. * kelvin,
-					    xe_perc_, helium_mass_num_, xe_he_xe_type_);
+					    xe_perc_, helium_mass_num_, xe_he_type_);
     } else if  (gas_ == "GAr") {
       vessel_gas_mat = materials::GAr(pressure_, temperature_);
     } else {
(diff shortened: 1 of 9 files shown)


kvjmistry (changes requested):
I have reviewed this PR and it looks good. I have a few extra requests before merging:
- Please could you also fix the GXeAr function which suffers from this same issue. 
- Please add a comment to the function that this is currently configured for enriched Xe + helium. In the future, we may want to add depleted Xe + He as an option. Or if you can think of a better way to include the Xe Enrichment that would be better.

jerenner:
The GXeAr function is now updated, and there are comments in the .cc and .h stating that the Xe/He mixture is enriched Xe + He. 

We could continue by adding depleted Xe + He (or adding some way to select between natural, depleted, and enriched), or we can do that in a new PR

kvjmistry:
> The GXeAr function is now updated, and there are comments in the .cc and .h stating that the Xe/He mixture is enriched Xe + He.
> 
> We could continue by adding depleted Xe + He (or adding some way to select between natural, depleted, and enriched), or we can do that in a new PR

If you have the time to add the option between switching between natural/depleted/enriched then that would be great. I think that would be useful for the tonne scale simulations in the future.

jerenner:
The last commit adds functionality to set the XeHeXenonType in the macro for `Next100`, `NextNew`, and `NextTonScale`, for example:
```
/Geometry/NextTonScale/XeHeXenonType enrichedXe
```
The options are: `naturalXe enrichedXe depletedXe`

source/geometries/Next100Vessel.cc
```diff
@@ -73,6 +73,7 @@ namespace nexus {
     // Visibility
     visibility_(0),
     gas_("enrichedXe"),
+    xe_he_xe_type_("naturalXe"),
```
kvjmistry: I think here i would prefer the var name `xe_he_type_` which is a little easier to read.

source/geometries/Next100Vessel.cc
```diff
@@ -104,6 +105,10 @@ namespace nexus {
     msg_->DeclareProperty("gas", gas_, "Gas being used");
     msg_->DeclareProperty("XePercentage", xe_perc_, "Percentage of xenon used in mixtures");
     msg_->DeclareProperty("helium_A", helium_mass_num_, "Mass number for helium used, 3 or 4");
+    G4GenericMessenger::Command& xe_he_xe_type_cmd =
+      msg_->DeclareProperty("XeHeXenonType", xe_he_xe_type_,
+                            "Xenon type used in XeHe mixtures.");
+    xe_he_xe_type_cmd.SetCandidates("naturalXe enrichedXe depletedXe");
```
kvjmistry: ive not see the use of `SetCandidates` before. Could you test this functionality by giving it a mis-spelled word and see if it works?

source/geometries/Next100Vessel.cc
kvjmistry: add a comment in the header for var definition that this refers to the xenon type mixed with the helium

source/geometries/Next100Vessel.cc
jerenner: The method [does exist in the Geant4 source](https://github.com/Geant4/geant4/blob/41f2dd79968018de4efb966f2546970b30c4af78/source/intercoms/src/G4GenericMessenger.cc#L412).

Running a test in a macro with a misspelling:
```
/Geometry/NextTonScale/XeHeXenonType enrichedXXe
```
gives:
```
parameter value (enrichedXXe) is not listed in the candidate List.
  Candidates are: naturalXe enrichedXe depletedXe

-------- EEEE ------- G4Exception-START -------- EEEE -------
*** G4Exception : ExecCommand()
      issued by : [BatchSession]
Illegal parameter: /Geometry/NextTonScale/XeHeXenonType enrichedXXe
*** Fatal Exception *** core dump ***
 **** Track information is not available at this moment
 **** Step information is not available at this moment

-------- EEEE -------- G4Exception-END --------- EEEE -------
```

kvjmistry: great, thanks for checking!

jerenner:
Variable names changed to `xe_he_type_` and commented in headers that the variable refers to the xenon type mixed with the helium

kvjmistry (approved):
This PR fixes an issue related to how we defined the gas mixtures. The fixes were (for the first time?) using AI to provide the changes. All looks good to me!

